PR #47, opened 2020-02-06
fix(storage): fix system test and change scope for iam access token

### LINKED ISSUE #46: Storage: access token field is unused in system test 'test_create_signed_read_url_*_w_access_token'
Currently in the system test `blob.generate_signed_url`  method passing [access_token=None](https://github.com/googleapis/python-storage/blob/master/tests/system.py#L981), Due to that `test_create_signed_read_url_*_w_access_token`  test not executed properly,so  needs to pass `value of access_token`. also need to change the [scope](https://github.com/googleapis/python-storage/blob/master/tests/system.py#L1058) which passing to generate the access token as the current one is not working.

HemangChothani:
Fixes #46

### DIFF
--- a/tests/system.py
+++ b/tests/system.py
@@ -1057,7 +1057,7 @@ class TestStorageSignURLs(unittest.TestCase):
         name = client.service_account_path("-", service_account_email)
         scope = [
             'https://www.googleapis.com/auth/devstorage.read_write',
-            'https://www.googleapis.com/auth/iam'
+            'https://www.googleapis.com/auth/iam',
         ]
         response = client.generate_access_token(name, scope)
         self._create_signed_read_url_helper(
@@ -1071,7 +1071,7 @@ class TestStorageSignURLs(unittest.TestCase):
         name = client.service_account_path("-", service_account_email)
         scope = [
             'https://www.googleapis.com/auth/devstorage.read_write',
-            'https://www.googleapis.com/auth/iam'
+            'https://www.googleapis.com/auth/iam',
         ]
         response = client.generate_access_token(name, scope)
         self._create_signed_read_url_helper(
--- a/tests/system.py
+++ b/tests/system.py
@@ -977,8 +977,8 @@ class TestStorageSignURLs(unittest.TestCase):
             method=method,
             client=Config.CLIENT,
             version=version,
-            service_account_email=None,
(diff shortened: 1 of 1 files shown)


HemangChothani:
System test throws 'Identity and Access Management (IAM) API has not been used in project 1065521786570 before or it is disabled. ' error, so need (IAM) permission for this project.

tests/system.py
```diff
@@ -1066,7 +1067,7 @@ def test_create_signed_read_url_v4_w_access_token(self):
         client = iam_credentials_v1.IAMCredentialsClient()
         service_account_email = Config.CLIENT._credentials.service_account_email
         name = client.service_account_path("-", service_account_email)
-        scope = ["https://www.googleapis.com/auth/devstorage.read_write"]
+        scope = ["https://www.googleapis.com/auth/cloud-platform"]
```
crwilcox: This seems like a lot of access. @jkwlui @frankyn is this scope needed for this?

tests/system.py
```diff
@@ -1055,8 +1055,9 @@ def test_create_signed_read_url_v2_w_access_token(self):
...
         response = client.generate_access_token(name, scope)
+        print(response.access_token)
```
crwilcox: Is this leftover from debugging by chance?

tests/system.py
crwilcox: (we don't have to block merging on this as it is a test, but if this is needed it seems like a large scope for narrow use?)

tseaver: FWIW, I can confirm that both tests fail on master with a 403 without this patch.

tests/system.py
tseaver: It should definitely go:  we don't want stray output in normal test runs.

tests/system.py
frankyn: Could you narrow the scope down to: `https://www.googleapis.com/auth/iam`

Documented at the bottom of the following document: https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signBlob

frankyn: To clarify, you can do join both:

```
scope = ['https://www.googleapis.com/auth/devstorage.read_write', 'https://www.googleapis.com/auth/iam']
```

PR #48, opened 2020-02-07
feat(storage): improve v4 signature query parameters encoding

### LINKED ISSUE #9: [Storage] V4 Signature Supplying Additional Query Parameters With GCS Signed URLs

[New Client Request - Support For Query Params With GCS Signed URLs](https://docs.google.com/document/d/1PnH2K-M-2wkkYkUyLBh7JDDRqYgBNimP5fUFuEeB4u4/edit#heading=h.xgsgzd636vko)

IlyaFaer:
Closes #9

### DIFF
--- a/google/cloud/storage/_signing.py
+++ b/google/cloud/storage/_signing.py
@@ -510,7 +510,7 @@ def generate_signed_url_v4(
 
     :type query_parameters: dict
     :param query_parameters:
-        (Optional) Additional query paramtersto be included as part of the
+        (Optional) Additional query parameters to be included as part of the
         signed URLs.  See:
         https://cloud.google.com/storage/docs/xml-api/reference-headers#query
 
@@ -586,8 +586,7 @@ def generate_signed_url_v4(
     if generation is not None:
         query_parameters["generation"] = generation
 
-    ordered_query_parameters = sorted(query_parameters.items())
-    canonical_query_string = six.moves.urllib.parse.urlencode(ordered_query_parameters)
+    canonical_query_string = _url_encode(query_parameters)
 
     canonical_elements = [
         method,
@@ -666,3 +665,35 @@ def _sign_message(message, access_token, service_account_email):
 
     data = json.loads(response.data.decode("utf-8"))
     return data["signature"]
+
+
+def _url_encode(query_params):
+    """Encode query params into URL.
+
+    :type query_params: dict
+    :param query_params: Query params to be encoded.
+
+    :rtype: str
+    :returns: URL encoded query params.
+    """
+    query_params = sorted(query_params.items())
+
+    params = []
+    for name, value in query_params:
+        params.append(_quote_param(name) + "=" + _quote_param(value))
+
+    return "&".join(params)
+
+
+def _quote_param(param):
+    """Quote query param.
+
+    :type param: dict
+    :param param: Query param to be encoded.
+
+    :rtype: str
+    :returns: URL encoded query param.
+    """
+    if not isinstance(param, bytes):
+        param = str(param)
+    return six.moves.urllib.parse.quote(param, safe="~")
--- a/google/cloud/storage/_signing.py
+++ b/google/cloud/storage/_signing.py
@@ -682,9 +682,10 @@ def _url_encode(query_params):
     :rtype: str
     :returns: URL encoded query params.
     """
-    params = []
-    for name, value in query_params.items():
-        params.append(_quote_param(name) + "=" + _quote_param(value))
+    params = [
+        "{}={}".format(_quote_param(name), _quote_param(value))
+        for name, value in query_params.items()
+    ]
 
     return "&".join(sorted(params))
 
--- a/google/cloud/storage/_signing.py
+++ b/google/cloud/storage/_signing.py
@@ -693,7 +693,7 @@ def _url_encode(query_params):
 def _quote_param(param):
     """Quote query param.
 
-    :type param: dict
+    :type param: Union[str, bytes]
     :param param: Query param to be encoded.
 
     :rtype: str
--- a/google/cloud/storage/_signing.py
+++ b/google/cloud/storage/_signing.py
@@ -693,7 +693,7 @@ def _url_encode(query_params):
 def _quote_param(param):
     """Quote query param.
 
-    :type param: Union[str, bytes]
+    :type param: Any
     :param param: Query param to be encoded.
 
     :rtype: str
--- a/google/cloud/storage/_signing.py
+++ b/google/cloud/storage/_signing.py
@@ -676,13 +676,11 @@ def _url_encode(query_params):
     :rtype: str
     :returns: URL encoded query params.
     """
-    query_params = sorted(query_params.items())
-
     params = []
-    for name, value in query_params:
+    for name, value in query_params.items():
         params.append(_quote_param(name) + "=" + _quote_param(value))
 
-    return "&".join(params)
+    return "&".join(sorted(params))
 
 
 def _quote_param(param):
--- a/tests/unit/test__signing.py
+++ b/tests/unit/test__signing.py
@@ -702,6 +702,19 @@ class Test_sign_message(unittest.TestCase):
                 )
 
 
+class TestCustomURLEncoding(unittest.TestCase):
+    def test_url_encode(self):
+        from google.cloud.storage._signing import _url_encode
+
+        # param1 includes safe symbol ~
+        # param# includes symbols, which must be encoded
+        query_params = {"param1": "value~1-2", "param#": "*value+value/"}
+
+        self.assertEqual(
+            _url_encode(query_params), "param%23=%2Avalue%2Bvalue%2F&param1=value~1-2"
+        )
+
+
 _DUMMY_SERVICE_ACCOUNT = None
 
 
--- a/tests/unit/test__signing.py
+++ b/tests/unit/test__signing.py
@@ -754,6 +754,13 @@ class TestQuoteParam(unittest.TestCase):
         self.assertIsInstance(encoded_param, str)
         self.assertEqual(encoded_param, "%D0%81%D0%99%D0%A6%D0%AF%D0%A9%D0%AF%D0%A9")
 
+    def test_bytes(self):
+        from google.cloud.storage._signing import _quote_param
+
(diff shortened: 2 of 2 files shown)


google/cloud/storage/_signing.py
```diff
@@ -666,3 +665,33 @@ def _sign_message(message, access_token, service_account_email):
...
+    :param query_params: Query params to be encoded.
+
+    :rtype: str
+    :returns: URL encoded query params.
+    """
+    params = []
+    for name, value in query_params.items():
+        params.append(_quote_param(name) + "=" + _quote_param(value))
+
+    return "&".join(sorted(params))
```
IlyaFaer: First encode, then sort, then join into single one `str`

google/cloud/storage/_signing.py
```diff
@@ -666,3 +665,33 @@ def _sign_message(message, access_token, service_account_email):
...
+
+    :type param: dict
+    :param param: Query param to be encoded.
+
+    :rtype: str
+    :returns: URL encoded query param.
+    """
+    if not isinstance(param, bytes):
+        param = str(param)
+    return six.moves.urllib.parse.quote(param, safe="~")
```
IlyaFaer: Use standard `quote()`  with `~` as safe. This `safe` arg by default has value `/`, so with setting `safe` to `~` we're saying that `/` is not safe.

tests/unit/test__signing.py
```diff
@@ -702,6 +702,19 @@ def test_sign_bytes_failure(self):
...
+    def test_url_encode(self):
+        from google.cloud.storage._signing import _url_encode
+
+        # param1 includes safe symbol ~
+        # param# includes symbols, which must be encoded
+        query_params = {"param1": "value~1-2", "param#": "*value+value/"}
+
+        self.assertEqual(
+            _url_encode(query_params), "param%23=%2Avalue%2Bvalue%2F&param1=value~1-2"
+        )
```
IlyaFaer: Test that includes mentioned safe `~` and unsafe `/`

google/cloud/storage/_signing.py
```diff
@@ -586,8 +586,7 @@ def generate_signed_url_v4(
     if generation is not None:
         query_parameters["generation"] = generation
 
-    ordered_query_parameters = sorted(query_parameters.items())
-    canonical_query_string = six.moves.urllib.parse.urlencode(ordered_query_parameters)
```
IlyaFaer: No more standard Python `urlencode()`

google/cloud/storage/_signing.py
crwilcox: @frankyn is sorting needed here? The old code did the same though I wouldn't think the order was important?

IlyaFaer: @crwilcox, @frankyn, yeah, that's mentioned in the docs. It's why I moved sorting from the start of the encoding code to the end - to strictly follow these lines

image

frankyn:
@IlyaFaer this LGTM, but please update relevant conformance tests before merging this code to confirm that it's working as expected. 

I updated conformance tests in https://github.com/googleapis/conformance-tests/tree/master/storage/v1

If you have questions PLMK. Specifically for this change you should rely on the following two conformance tests:

* Query Parameter Encoding
* Query Parameter Ordering

google/cloud/storage/_signing.py
tseaver: I don't understand that comment:

```bash
$ .nox/unit-3-8/bin/python
Python 3.8.1 (default, Feb 11 2020, 17:29:15) 
[GCC 7.4.0] on linux
>>> import six
>>> six.moves.urllib
<module 'six.moves.urllib' (<six._SixMetaPathImporter object at 0x7fc1d4781550>)>
>>> six.moves.urllib.parse
<module 'six.moves.urllib_parse'>
>>> six.moves.urllib.parse.urlencode
<function urlencode at 0x7fc1d3e59b80>
```

google/cloud/storage/_signing.py
```diff
@@ -666,3 +665,33 @@ def _sign_message(message, access_token, service_account_email):
...
+    """Quote query param.
+
+    :type param: dict
+    :param param: Query param to be encoded.
+
+    :rtype: str
+    :returns: URL encoded query param.
+    """
+    if not isinstance(param, bytes):
+        param = str(param)
```
tseaver: We cannot use `str` here:  it returns bytes only under Python2!  Instead, we can let `quote` do the conversions for us:  it uses `utf-8` by default to encode non-ASCII characters.

tests/unit/test__signing.py
```diff
@@ -728,6 +741,7 @@ def _run_conformance_test(resource, test_data):
         method=test_data["method"],
         _request_timestamp=test_data["timestamp"],
         headers=test_data.get("headers"),
+        query_parameters=test_data.get("queryParameters"),
     )
```
tseaver: Please add a testcase class which explicitly tests the new `_quote_param` function with all kinds of expected values:

- Simple, non-quoted ASCII characters
- Characters which *must* be quoted (e.g., `"/"`)
- Characters which must *not* be quoted (`"~"`)
- Unicode characters, which must be converted to UTF-8 and encoded

The tests should assert that the returned value is of the correct type, as well as matching the expected value.

google/cloud/storage/_signing.py
```diff
@@ -666,3 +665,33 @@ def _sign_message(message, access_token, service_account_email):
...
+
+    :type query_params: dict
+    :param query_params: Query params to be encoded.
+
+    :rtype: str
+    :returns: URL encoded query params.
+    """
+    params = []
+    for name, value in query_params.items():
+        params.append(_quote_param(name) + "=" + _quote_param(value))
```
tseaver: Clearer and faster as a list comprehension using formatting, e.g.:

```python
    params = ["{}={}".format(_quote_param(name), _quote_param(value))
        for name, value in query_params.items()]

**Comment by tseaver** [comment]
This PR also addresses part of the change for #12.

**Comment by tseaver** [comment]
@frankyn We don't currently do *any* of the cross-language conformance tests inside `python-storage`.  I just made #57 to track that.

**Comment by frankyn** [comment]
@IlyaFaer please respond to @tseaver's comments before merging

File: google/cloud/storage/_signing.py (hunk 4 above)
Comment (IlyaFaer) [reply google/cloud/storage/_signing.py]: > OK, I see now that you mean to say we can't use urlencode, because it doesn't quote according to the V4 spec.
Yes

File: google/cloud/storage/_signing.py (hunk 5 above)
Comment (IlyaFaer) [reply google/cloud/storage/_signing.py]: Not sure I understand. I've taken this conversion from original `urlencode()` (I mean, I've tried not to diverge too much from it). If we'll erase these lines, we'll get a lot of `TypeError` from conformance tests. `quote()` [seems to be expecting](https://github.com/python/cpython/blob/8dbdf5f275c6462bb522bcf3a29054239d72989d/Lib/urllib/parse.py#L806-L818) only `str` or `bytes` as an arg

File: google/cloud/storage/_signing.py [hunk 8]
```diff
@@ -672,3 +671,34 @@ def _sign_message(message, access_token, service_account_email):
...
+        for name, value in query_params.items()
+    ]
+
+    return "&".join(sorted(params))
+
+
+def _quote_param(param):
+    """Quote query param.
+
+    :type param: dict
```
tseaver: This type is incorrect:  it should be one of `str` or `bytes` (or allow both).

google/cloud/storage/_signing.py
tseaver: The example you link here doesn't use `str()` to convert anything: t uses `string.encode` to convert *text* to bytes, then (in `quote_from_bytes`) does the quoting at the bytes level, and finally returns *text* using `''.join()`.

I'm not sure what you mean about a `TypeError`:  do the conformance tests pass anything other than text / bytes as a value?

google/cloud/storage/_signing.py
tseaver: OK, so we can't use `urlencode`:  is there a reason why we can't use the stdlib's `quote` function, passing our own `safe` set?

google/cloud/storage/_signing.py
IlyaFaer: Yes, `X-Goog-Expires` parameter comes in as an `int`, so all of the conformance tests are failing, as all of them have:
https://github.com/googleapis/python-storage/blob/23df542d0669852b05139023d5ef1ae14a09f4c7/tests/unit/url_signer_v4_test_data.json#L7-L7

as well as `test_with_access_token`, `test_w_response_type`, `test_w_response_disposition`, etc. - all with the same `TypeError`, because of expiration parameter.

> The example you link here doesn't use str()

No, no, I've linked `quote()` itself to see that it's expecting `str` or `bytes`. If we're speaking about conversion (which I'm keeping), than we should look [here](https://github.com/python/cpython/blob/0d860dd43c72dc7046a5d18fc72d495cadd4a2df/Lib/urllib/parse.py#L899-L907). Thus, if `k` or `v` is not `bytes`, they're converting it into `str`.

It's the `urllib.parse.urlencode()` behavior, so I suppose this conversion was always here

google/cloud/storage/_signing.py
frankyn: I think the implementation LGTM. @tseaver do you still have concerns?

frankyn:
Friendly ping and pending Kokoro passing.

frankyn:
Merging when green.

google/cloud/storage/_signing.py
tseaver: If this `_quote_param` function is meant to take only `str` or `bytes` (as you've updated the docstring to indicate), then it is up to the caller to handle converting any other types:  this method should *not* coerce using `str`.

IlyaFaer: No, it actually will deal with any type. Fixed

frankyn:
Pending approval from @tseaver as there is on-going discussion.

frankyn:
@tseaver friendly ping

google/cloud/storage/_signing.py
crwilcox: As written this seems to handle python 2 and python 3. I am going to leave this as is. I agree the coercion isn't ideal.

PR #56, opened 2020-02-12
fix: make v4 signing formatting consistent w/ spec

### LINKED ISSUE #12: [Storage] V4 Signature: Formatting Inconsistencies

[New Client Request - GCS SigV4 Formatting Inconsistencies](https://docs.google.com/document/d/11OpYYj2FMFsWe-slIKK-JQC97CUh4CYOWeraR0wuImo).

tseaver:
Fixes #12 

W/ exception of query string parameter quoting bits being addressed in PR #48.

### DIFF
--- a/google/cloud/storage/_signing.py
+++ b/google/cloud/storage/_signing.py
@@ -18,7 +18,6 @@ import binascii
 import collections
 import datetime
 import hashlib
-import re
 import json
 
 import six
@@ -31,8 +30,6 @@ from google.cloud import _helpers
 
 
 NOW = datetime.datetime.utcnow  # To be replaced by tests.
-MULTIPLE_SPACES_RE = r"\s+"
-MULTIPLE_SPACES = re.compile(MULTIPLE_SPACES_RE)
 
 SERVICE_ACCOUNT_URL = (
     "https://googleapis.dev/python/google-api-core/latest/"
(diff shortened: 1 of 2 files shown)


google/cloud/storage/_signing.py
```diff
@@ -589,13 +586,24 @@ def generate_signed_url_v4(
     ordered_query_parameters = sorted(query_parameters.items())
     canonical_query_string = six.moves.urllib.parse.urlencode(ordered_query_parameters)
 
+    quoted_resource = six.moves.urllib.parse.quote(resource)
+
+    lowercased_headers = dict(ordered_headers)
+
+    if "x-goog-content-sha256" in lowercased_headers:
+        payload = lowercased_headers["x-goog-content-sha256"]
+    elif "x-amz-content-sha256" in lowercased_headers:
```
frankyn: Only support x-goog- and ignore x-amz.

tests/unit/test__signing.py
```diff
@@ -577,7 +577,7 @@ def _generate_helper(
         )
         self.assertEqual(scheme, expected_scheme)
         self.assertEqual(netloc, expected_netloc)
-        self.assertEqual(path, resource)
+        self.assertEqual(path, six.moves.urllib.parse.quote(resource))
```
frankyn: Does this encode '/' as well? It should not.

tseaver: It does not:  it is actually designed to quote the path part.

frankyn: This might hit an issue with conformance tests when that's added but approving for now.

tseaver: As it happens, I backed it out here, because the bucket / blob already do the urlencoding of the resource before calling `_sigining.generate_signed_url_v4`.

PR #58, opened 2020-02-13
feat: generate signed URLs for blobs/buckets using virtual hostname

### LINKED ISSUE #10: [Storage] V4 Signature Support for virtual hosted-style signed

[New Client Request - Support for virtual hosted-style signed URLs](https://docs.google.com/document/d/1TkoDiZjz6nINNVQrYbrfaODhmcMTR-crb94aR1cXZVA/edit#)

tseaver:
Closes #10.

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -361,7 +361,7 @@ class Blob(_PropertyMixin):
         version=None,
         service_account_email=None,
         access_token=None,
-        use_virtual_hostname=False,
+        virtual_host_style=False,
     ):
         """Generates a signed URL for this blob.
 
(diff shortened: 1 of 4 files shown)


google/cloud/storage/blob.py
```diff
@@ -361,6 +361,7 @@ def generate_signed_url(
         version=None,
         service_account_email=None,
         access_token=None,
+        use_virtual_hostname=False,
```
frankyn: Please use `virtual_hosted_style` instead.

frankyn: Please use `virtual_host_style`

tseaver: Which spelling?  You typed `virtual_hosted_style` above. :)

tseaver: I picked this one, as it is shorter. LMK if I guessed wrong.

frankyn: Whoops :man_facepalming: use `virtual_hosted_style` so it's consistent across languages.

tseaver: Done: b75fd1d

PR #67, opened 2020-02-19
feat: support '<' comparision between blob objects

### LINKED ISSUE #66: Support '<' comparision for some blob objects
It would be great to take a list of blob objects, from the same bucket, and sort them lexicographically by name. While it appears that `list_blobs` always returns blobs sorted by name, the [API endpoint documentation](https://cloud.google.com/storage/docs/json_api/v1/objects/list) does not provide any such guarantee. Note, however, that there is such a mention here: https://cloud.google.com/storage/docs/listing-objects.

Unfortunately, without said guarantee on sort order, we provide an explicit sort sometimes when iterating through blobs. Said code ends up looking like this:

```python
for blob in sorted(bucket.list_blobs(), key=lambda blob: blob.name):
    ...
```

While that's not really terrible, it'd be great to more concisely write:

```python
for blob in sorted(bucket.list_blobs()):
    ...
```

Even if the documentation were to guarantee an order for `list_blobs`, it may still be valuable to be able to sort collections of blobs.

Providing such support requires only defining `__lt__` on the bucket object. In fact I've said changes and will link such pull request momentarily.

bboe:
Fixes #66 🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -190,6 +190,14 @@ class Blob(_PropertyMixin):
         if generation is not None:
             self._properties["generation"] = generation
 
+    def __lt__(self, other):
+        if not isinstance(other, Blob):
+            return super(Blob, self).__lt__(other)
+
+        if self.bucket != other.bucket:
+            raise TypeError("'<' not supported between blobs from different buckets")
+        return self.name < other.name
+
     @property
     def bucket(self):
         """Bucket which contains the object.
--- a/tests/unit/test_blob.py
+++ b/tests/unit/test_blob.py
@@ -3224,6 +3224,28 @@ class Test_Blob(unittest.TestCase):
         self.assertEqual(blob.name, "b")
         self.assertEqual(blob.bucket.name, "buckets.example.com")
 
+    def test__lt__(self):
+        BUCKET = object()
+        blob_first = self._make_one("A", bucket=BUCKET)
+        blob_second = self._make_one("a", bucket=BUCKET)
+        blob_third = self._make_one("b", bucket=BUCKET)
+
+        self.assertLess(blob_first, blob_second)
+        self.assertLess(blob_second, blob_third)
+
+        self.assertEqual([blob_first, blob_second, blob_third], sorted([blob_third, blob_second, blob_first]))
+
+    def test__lt__w_other_instance(self):
+        with self.assertRaises(TypeError):
+            self._make_one("A", bucket=object()) < None
+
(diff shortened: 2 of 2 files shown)


tests/unit/test_blob.py
```diff
@@ -3224,6 +3224,28 @@ def test_from_string_w_domain_name_bucket(self):
...
+        blob_third = self._make_one("b", bucket=BUCKET)
+
+        self.assertLess(blob_first, blob_second)
+        self.assertLess(blob_second, blob_third)
+
+        self.assertEqual([blob_first, blob_second, blob_third], sorted([blob_third, blob_second, blob_first]))
+
+    def test__lt__w_other_instance(self):
+        with self.assertRaises(TypeError):
+            self._make_one("A", bucket=object()) < None
```
tseaver: The `TypeError` here won't be raised in Python 2.x.:

```python
Python 2.7.17 (default, Nov  7 2019, 10:07:09) 
[GCC 7.4.0] on linux2
>>> class Foo(object): pass
... 
>>> Foo() < None
False
>>> Foo() < object()
True
```

tseaver: Also, please move the construction of the blob instance outside the `with` statement, just in case *it* is the source of the `TypeError`:  we only want to trap the error from the comparison itself.  E.g.:

```python
    def test__lt__w_other_instance(self):
        blob = self._make_one("A", bucket=object())
        with self.assertRaises(TypeError):
             blob < None
```

google/cloud/storage/blob.py
```diff
@@ -190,6 +190,14 @@ def __init__(
         if generation is not None:
             self._properties["generation"] = generation
 
+    def __lt__(self, other):
+        if not isinstance(other, Blob):
+            return super(Blob, self).__lt__(other)
+
+        if self.bucket != other.bucket:
+            raise TypeError("'<' not supported between blobs from different buckets")
```
tseaver: Hmm, I wonder if it would make sense to allow blobs from different buckets to be sorted, using the bucket name?  E.g.:

```python
        return self.bucket.name < other.bucket.name and self.name < other.name
```

tests/unit/test_blob.py
bboe: > The TypeError here won't be raised in Python 2.x.

That is interesting.  I could either _always_ manually raise a `TypeError` in said cases, or simply adjust the test to handle the specific Python 2.7 case. The former approach might be considered a backwards incompatible change in the event any python 2.7 people happen to be sorting said lists already. My preference would be for the latter case -- simply ignore that test in python 2.7.

PR #88, opened 2020-03-17
feat: add support for crc32c checksums

crwilcox:
Adds dependencies for crc32c checksumming libraries.

### DIFF
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -18,7 +18,6 @@ These are *not* part of the API.
 """
 
 import base64
-import crc32c
 from hashlib import md5
 import os
 
@@ -342,6 +341,6 @@ def _get_crc32c_module():
 
             hash_obj = crcmod.predefined.Crc('crc-32c')
         except:
-            raise Exception("Missing a crc package. Install google-crc32c or crcmod")
+            raise ImportError("Failed to import either `google-crc32c` or `crcmod`)
 
     return hash_obj
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -335,16 +335,21 @@ def _get_crc32c_object():
         try:
             import crcmod
 
-            # Determine if this is using the slow form of crcmod.
-            nested_crcmod = __import__(
-                "crcmod.crcmod", globals(), locals(), ["_usingExtension"], 0,
-            )
-            fast_crc = getattr(nested_crcmod, "_usingExtension", False)
-            if not fast_crc:
-                warnings.warn(_SLOW_CRC32C_WARNING, RuntimeWarning, stacklevel=2)
-
             crc_obj = crcmod.predefined.Crc("crc-32c")
+            _is_fast_crcmod()
+
         except ImportError:
             raise ImportError("Failed to import either `google-crc32c` or `crcmod`")
 
     return crc_obj
+
+
+def _is_fast_crcmod():
+    # Determine if this is using the slow form of crcmod.
+    nested_crcmod = __import__(
+        "crcmod.crcmod", globals(), locals(), ["_usingExtension"], 0,
+    )
+    fast_crc = getattr(nested_crcmod, "_usingExtension", False)
+    if not fast_crc:
+        warnings.warn(_SLOW_CRC32C_WARNING, RuntimeWarning, stacklevel=2)
+    return fast_crc
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -18,6 +18,7 @@ These are *not* part of the API.
 """
 
 import base64
+import crc32c
 from hashlib import md5
 import os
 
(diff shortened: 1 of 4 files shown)


crwilcox:
All is passing, Coverage needs to be brought back up.

crwilcox:
When reviewing, would like feedback around testing imports (https://github.com/googleapis/python-storage/pull/88/files#diff-1b28f95dd7d9d14e05bbb9f4434772d2R394)

google/cloud/storage/_helpers.py
```diff
@@ -297,3 +304,52 @@ def _base64_md5hash(buffer_object):
     _write_buffer_to_hash(buffer_object, hash_obj)
     digest_bytes = hash_obj.digest()
     return base64.b64encode(digest_bytes)
+
+
+def _base64_crc32chash(buffer_object):
```
frankyn: How will end-users use these helper methods?

google/cloud/storage/_helpers.py
```diff
@@ -297,3 +304,52 @@ def _base64_md5hash(buffer_object):
...
+
+    :type buffer_object: bytes buffer
+    :param buffer_object: Buffer containing bytes used to compute an MD5
+                          hash (as base64).
+
+    :rtype: str
+    :returns: A base64 encoded digest of the MD5 hash.
+    """
+    hash_obj = _get_crc32c_object()
+    _write_buffer_to_hash(buffer_object, hash_obj)
```
frankyn: This asserts that the object is being hashed before upload right?

google/cloud/storage/_helpers.py
```diff
@@ -297,3 +304,52 @@ def _base64_md5hash(buffer_object):
...
+
+def _base64_crc32chash(buffer_object):
+    """Get crc32c hash of bytes (as base64).
+
+    :type buffer_object: bytes buffer
+    :param buffer_object: Buffer containing bytes used to compute an MD5
+                          hash (as base64).
+
+    :rtype: str
+    :returns: A base64 encoded digest of the MD5 hash.
```
frankyn: crc32c not MD5 hash.

google/cloud/storage/_helpers.py
```diff
@@ -297,3 +304,52 @@ def _base64_md5hash(buffer_object):
     _write_buffer_to_hash(buffer_object, hash_obj)
     digest_bytes = hash_obj.digest()
     return base64.b64encode(digest_bytes)
+
+
+def _base64_crc32chash(buffer_object):
+    """Get crc32c hash of bytes (as base64).
+
+    :type buffer_object: bytes buffer
+    :param buffer_object: Buffer containing bytes used to compute an MD5
```
frankyn: crc32c not MD5 hash.

google/cloud/storage/_helpers.py
```diff
@@ -297,3 +304,52 @@ def _base64_md5hash(buffer_object):
...
+    try:
+        import crc32c
+
+        crc_obj = crc32c.Checksum()
+    except ImportError:
+        try:
+            import crcmod
+
+            crc_obj = crcmod.predefined.Crc("crc-32c")
+            _is_fast_crcmod()
```
frankyn: should you store the return value coming from `_is_fast_crcmod()` into `crc_obj`?

google/cloud/storage/_helpers.py
```diff
@@ -297,3 +304,52 @@ def _base64_md5hash(buffer_object):
...
+    hash_obj = _get_crc32c_object()
+    _write_buffer_to_hash(buffer_object, hash_obj)
+    digest_bytes = hash_obj.digest()
+    return base64.b64encode(digest_bytes)
+
+
+def _get_crc32c_object():
+    """ Get crc32c object
+    Attempt to use the Google-CRC32c package. If it isn't available, try
+    to use CRCMod. CRCMod might be using a 'slow' varietal. If so, warn...
```
frankyn: I might be confused but is CRCMod the Python version of crc32c implementation instead of the C++ library? 
Should it be called Python Native version which is slow?

crwilcox: This is a bit confusing. 
crcmod -> python 2
  - contains c and a python impl
crc32c -> python 3
  - contains c and python impl

crwilcox: I could reword this message maybe though?

frankyn: Oh, CRCmod is only for Python 2? That wasn't super clear to me.
Why not support both through the new crc32c? (If it's a maintenance burden then that's fine.)

google/cloud/storage/_helpers.py
crwilcox: I could do this, though it isn't actually on the type so I am hesitant to do that, especially since crcmod is an external dependency.

google/cloud/storage/_helpers.py
crwilcox: I believe that is where this is ultimately used yes. This is near identical to the calculation to the md5 hashing above as the two hashes provide a shared interface.

google/cloud/storage/_helpers.py
crwilcox: It is a large maintenance burden and requires different c extension style. Python 2 is deprecated also and I anticipate dropping support from new releases within the year.

crwilcox:
Hold off on this. we need to add crc32c support to resumable media, then remove direct crc32 deps from here.

frankyn:
Closing this PR and  leaving branch for @andrewsg information.

PR #89, opened 2020-03-24
docs(storage): add documentation for blob and bucket name property

### LINKED ISSUE #82: Documentation: Blob name is not documented as a property
Received an internal request to update Storage Documentation to include Blob.name property.

HemangChothani:
Fixes #82

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -123,20 +123,23 @@ class Blob(_PropertyMixin):
     :param bucket: The bucket to which this blob belongs.
 
     :type chunk_size: int
-
-    :param chunk_size: The size of a chunk of data whenever iterating (in
-                       bytes). This must be a multiple of 256 KB per the API
-                       specification.
+    :param chunk_size:
+        (Optional) The size of a chunk of data whenever iterating (in bytes).
+        This must be a multiple of 256 KB per the API specification.
 
     :type encryption_key: bytes
     :param encryption_key:
-        Optional 32 byte encryption key for customer-supplied encryption.
+        (Optional) 32 byte encryption key for customer-supplied encryption.
         See https://cloud.google.com/storage/docs/encryption#customer-supplied.
 
     :type kms_key_name: str
     :param kms_key_name:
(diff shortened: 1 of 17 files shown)


HemangChothani:
@googlebot I signed it!

google/cloud/storage/blob.py
```diff
@@ -193,6 +194,24 @@ def __init__(
...
+    @property
+    def name(self):
+        """ Get the blob's name.
+
+        :rtype: str
+        :returns: The name for the blob object.
+        """
+        return self._name
+
+    @name.setter
```
frankyn: Recommend removing setter.

google/cloud/storage/bucket.py
```diff
@@ -512,10 +512,29 @@ def __init__(self, client, name=None, user_project=None):
...
+    def name(self):
+        """ Get the bucket's name.
+
+        :rtype: str
+        :returns: The name for the bucket object.
+        """
+        return self._name
+
+    @name.setter
+    def name(self, value):
```
frankyn: Here as well

google/cloud/storage/blob.py
HemangChothani: @frankyn  If i remove setter then it throws an error like `AttributeError: can't set attribute`

frankyn: Is this why you recommended only adding pydocs instead? If so, then let's do that instead.

HemangChothani: @frankyn  Yes , that's why i recommend doc static doc string. i will update the PR. Thanks.

google/cloud/storage/bucket.py
```diff
@@ -505,13 +505,18 @@ class Bucket(_PropertyMixin):
...
+            Get the bucket's name.
+        """
         name = _validate_name(name)
         super(Bucket, self).__init__(name=name)
         self._client = client
         self._acl = BucketACL(self)
         self._default_object_acl = DefaultObjectACL(self)
         self._label_removals = set()
         self._user_project = user_project
+        self._name = name
```
frankyn, line 519: is this still necessary?

PR #95, opened 2020-04-01
fix(storage): use OrderedDict while encoding POST policy

IlyaFaer:
Towards https://github.com/googleapis/python-storage/pull/64#issuecomment-607430718
Sometimes policy losses it's order while encoding into JSON. Adding `OrderedDict()` to fix this.

### DIFF
--- a/google/cloud/storage/client.py
+++ b/google/cloud/storage/client.py
@@ -16,6 +16,7 @@
 
 import base64
 import binascii
+import collections
 import datetime
 import functools
 import json
@@ -972,7 +973,14 @@ class Client(ClientWithProject):
 
         # encode policy for signing
         policy = json.dumps(
-            {"conditions": conditions, "expiration": policy_expires.isoformat() + "Z"},
+            collections.OrderedDict(
+                sorted(
+                    {
+                        "conditions": conditions,
+                        "expiration": policy_expires.isoformat() + "Z",
+                    }.items()
+                )
+            ),
             separators=(",", ":"),
         )
         str_to_sign = base64.b64encode(policy.encode("utf-8"))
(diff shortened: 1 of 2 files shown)


google/cloud/storage/client.py
```diff
@@ -972,7 +973,14 @@ def generate_signed_post_policy_v4(
...
         policy = json.dumps(
-            {"conditions": conditions, "expiration": policy_expires.isoformat() + "Z"},
+            collections.OrderedDict(
+                sorted(
+                    {
+                        "conditions": conditions,
+                        "expiration": policy_expires.isoformat() + "Z",
+                    }.items()
+                )
+            ),
```
IlyaFaer, line 983: `expiration` sometimes encoded before `conditions` - in such a cases tests are failing

tests/unit/test_client.py
```diff
@@ -1762,6 +1762,10 @@ def test_conformance_post_policy(test_data):
                     scheme=in_data.get("scheme"),
                 )
     fields = policy["fields"]
+    out_data = test_data["policyOutput"]
+
+    decoded_policy = base64.b64decode(fields["policy"]).decode("unicode_escape")
+    assert decoded_policy == out_data["expectedDecodedPolicy"]
```
IlyaFaer, line 1768: Moved `decoded_policy` assert to the top: if something gone wrong in policy, it'll be easier to see it in decoded state - probably will save some time on debugging

IlyaFaer:
@frankyn, kokoro faild in `TestStorageCompose.test_compose_create_new_blob_wo_content_type`

google/cloud/storage/client.py
frankyn: I'm a little concerned about imposing order in signing code given it's only used for conformance tests.

However, it doesn't change the outcome of expectation and will work as expected as a user.

PR #98, opened 2020-04-03
Add validation of generation and metageneration to upload functions

### LINKED ISSUE #16: Storage: upload_from_string() with ifGenerationMatch=0
The GCS HTTP protocol -- but not the Python API -- has the ability to set `ifGenerationMatch` when [creating a storage object](https://cloud.google.com/storage/docs/json_api/v1/objects/insert#parameters):

> Makes the operation conditional on whether the object's current generation matches the given value. Setting to 0 makes the operation succeed only if there is a live version of the object.

**Why it's useful:** With this feature, the client could create a directory placeholder entry (a 0-byte object with a name ending in `'/'`) very efficiently like this:

    blob = bucket.blob('path/to/my/subdirectory/')
    blob.upload_from_string(b'', if_generation_match=0)

That one round trip creates the directory placeholder entry if it doesn't already exist. The alternatives are to first make a round trip to check if the entry exists or else to let the bucket accumulate identical placeholder entries (esp. for top level directories) by blindly creating them. [<strike>Or does GCS check if an uploaded object matches the current generation and optimize that case?</strike> -- Nope.]

**Why that matters:** Directory placeholders speed up [gcsfuse](https://github.com/GoogleCloudPlatform/gcsfuse) by an order of magnitude. Without the placeholders, you have to use `gcsfuse` in `--implicit-dirs` mode, and such a mount is frustratingly slow for interactive work. E.g. it takes several seconds just to list a tiny directory containing 2 files. With the placeholders, you can run `gcsfuse` without `--implicit-dirs`, and that mount lists directories in a tenth of a second or two.

**Proposal:** I could create a Pull Request adding this feature if you like, with either the specific `if_generation_match` query parameter or a way to pass in additional query parameters.

Another alternative is recommend that callers do something like subclass `Blob` to override `_add_query_parameters()` to add the `if_generation_match=0` name-value pair. That's ugly and fragile.

Is there a way to do this that I'm missing? Are there better alternatives?

HollayHorvath:
- [x] An issue was already open for this problem, linked below
- [ ] Code coverage does not decrease - I'm not quite sure about this
- [x] Appropriate docs were updated - I have updated the comments by adding the new parameters, if there are other places to update please let me know

Fixes #16  🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -180,10 +180,9 @@ class Blob(_PropertyMixin):
         self.chunk_size = chunk_size  # Check that setter accepts value.
         self._bucket = bucket
         self._acl = ObjectACL(self)
-        if encryption_key is not None and kms_key_name is not None:
-            raise ValueError(
-                "Pass at most one of 'encryption_key' " "and 'kms_key_name'"
-            )
+        _raise_for_more_than_one_none(
+            encryption_key=encryption_key, kms_key_name=kms_key_name,
+        )
 
         self._encryption_key = encryption_key
 
@@ -966,7 +965,17 @@ class Blob(_PropertyMixin):
         return headers, object_metadata, content_type
 
     def _do_multipart_upload(
-        self, client, stream, content_type, size, num_retries, predefined_acl
+        self,
+        client,
+        stream,
+        content_type,
+        size,
+        num_retries,
+        predefined_acl,
+        generation_match,
+        generation_not_match,
+        metageneration_match,
+        metageneration_not_match,
     ):
         """Perform a multipart upload.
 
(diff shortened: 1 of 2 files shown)


HollayHorvath:
@googlebot I signed it!

google/cloud/storage/blob.py
```diff
@@ -966,7 +965,17 @@ def _get_upload_arguments(self, content_type):
...
     def _do_multipart_upload(
-        self, client, stream, content_type, size, num_retries, predefined_acl
+        self,
+        client,
+        stream,
+        content_type,
+        size,
+        num_retries,
+        predefined_acl,
+        generation_match,
```
frankyn, line 975: Could you update the name with expected prefix of `if_` to follow the API spec?

tests/unit/test_blob.py
```diff
@@ -1785,6 +1897,10 @@ def _upload_from_file_helper(self, side_effect=None, **kwargs):
         content_type = u"font/woff"
         client = mock.sentinel.client
         predefined_acl = kwargs.get("predefined_acl", None)
+        generation_match = kwargs.get("generation_match", None)
```
frankyn, line 1900: One more request, could you add system tests that exercise these new parameters?

frankyn:
Thanks for your support @HollayHorvath, you helped get this ball moving and will have it out in the library soon. 

@IlyaFaer is taking the work you've done and making necessary changes to get into the library. I'm closing out this PR and continuing the remaining discussions in #111.

PR #103, opened 2020-04-10
docs(storage): fix docs for new sphinx version

### LINKED ISSUE #100: Building the docs breaks
I noticed that building the docs locally on master fails due to the following error:

`sphinx.errors.SphinxWarning: /python-storage/docs/index.rst:27:toctree contains reference to document 'changelog' that doesn't have a title: no link will be generated`

It turns out that the issue does not occur with the older version of Sphinx (< 3.x), thus it appears as a bug in the Sphinx package itself. A workaround is to pin that dependency to an older version.

/cc @frankyn @crwilcox

HemangChothani:
Fixes #100

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -38,6 +38,7 @@ extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.todo",
     "sphinx.ext.viewcode",
+    "recommonmark",
 ]
 
 # autodoc/autosummary flags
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -49,10 +50,6 @@
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
 
-# Allow markdown includes (so releases.md can include CHANGLEOG.md)
-# http://www.sphinx-doc.org/en/master/markdown.html
-source_parsers = {".md": "recommonmark.parser.CommonMarkParser"}
```
frankyn, line 54: Could you tell me more about this change? I see it's blocking your other PR.

docs/conf.py
```diff
@@ -38,6 +38,7 @@
     "sphinx.ext.napoleon",
     "sphinx.ext.todo",
     "sphinx.ext.viewcode",
+    "recommonmark",
```
frankyn, line 41: Found relevant information. Thanks for your patience on this PR.

https://github.com/readthedocs/recommonmark

```
# for Sphinx-1.4 or newer
extensions = ['recommonmark']

# for Sphinx-1.3
from recommonmark.parser import CommonMarkParser

source_parsers = {
    '.md': CommonMarkParser,
}

source_suffix = ['.rst', '.md']
```

PR #104, opened 2020-04-13
refactor(storage): deprecate preserve_acl arg in copy_blob() method

### LINKED ISSUE #37: Storage: copy_blob doesn't respect preserve_acl

#### Environment details

* macOS 10.15
* Python 3.6.5
* google-cloud-storage==1.20.0

#### Steps to reproduce

  1. Upload a blob with a predefined ACL of publicRead
  2. Copy this blob with preserve_acl set to true
  3. Expected result is that the new blob is set to publicRead, but this ACL isn't actually preserved
  4. Calling `make_public()` on the new blob correct sets it to publicRead

#### Code example

```python
file_obj = BytesIO(...)
blob.upload_from_file(file_obj, content_type='...', predefined_acl='publicRead')

bucket = storage_client.get_bucket(...)
bucket.copy_blob(blob, destination_bucket=bucket, new_name='new-name.jpg', preserve_acl=True)
# At this point, the blob will not be set to public
```

IlyaFaer:
Closes #37

### DIFF
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -505,6 +505,10 @@ class Bucket(_PropertyMixin):
     """Allowed values for :attr:`location_type`."""
 
     def __init__(self, client, name=None, user_project=None):
+        """
+        property :attr:`name`
+            Get the bucket's name.
+        """
         name = _validate_name(name)
         super(Bucket, self).__init__(name=name)
         self._client = client
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -1268,7 +1268,8 @@ class Bucket(_PropertyMixin):
                        to the ``client`` stored on the current bucket.
 
         :type preserve_acl: bool
-        :param preserve_acl: (Optional) Copies ACL from old blob to new blob.
+        :param preserve_acl: DEPRECATED. This argument is not functional!
+                             (Optional) Copies ACL from old blob to new blob.
                              Default: True.
 
         :type source_generation: long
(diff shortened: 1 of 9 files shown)


google/cloud/storage/bucket.py
```diff
@@ -1307,8 +1307,13 @@ def copy_blob(
             timeout=timeout,
         )
 
-        if not preserve_acl:
-            new_blob.acl.save(acl={}, client=client, timeout=timeout)
```
IlyaFaer, line 1311: As "preserve acl" behavior was never supported by backend, these line doesn't make sense - deleting them

tests/unit/test_bucket.py
```diff
@@ -1144,29 +1144,31 @@ def test_copy_blobs_preserve_acl(self):
...
+        with mock.patch("warnings.warn") as warn:
+            new_blob = source.copy_blob(
+                blob, dest, NEW_NAME, client=client, preserve_acl=False
+            )
+            warn.assert_called_once_with(
+                "preserve_acl arg is deprecated and will be removed in future."
+                "Do a subsequent ACL update instead.",
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
```
IlyaFaer, line 1156: The only test for `preserve_acl` rewritten to check if warning was pushed.

IlyaFaer:
> I'd prefer to have a user do a subsequent ACL update in this case or that we have a FR to introduce a parameter to supply metadata when using copy_blob().

As for the feature for copying ACLs, I suppose, we could use `destinationPredefinedAcl` option, as we're using it in other methods with `predefined_acl` arg, for example `Blob.upload_from_file()`.

google/cloud/storage/bucket.py
```diff
@@ -1244,7 +1244,7 @@ def copy_blob(
         destination_bucket,
         new_name=None,
         client=None,
-        preserve_acl=True,
+        preserve_acl=None,
```
frankyn, line 1247: Leave default value alone, instead update documentation that this parameter is not functional and label it deprecated. Do not emit a warning only add the deprecation documentation. Additionally, provide a snippet of how to copy ACLs over from one object to another with the existing client.

PR #107, opened 2020-04-15
fix(storage): anonymous credentials for private bucket

### LINKED ISSUE #102: "Anonymous credentials cannot be refreshed" with fake-gcs-server
- OS type and version: Ubuntu 18.04
  - Python version: 3.7.6
  - pip version: 20.0.2
  - `google-cloud-storage` version: 1.27.0

I'm trying to use [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) for local GCS testing.

Basic client setup works fine:
```python
>>> creds = AnonymousCredentials() 
>>> options = ClientOptions(api_endpoint="http://127.0.0.1:4443") 
>>> client = gcs.Client(credentials=creds, client_options=options, project='') 
```

I can list buckets and blobs:
```python
>>> print(list(client.list_buckets()))                                                                                                                                      
[<Bucket: bucket1>, <Bucket: bucket2>]
>>> print(list(client.list_blobs('bucket1')))                                                                                                                               
[<Blob: bucket1, hello.txt, 1586453246295809>,
 <Blob: bucket1, subdir/toto.txt, 1586453246295810>]
>>> a, b = client.list_blobs('bucket1')                                                                                                                                     
>>> a.size                                                                                                                                                                  
548
>>> a.name                                                                                                                                                                  
'hello.txt'
>>> a.id                                                                                                                                                                    
'bucket1/hello.txt'
```

`fake-gcs-server` doesn't need any credentials, so reading the file should just work. But for some reason it doesn't:
```python
>>> a.download_as_string()                                                                                                                                                  
  File "<ipython-input-19-72ec41b6efbf>", line 1, in <module>
    a.download_as_string()
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.7/site-packages/google/cloud/storage/blob.py", line 879, in download_as_string
    raw_download=raw_download,
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.7/site-packages/google/cloud/storage/blob.py", line 796, in download_to_file
    transport, file_obj, download_url, headers, start, end, raw_download
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.7/site-packages/google/cloud/storage/blob.py", line 720, in _do_download
    download.consume(transport)
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.7/site-packages/google/resumable_media/requests/download.py", line 151, in consume
    result = _helpers.http_request(transport, method, url, **request_kwargs)
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.7/site-packages/google/resumable_media/requests/_helpers.py", line 136, in http_request
    return _helpers.wait_and_retry(func, RequestsMixin._get_status_code, retry_strategy)
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.7/site-packages/google/resumable_media/_helpers.py", line 150, in wait_and_retry
    response = func()
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.7/site-packages/google/auth/transport/requests.py", line 482, in request
    self.credentials.refresh(auth_request)
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.7/site-packages/google/auth/credentials.py", line 148, in refresh
    raise ValueError("Anonymous credentials cannot be refreshed.")
ValueError: Anonymous credentials cannot be refreshed.
```

HemangChothani:
Fixes #102 

- We can override the private variables of blob class related to api_endpoint, upload_uri, download_uri.

- Still confusion with these two parameters:
     1. blob._API_ACCESS_ENDPOINT 
     2. bucket._API_ACCESS_ENDPOINT


google/cloud/storage/client.py
```diff
@@ -122,6 +124,7 @@ def __init__(
             if client_options.api_endpoint:
                 api_endpoint = client_options.api_endpoint
                 kw_args["api_endpoint"] = api_endpoint
+                _override_private_url_variables(api_endpoint)
```
frankyn, line 127: Thanks for putting this together. I think the code needs to be refactored for maintainability but the this is a good start.

The URLs in URL Templates could be refactored to be pulled in from the client which is passed around the library.

For example:
https://github.com/googleapis/python-storage/blob/a50cdd145d39cb67e0d7021c376acbe376436178/google/cloud/storage/blob.py#L1020

Could be updated to:
```
_BASE_UPLOAD_TEMPLATE =  u"{hostname}/upload/storage/v1{bucket_path}/o?uploadType="
_MULTIPART_URL_TEMPLATE = _BASE_UPLOAD_TEMPLATE + u"multipart"
# truncated....
base_url = _MULTIPART_URL_TEMPLATE.format(hostname=storage_client._connection.API_BASE_URL,bucket_path=self.bucket.path)
```

WDYT?

google/cloud/storage/client.py
```diff
@@ -1056,3 +1059,14 @@ def _item_to_hmac_key_metadata(iterator, item):
...
+
+def _override_private_url_variables(api_endpoint):
+    blob._DOWNLOAD_URL_TEMPLATE = api_endpoint + u"/download/storage/v1{path}?alt=media"
+    blob._BASE_UPLOAD_TEMPLATE = (
+        api_endpoint + u"/upload/storage/v1{bucket_path}/o?uploadType="
+    )
+    blob._MULTIPART_URL_TEMPLATE = blob._BASE_UPLOAD_TEMPLATE + u"multipart"
+    blob._RESUMABLE_URL_TEMPLATE = blob._BASE_UPLOAD_TEMPLATE + u"resumable"
+    blob._API_ACCESS_ENDPOINT = api_endpoint
+    bucket._API_ACCESS_ENDPOINT = api_endpoint
```
frankyn, line 1072: These URLs are used for Signature based requests like Signed URLs. These can be left alone because the method has a way to update the hostname because they don't use the same API (Download and Upload use JSON while Signature requests use XML API).

HemangChothani:
@frand still we can refactor, need your suggestion.

-  We can remove [_get_transport_method()](https://github.com/googleapis/python-storage/blob/cecc7ac95a00ab1437a24bee44f8b6c5b8a7189d/google/cloud/storage/blob.py#L638) from blob class as we are calling [_require_client()](https://github.com/googleapis/python-storage/pull/107/files#diff-5d3277a5f0f072a447a1eb89f9fa1ae0R796) method just above  _get_transport_method() call, so no need of that method and we can also directly pass client._http.

OR

- We can modify return type of [_get_transport_method](https://github.com/googleapis/python-storage/blob/cecc7ac95a00ab1437a24bee44f8b6c5b8a7189d/google/cloud/storage/blob.py#L638) method, as of now it returns transport object but also need to return client object as well, so no need to call [_require_client()](https://github.com/googleapis/python-storage/pull/107/files#diff-5d3277a5f0f072a447a1eb89f9fa1ae0R796) in the respective method where we are using `*_URL_TEMPLATE` global variables.

google/cloud/storage/blob.py
```diff
@@ -650,19 +648,24 @@ def _get_transport(self, client):
         client = self._require_client(client)
         return client._http
 
-    def _get_download_url(self):
+    def _get_download_url(self, client):
```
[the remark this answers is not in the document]
frankyn: this was a bad call, you'll need to do this along with `_require_client()` otherwise you can end up in a bad state.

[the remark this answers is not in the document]
HemangChothani: `_require_client()` has been call by [_get_transport](https://github.com/googleapis/python-storage/blob/cecc7ac95a00ab1437a24bee44f8b6c5b8a7189d/google/cloud/storage/blob.py#L638) method so moving `_get_download_url` method call after that.  

One system test has failed so fixed that.

google/cloud/storage/client.py
```diff
@@ -530,7 +530,7 @@ def download_blob_to_file(self, blob_or_uri, file_obj, start=None, end=None):
         try:
             blob_or_uri.download_to_file(file_obj, client=self, start=start, end=end)
         except AttributeError:
-            blob = Blob.from_string(blob_or_uri)
+            blob = Blob.from_string(blob_or_uri, client=self)
```
frankyn, line 533: was this an accidental miss in implementation? Seems like it but want to confirm.

frankyn: Looking at it again, client is passed below to download_to_file(). The change doesn't look necessary.

HemangChothani: Yes, but the problem arise here , if i didn't pass the client so bucket class has received `None`
https://github.com/googleapis/python-storage/blob/18b1e07d66b6dfde2de5f1d97a09ab0f25db1d1b/google/cloud/storage/blob.py#L351

 so when call  `_require_client()` method it find the client object which  passed in` download_to_file()` so `hostname=self.client._connection.API_BASE_URL` line hit the error as self.client found `None`

frankyn: Gotcha, when I raised that you should revert my last comment, I was referring to updating the method to accept the client because even though `_get_transport` calls `_require_client()` it doesn't update `self.client`.

I think you may want to pass client returned from `_require_client()` to `_get_download_url()`.

Thank you for your patience.

frankyn:
@HemangChothani asked @crwilcox to take another pass to confirm.

PR #111, opened 2020-04-21
feat(storage): add arguments for *GenerationMatch uploading options

### LINKED ISSUE #16: Storage: upload_from_string() with ifGenerationMatch=0

IlyaFaer:
These changes were originally made in another [PR](https://github.com/googleapis/python-storage/pull/98). Here is the version with changes requested by reviews.
Closes #16

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -185,10 +185,9 @@ class Blob(_PropertyMixin):
         self.chunk_size = chunk_size  # Check that setter accepts value.
         self._bucket = bucket
         self._acl = ObjectACL(self)
-        if encryption_key is not None and kms_key_name is not None:
-            raise ValueError(
-                "Pass at most one of 'encryption_key' " "and 'kms_key_name'"
-            )
+        _raise_for_more_than_one_none(
+            encryption_key=encryption_key, kms_key_name=kms_key_name,
+        )
 
         self._encryption_key = encryption_key
 
@@ -974,7 +973,17 @@ class Blob(_PropertyMixin):
         return headers, object_metadata, content_type
 
     def _do_multipart_upload(
-        self, client, stream, content_type, size, num_retries, predefined_acl
+        self,
+        client,
+        stream,
+        content_type,
+        size,
+        num_retries,
+        predefined_acl,
+        if_generation_match,
+        if_generation_not_match,
+        if_metageneration_match,
+        if_metageneration_not_match,
     ):
         """Perform a multipart upload.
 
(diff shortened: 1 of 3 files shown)


google/cloud/storage/blob.py
```diff
@@ -1007,6 +1016,18 @@ def _do_multipart_upload(
...
+        :param if_generation_match: (Optional) Precondition control if generation match
+
+        :type if_generation_not_match: long
+        :param if_generation_not_match: (Optional) Precondition control if generation not match
+
+        :type if_metageneration_match: long
+        :param if_metageneration_match: (Optional) Precondition control if metageneration match
+
+        :type if_metageneration_not_match: long
+        :param if_metageneration_not_match: (Optional) Precondition control if metageneration not match
```
frankyn, line 1029: Missed in first pass @IlyaFaer, could update documentation to reflect what's stated in cloud storage API reference documentation:

https://cloud.google.com/storage/docs/json_api/v1/objects/insert

IlyaFaer: Okay, pushed the comments very similar to those in docs

1fish2:
👍 Thanks to @HollayHorvath and @frankyn!

The code to create a directory placeholder changes from

```python
blob = bucket.blob(name)
if not blob.exists():
    blob.upload_from_string(b'', content_type=OCTET_STREAM)
```

to

```python
blob = bucket.blob(name)
try:
    blob.upload_from_string(
        b'', content_type=OCTET_STREAM, if_generation_match=0)
except PreconditionFailed:
    pass
```

because, as the doc says, "Setting to 0 makes the operation succeed only if there are no live versions of the blob."

In a handful of timings, the new code saved more than half the didn't-exist time and a bit of the already-exists time (although that might be measurement noise).

PR #122, opened 2020-04-30
feat: add *generation*match args into Blob.compose()

IlyaFaer:
Add `ifGenerationMatch` and `ifMetagenerationMatch` arguments into `Blob.compose()` method.
Towards #127

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -2038,14 +2038,13 @@ class Blob(_PropertyMixin):
         :param if_generation_match: (Optional) Make the operation conditional on whether
                                     the blob's current generation matches the given value.
                                     Setting to 0 makes the operation succeed only if there
-                                    are no live versions of the blob. Generation numbers list
-                                    should have the same order that ``sources`` have.
+                                    are no live versions of the blob. The list must match
+                                    ``sources`` item-to-item.
 
         :type if_metageneration_match: list of long
-        :param if_metageneration_match: (Optional) Make the operation conditional on whether the
-                                        blob's current metageneration matches the given value.
-                                        Metageneration numbers list should have the same order
-                                        that ``sources`` have.
+        :param if_metageneration_match: (Optional) Make the operation conditional on whether
+                                        the blob's current metageneration matches the given
+                                        value. The list must match ``sources`` item-to-item.
 
         Example:
             Compose blobs using generation match preconditions.
@@ -2056,7 +2055,7 @@ class Blob(_PropertyMixin):
 
             >>> blobs = [bucket.blob("blob-name-1"), bucket.blob("blob-name-2")]
             >>> if_generation_match = [None] * len(blobs)
-            >>> if_generation_match[0] = "123"
+            >>> if_generation_match[0] = "123"  # precondition for "blob-name-1"
 
             >>> composed_blob = bucket.blob("composed-name")
             >>> composed_blob.compose(blobs, if_generation_match)
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -2046,7 +2046,35 @@ class Blob(_PropertyMixin):
                                         blob's current metageneration matches the given value.
                                         Metageneration numbers list should have the same order
                                         that ``sources`` have.
+
+        Example:
+            Compose blobs using generation match preconditions.
+
+            >>> from google.cloud import storage
+            >>> client = storage.Client()
+            >>> bucket = client.bucket('bucket-name')
+
+            >>> blobs = [bucket.blob("blob-name-1"), bucket.blob("blob-name-2")]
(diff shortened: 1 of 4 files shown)


google/cloud/storage/blob.py
```diff
@@ -2006,34 +2006,71 @@ def make_private(self, client=None):
...
         if self.user_project is not None:
             query_params["userProject"] = self.user_project
 
+        source_objects = []
+        for index, source in enumerate(sources):
+            source_object = {"name": source.name}
+
+            preconditions = {}
+            if if_generation_match is not None:
+                preconditions["ifGenerationMatch"] = if_generation_match[index]
```
andrewsg, line 2062: This pattern will error out with IndexError and a stacktrace internal to our code if the user supplies a list that is too short. Perhaps we should catch the IndexError or check the length in advance, and error out with a ValueError and a useful error message?

tests/system/test_system.py
```diff
@@ -1234,6 +1234,27 @@ def test_compose_replace_existing_blob(self):
         composed = original.download_as_string()
         self.assertEqual(composed, BEFORE + TO_APPEND)
 
+    def test_compose_with_generation_match(self):
```
andrewsg, line 1237: If I'm not mistaken, the behavior of this test would be identical if the generation match and metageneration match flags were not set. Can we additionally test for the case of a mismatch and verify that the flags are doing something (causing the request to error out, presumably)?

google/cloud/storage/blob.py
IlyaFaer: Yeah, I had thoughts on that, not sure about the best solution. Raising an error in case of different lengths makes sense, but what if user will try to compose 10 blobs? For example, he wanted to use concrete generation for one blob only - in this case he'll have to write `None` for nine others. I've tried to deal with this problem using `iter()` and `next()` in [another related PR](https://github.com/googleapis/python-storage/pull/130#discussion_r420131309), PTAL. Maybe we should do something similar here? (I was going to, actually, but we got on pause)

@frankyn, what do you think?

frankyn: I would provide the following example:

# Option 1
```python

bucket = storage_client.bucket('anima-frank')

list_of_blobs = [bucket.blob('name'), bucket.bloc('name)]
if_generation_match_list = [None] * len(list_of_blobs)
if_generation_match_list[0] = `123`

composed_blob = bucket.blob('composed-name)
composed_blob.compse(list_of_blobs, if_generation_match)
```

It still feels clunky, what about instead doing something like:

# Option 2
```python
list_of_blobs = [{'blob': bucket.blob('name'), bucket.bloc('name), 'if_generation_match': 123), bucket.blob('name2')]


composed_blob = bucket.blob('composed-name)
composed_blob.compse(list_of_blobs)
```

If the element is a dict() then get values?

IlyaFaer: Oh, I thought mix-typed args are usually off the table. But I definitely like the idea

I've done everything locally, but I suppose I'll hold pushing until we merge #123, as I'm using `_add_generation_match_parameters()` in new version, which will be added with pt1. I'd like not to merge #123 into this PR - it'll be +1.5 thousands lines.

image

frankyn: Thanks @IlyaFaer, let's hold on this PR until early next week to get @crwilcox's feedback. We want to be careful with this surface because it's unlike others in the library. 

Thank you for your patience on this.

andrewsg: Hi Ilya,

On discussion we've decided to go ahead with option 1. I've cleared up some ambiguities; see example of UI:

```storage_client = storage.Client()

bucket = storage_client.bucket('anima-frank')

list_of_blobs = [bucket.blob('name1'), bucket.blob('name2')]
if_generation_match_list = [None] * len(list_of_blobs)
if_generation_match_list[0] = `123`

composed_blob = bucket.blob('composed-name)
composed_blob.compse(list_of_blobs, if_generation_match)```

Comment (IlyaFaer) [reply google/cloud/storage/blob.py:2093]: Got it. Added an example and error raising. Also I've added one more condition to avoid sending generation match parameters with `None` values (added unit test to check this case)

File: google/cloud/storage/blob.py [hunk 3]
```diff
@@ -2006,34 +2006,105 @@ def make_private(self, client=None):
...
                        to the ``client`` stored on the blob's bucket.
+
         :type timeout: float or tuple
         :param timeout: (Optional) The amount of time, in seconds, to wait
             for the server response.
 
             Can also be passed as a tuple (connect_timeout, read_timeout).
             See :meth:`requests.Session.request` documentation for details.
+
+        :type if_generation_match: list of long
```
andrewsg, line 2037: The example code is helpful, but please also explain briefly in the docstring that the list of long here is an index-for-index match to the list of blobs.

PR #123, opened 2020-05-01
feat: add if*generation*Match support, pt1

IlyaFaer:
Add `if_generation_match`, `if_generation_not_match`, `if_metageneration_match` and `if_metageneration_not_match` arguments into methods:

`Client()`: `lookup_bucket()`, `get_bucket()`

`Bucket()`: `get_blob()`, `delete_blob()`, `delete()`, `exists()`, `patch()`, `reload()`, `update()`

`Blob()`: `delete()`, `download_to_file()`, `download_to_filename()`, `download_as_string()`, `rewrite()`, `update_storage_class()`, `exists()`, `patch()`, `reload()`, `update()`, `reload()`

Towards #127


google/cloud/storage/_helpers.py
```diff
@@ -312,3 +402,44 @@ def _convert_to_timestamp(value):
...
+    """
+    for snakecase_name, camelcase_name in _GENERATION_MATCH_PARAMETERS:
+        value = parameters.get(snakecase_name)
+
+        if value is not None:
+            if isinstance(name_value_pairs, list):
+                name_value_pairs.append((camelcase_name, value))
+
+            elif isinstance(name_value_pairs, dict):
+                name_value_pairs[camelcase_name] = value
```
IlyaFaer, line 424: Added one more condition into helper, as there are methods in which `query_params` is `dict`, not `list`.

google/cloud/storage/_helpers.py
```diff
@@ -312,3 +402,44 @@ def _convert_to_timestamp(value):
...
+        escaped_keys = ["'%s'" % name for name in kwargs.keys()]
+
+        keys_but_last = ", ".join(escaped_keys[:-1])
+        last_key = escaped_keys[-1]
+
+        msg = "Pass at most one of {keys_but_last} and {last_key}".format(
+            keys_but_last=keys_but_last, last_key=last_key
+        )
+
+        raise ValueError(msg)
```
IlyaFaer, line 445: Moved into `_helpers`. Seems to me the name of the helper is not very correct. It raises an error in case if more than one arg is **not** `None`. Maybe should be renamed into `_raise_for_more_than_one_set()`?

tests/unit/test_client.py
```diff
@@ -436,6 +436,54 @@ def test_get_bucket_with_string_hit(self):
...
+        URI2 = "/".join(
+            [
+                client._connection.API_BASE_URL,
+                "storage",
+                client._connection.API_VERSION,
+                "b",
+                "%s?ifMetagenerationMatch=%s&projection=noAcl"
+                % (BUCKET_NAME, METAGENERATION_NUMBER),
+            ]
+        )
```
IlyaFaer, line 467: URI params order is not guaranted, so checking both variant. The alternative here is to add sorting somewhere in `JSONConnection.build_api_url()`

google/cloud/storage/_helpers.py
```diff
@@ -312,3 +402,44 @@ def _convert_to_timestamp(value):
...
+    """
+    for snakecase_name, camelcase_name in _GENERATION_MATCH_PARAMETERS:
+        value = match_parameters.get(snakecase_name)
+
+        if value is not None:
+            if isinstance(parameters, list):
+                parameters.append((camelcase_name, value))
+
+            elif isinstance(parameters, dict):
+                parameters[camelcase_name] = value
```
IlyaFaer, line 424: Added one more condition into helper, as there are methods in which query_params is dict, not list with (key, value) pairs.

google/cloud/storage/_helpers.py
```diff
@@ -121,27 +129,75 @@ def _query_params(self):
...
+                                        blob's current metageneration matches the given value.
+
+        :type if_metageneration_not_match: long
+        :param if_metageneration_not_match: (Optional) Make the operation conditional on whether the
+                                            blob's current metageneration does not match the given value.
+
+        :raises: :class:`ValueError` if ``if_metageneration_match`` and
+                 ``if_metageneration_not_match`` are both set.
         """
+        _raise_for_more_than_one_none(
```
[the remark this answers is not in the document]
IlyaFaer: Done

google/cloud/storage/_helpers.py
andrewsg: May I suggest we use duck typing here and just try to append, catch a failure, and then use the dict access, in order to safely handle any list-alikes or dict-alikes that may sneak into the code later? Also, might be good to error out if neither the list or dict access works.

google/cloud/storage/_helpers.py
andrewsg: Good catch. I agree we should change the name. If you prefer not to change the name in this PR, please open an issue and either tackle it separately or assign it to me so it doesn't get lost.

IlyaFaer:
Lint session failed, but the problem is not related
```
Running session lint
Creating virtual environment (virtualenv) using python3.8 in .nox/lint
pip install flake8 black==19.10b0
black --check docs google tests noxfile.py setup.py
All done! ✨ 🍰 ✨
35 files would be left unchanged.
flake8 google tests
google/cloud/storage/_signing.py:156:13: F523 '...'.format(...) has unused arguments at position(s): 0
Command flake8 google tests failed with exit code 1
Session lint failed.
```

IlyaFaer:
I've added more system tests, and re-reviewed it all once again - fixed a couple of nits.

google/cloud/storage/blob.py
```diff
@@ -1540,12 +1753,12 @@ def upload_from_file(
...
             warnings.warn(_NUM_RETRIES_MESSAGE, DeprecationWarning, stacklevel=2)
 
-        _raise_for_more_than_one_none(
+        _raise_if_more_than_one_set(
             if_generation_match=if_generation_match,
             if_generation_not_match=if_generation_not_match,
         )
 
-        _raise_for_more_than_one_none(
+        _raise_if_more_than_one_set(
```
frankyn, line 1761: These should be removed.

IlyaFaer: Done

google/cloud/storage/_helpers.py
frankyn: Just double checking, but @andrewsg are you okay with how this was resolved?

andrewsg: Yes, adding the "else" clause works for me.

PR #125, opened 2020-05-01
feat: add offset and includeTrailingPrefix options to list_blobs

andrewsg:


tests/system/test_system.py
```diff
@@ -909,6 +942,17 @@ def test_third_level(self):
...
+    @RetryErrors(unittest.TestCase.failureException)
+    def test_include_trailing_delimiter(self):
+        iterator = self.bucket.list_blobs(
+            delimiter="/", include_trailing_delimiter=True
+        )
+        page = six.next(iterator.pages)
+        blobs = list(page)
+        self.assertEqual([blob.name for blob in blobs], ["file01.txt", "parent/"])
+        self.assertIsNone(iterator.next_page_token)
+        self.assertEqual(iterator.prefixes, set(["parent/"]))
```
frankyn, line 954: how is this request different when `include_trailinig_delimitier=False`? I don't see the difference between the system test in this case.

andrewsg: The difference is that this is the only name that includes `parent/` in `self.assertEqual([blob.name for blob in blobs], ["file01.txt", "parent/"])`.

tests/system/test_system.py
```diff
@@ -825,11 +825,42 @@ def test_paginate_files(self):
...
+
+        page1 = six.next(page_iter)
+        blobs = list(page1)
+        self.assertEqual(len(blobs), count)
+        self.assertEqual(blobs[0].name, desired_files[0])
+        self.assertIsNotNone(iterator.next_page_token)
+        # Technically the iterator is exhausted.
+        self.assertEqual(iterator.num_results, iterator.max_results)
+        # But we modify the iterator to continue paging after
+        # articially stopping after ``count`` items.
```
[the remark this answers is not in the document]
andrewsg: fixed

tests/system/test_system.py
andrewsg: I'll leave this for you to resolve if you're satisfied with that explanation

PR #137, opened 2020-05-07
feat: add helper for bucket bound hostname URLs

### LINKED ISSUE #121: Move bucket_bound_hostname code into helper
For now we have three places in which we're using `bucket_bound_hostname`. The code is almost the same in all three places, so it'll be good to union it in single helper:
https://github.com/googleapis/python-storage/blob/616f1079e2c8b5a4c900f663a9688da64e2bc1b0/google/cloud/storage/blob.py#L517-L522

https://github.com/googleapis/python-storage/blob/616f1079e2c8b5a4c900f663a9688da64e2bc1b0/google/cloud/storage/bucket.py#L2531-L2536

https://github.com/googleapis/python-storage/blob/616f1079e2c8b5a4c900f663a9688da64e2bc1b0/google/cloud/storage/client.py#L1014-L1020

IlyaFaer:
Closes #121


google/cloud/storage/_helpers.py
```diff
@@ -312,3 +312,25 @@ def _convert_to_timestamp(value):
...
+    :type end_slash: bool
+    :param end_slash: (Optional) If True, add "/" slash to
+                      the end of the result URL.
+    """
+    if ":" in host:
+        return host
+
+    return "{scheme}://{host}{slash}".format(
+        scheme=scheme, host=host, slash="/" if end_slash else ""
+    )
```
IlyaFaer, line 336: In V4 POST policies we're using URLs with end slash (that's what expected in conformance tests), but in other places URLs doesn't add it (and if I'll add it, several tests will fail). Keeping this behavior in the helper with `end_slash` arg. As it's just an internal helper I think we can afford it.

frankyn: What is the difference between the failures in this case?

IlyaFaer: If we'll add a slash permanently:
`Test_Bucket.test_generate_signed_url_v4_w_bucket_bound_hostname_w_bare_hostname()` fails on assert:
```
AssertionError: expected call not found.
E         Expected: generate_signed_url_v4(None, resource='/', expiration=datetime.datetime(2020, 5, 15, 9, 28, 17, 120499, tzinfo=<UTC>), api_access_endpoint='http://cdn.example.com', method='GET', headers=None, query_parameters=None)
E           Actual: generate_signed_url_v4(None, resource='/', expiration=datetime.datetime(2020, 5, 15, 9, 28, 17, 120499, tzinfo=<UTC>), api_access_endpoint='http://cdn.example.com/', method='GET', headers=None, query_parameters=None)
```
`Test_Blob.test_generate_signed_url_v4_w_bucket_bound_hostname_w_bare_hostname()` fails on:
```
AssertionError: expected call not found.
E         Expected: generate_signed_url_v4(<object object at 0x000001515522AF50>, resource='/blob-name', expiration=datetime.datetime(2020, 5, 15, 9, 28, 15, 798333, tzinfo=<UTC>), api_access_endpoint='http://cdn.example.com', method='GET', content_md5=None, content_type=None, response_type=None, response_disposition=None, generation=None, headers=None, query_parameters=None, access_token=None, service_account_email=None)
E           Actual: generate_signed_url_v4(<object object at 0x000001515522AF50>, resource='/blob-name', expiration=datetime.datetime(2020, 5, 15, 9, 28, 15, 798333, tzinfo=<UTC>), api_access_endpoint='http://cdn.example.com/', method='GET', content_md5=None, content_type=None, response_type=None, response_disposition=None, generation=None, headers=None, query_parameters=None, service_account_email=None, access_token=None)
```
So, the `api_access_endpoint` doesn't match because of the end slash.

-

If we'll delete the end slash permanently, than `test_conformance_post_policy` fails with `POST Policy Simple Bucket Bound Hostname HTTP` and `'POST Policy Simple Bucket Bound Hostname'` test data:
```
assert policy["url"] == test_data["policyOutput"]["url"]
E       AssertionError: assert 'http://mydomain.tld' == 'http://mydomain.tld/'
E         - http://mydomain.tld/
E         ?                    -
E         + http://mydomain.tld
```

frankyn: I would consolidate the two cases and prefer conformance test format over api_access_endpoint. 

I don't want to have two paths if possible.

google/cloud/storage/_helpers.py
```diff
@@ -491,3 +491,22 @@ def _raise_if_more_than_one_set(**kwargs):
...
+    :param host: Host name.
+
+    :type scheme: str
+    :param scheme: (Optional) Web scheme. If passed, use it
+                   as a scheme in the result URL.
+
+    :rtype: str
+    :returns: A bucket bound hostname URL.
+    """
+    if ":" in host:
```
frankyn, line 509: This is brittle what about using urllib?

IlyaFaer: Yeah, that makes sense, pushed. I suppose, checking if scheme and netloc present is enough.
Maybe it worth adding a final check into the helper? I mean, if the given arg will give `False` for the first check `if url_parts.scheme and url_parts.netloc:`, it'll build a URL without any more checks. Maybe we should check the result before returning it?

IlyaFaer: Here is what I'm talking about:

Безымянный

As `bucket_bound_hostname` comes from user, we can get any garbage value in it (in theory). It probably should be checked before sending a real request into this URL.

frankyn: I don't understand what you mean by doing a check after returning it.

IlyaFaer: I mean check before returning. Like this:
```python
def _bucket_bound_hostname_url(host, scheme=None):
    url_parts = urlsplit(host)
        if url_parts.scheme and url_parts.netloc:
            return host

    url = "{scheme}://{host}/".format(scheme=scheme, host=host)
    check_url_is_correct(url)
    return url
```

IlyaFaer: Oh, I got it, on my screenshot there is a wrapper `_call_fut()`. I wanted to show you that if I'll pass:
`bucker_bound_hostname=":scheme:domain.tcl/"`
than the helper will return an invalid URL like: `"None://:scheme:domain.tcl"`. And this URL will be tried to be used in POST policy, or while building a signed URL. Which seems very wrong to me.

frankyn: Thanks, there should be a canonical way to verify that without writing the checker manually. Could you look into. I'll be out until next Tuesday and this is my last message this week. I'll followup early next week

IlyaFaer: Sure, I just wanted to show the place, wasn't actually going to write own validator.

After taking a closer look at the Python libs I don't see methods to validate a URL. There is a couple of functions, which are doing simple validation as a side effect (like `requests.models.PreparedRequest.prepare_url()`), and some side/unofficial tools (validators in Django framework, as an example), but I assume they are off the table. And titanic size regexes, of course.

I think I'm closing my own question as resolved, as I don't see reliable enough methods to validate a URL. Sorry for wasting your time.

PR #146, opened 2020-05-14
fix(storage): add projection parameter to blob.reload method

### LINKED ISSUE #136: Google cloud python API returns None for blob.owner attribute
I have created a Google bucket with "Fine-grained access control" and a few users have uploaded files to it.  Using the python API I can't seem to get any information on who uploaded each.  The blob.owner property just returns None:

    sclient = storage.Client(project=GCLOUD_PROJECT)
    bucket = storage.bucket.Bucket(client=sclient, name=GCLOUD_BUCKET)
    blob = bucket.get_blob('foo.bar')
    blob.reload()
    print(blob.owner)

I'm calling reload() there because the documentation states it's required to pull some attributes from the server.  All other properties I try print fine (size, updated, etag, md5_hash, etc.).

How can I recover the uploader identification?

#### Environment details

  - OS type and version:  Ubuntu 18.04
  - Python version: 3.7.0
  - pip version: 18.1
  - `google-cloud-storage` version: 1.20.0

#### Steps to reproduce

  1. Place file in google bucket
  2. Attempt to use API to get owner information from the blob

#### Code example

```    sclient = storage.Client(project=GCLOUD_PROJECT)
    bucket = storage.bucket.Bucket(client=sclient, name=GCLOUD_BUCKET)
    blob = bucket.get_blob('foo.bar')
    blob.reload()
    print(blob.owner)
```

#### Stack trace
Doesn't error - just returns None

HemangChothani:
Fixes #136

### DIFF
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -121,10 +121,15 @@ class _PropertyMixin(object):
             params["userProject"] = self.user_project
         return params
 
-    def reload(self, client=None, timeout=_DEFAULT_TIMEOUT):
+    def reload(self, projection="noAcl", client=None, timeout=_DEFAULT_TIMEOUT):
         """Reload properties from Cloud Storage.
 
         If :attr:`user_project` is set, bills the API request to that project.
+
+        :type projection: str
+        :param projection: (Optional) If used, must be 'full' or 'noAcl'.
+                           Defaults to ``'noAcl'``. Specifies the set of
+                           properties to return.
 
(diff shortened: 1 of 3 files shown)


google/cloud/storage/_helpers.py
```diff
@@ -135,6 +135,7 @@ def _query_params(self):
 
     def reload(
         self,
+        projection="noAcl",
```
frankyn, line 138: Can you move this parameter after the client parameter?

Issue #127, opened 2020-05-04
Support If*generation preconditions


frankyn:
@IlyaFaer has started on this work in the following PR's:
#122 (merged)
#123 (merged)
#130 (merged)
#141 (merged)

* [Bucket Delete](https://cloud.google.com/storage/docs/json_api/v1/buckets/delete): [google.cloud.storage.bucket.Bucket.delete](https://googleapis.dev/python/storage/latest/buckets.html#google.cloud.storage.bucket.Bucket.delete)

    * ifMetagenerationMatch

    * ifMetagenerationNotMatch

* [Bucket Get](https://cloud.google.com/storage/docs/json_api/v1/buckets/get): [google.cloud.storage.client.Client.get_bucket](https://googleapis.dev/python/storage/latest/client.html#google.cloud.storage.client.Client.get_bucket)

    * ifMetagenerationMatch

    * ifMetagenerationNotMatch

* [Bucket Get](https://cloud.google.com/storage/docs/json_api/v1/buckets/get): [google.cloud.storage.bucket.Bucket.reload](https://googleapis.dev/python/storage/latest/buckets.html#google.cloud.storage.bucket.Bucket.reload)

    * ifMetagenerationMatch

    * ifMetagenerationNotMatch

* [Bucket Get](https://cloud.google.com/storage/docs/json_api/v1/buckets/get): [google.cloud.storage.client.Client.lookup_bucket](https://googleapis.dev/python/storage/latest/client.html#google.cloud.storage.client.Client.lookup_bucket)

    * ifMetagenerationMatch

    * ifMetagenerationNotMatch

* [Bucket Get](https://cloud.google.com/storage/docs/json_api/v1/buckets/get): [google.cloud.storage.bucket.Bucket.exists](https://googleapis.dev/python/storage/latest/buckets.html#google.cloud.storage.bucket.Bucket.exists)

    * ifMetagenerationMatch

    * ifMetagenerationNotMatch

* [Bucket Patch](https://cloud.google.com/storage/docs/json_api/v1/buckets/patch): [google.cloud.storage.bucket.Bucket.patch](https://googleapis.dev/python/storage/latest/buckets.html#google.cloud.storage.bucket.Bucket.patch)

    * ifMetagenerationMatch

    * ifMetagenerationNotMatch

[...]

PR #153, opened 2020-05-20
feat: add Bucket.reload() and Bucket.update() wrappers to restrict generation match args

IlyaFaer:
Towards #127

### DIFF
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -783,6 +783,91 @@ class Bucket(_PropertyMixin):
             timeout=timeout,
         )
 
+    def update(
+        self,
+        client=None,
+        timeout=_DEFAULT_TIMEOUT,
+        if_metageneration_match=None,
+        if_metageneration_not_match=None,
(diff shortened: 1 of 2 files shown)


IlyaFaer:
@andrewsg, @frankyn, I've taken one more self review and found the thing, that got no our attention yet. We have only two `if_metageneration*` args in `Bucket.reload()` and `Bucket.update()` methods according to the design, and all four `if_*generation*` args in `Blob.reload()` and `Blob.update()` methods.

All of the above methods are inherited from `_PropertyMixin` class. I've added all four args into its methods, but that means user can pass `if_generation_match` arg into `Bucket.reload()`, which is not correct. More than that these args will be shown in docs, like they are okay to be passed:

Безымянный


I'm proposing to add wrapper methods into `Bucket` class to restrict `if_generation*` args, and keep only `if_metageneration*` ones.

andrewsg:
LGTM but I'll pass this by Frank before merging just in case I'm missing something.

PR #158, opened 2020-05-21
fix(storage): fix upload object with bucket cmek enabled

### LINKED ISSUE #155: Unable to upload new object when CMEK is enabled for a bucket.
Tracking issue for a customer.

The issue is the Python library does a metadata get on an Object which includes kmsKeyName version resource ID metadata. 
If you use the same Blob instance to perform an upload the Python library will use the kmsKeyName version resource ID instead of the kmsKeyName resource ID. 
Cloud Storage API expects the kmsKeyName kmsKeyName without version information.
 
Here's an example for illustration:
kmsKeyName version resource ID:  
```
projects/project-id/us/keyRings/testmrbucket/cryptoKeys/testkey/cryptoKeyVersions/1
```
kmsKeyName resource ID: 

```
projects/project-id/locations/us/keyRings/testmrbucket/cryptoKeys/testkey
```
# Reproduction

```python
 
bucket_name = 'your-bucket-name'
blob_name = 'your-object-name'
 
client = storage.Client()
bucket = client.bucket(bucket_name)

# Creates a random encrypted blob.
blob = bucket.blob(blob_name)
blob.upload_from_string("oldcontent")
blob.upload_from_string("newcontent", if_generation_match=blob.generation)
```

# Workaround

```python
 
bucket_name = 'your-bucket-name'
blob_name = 'your-object-name'
 
client = storage.Client()
bucket = client.bucket(bucket_name)

# Creates a random encrypted blob.
blob = bucket.blob(blob_name)
blob.upload_from_string("oldcontent")
# Store the generation
saved_generation = blob.generation

# Get a new instance of Blob to unset value of kmsKeyName
blob = bucket.blob(blob_name)
blob.upload_from_string("newcontent", if_generation_match=saved_generation)
```

Potential Fix:

* `blob.upload_from_*` should only use kmsKeyName if it doesn't end with a version.

HemangChothani:
Fixes #155

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -1280,7 +1280,10 @@ class Blob(_PropertyMixin):
         if self.user_project is not None:
             name_value_pairs.append(("userProject", self.user_project))
 
-        if self.kms_key_name is not None:
+        if (
+            self.kms_key_name is not None
+            and "cryptoKeyVersions" not in self.kms_key_name
+        ):
             name_value_pairs.append(("kmsKeyName", self.kms_key_name))
 
         if predefined_acl is not None:
@@ -1417,7 +1420,10 @@ class Blob(_PropertyMixin):
         if self.user_project is not None:
             name_value_pairs.append(("userProject", self.user_project))
 
-        if self.kms_key_name is not None:
+        if (
+            self.kms_key_name is not None
+            and "cryptoKeyVersions" not in self.kms_key_name
+        ):
             name_value_pairs.append(("kmsKeyName", self.kms_key_name))
 
         if predefined_acl is not None:
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -1280,10 +1280,11 @@ class Blob(_PropertyMixin):
         if self.user_project is not None:
             name_value_pairs.append(("userProject", self.user_project))
 
-        # When cmek is enabled for bucket blob.metadata stores the value of
-        # kmsKeyName with version information whereas upload method expect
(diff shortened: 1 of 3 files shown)


google/cloud/storage/blob.py
```diff
@@ -1280,7 +1280,10 @@ def _do_multipart_upload(
         if self.user_project is not None:
             name_value_pairs.append(("userProject", self.user_project))
 
-        if self.kms_key_name is not None:
+        if (
+            self.kms_key_name is not None
+            and "cryptoKeyVersions" not in self.kms_key_name
```
frankyn, line 1285: Please add comments on why this is needed for posterity. Mainly that Cloud Storage object metadata stores kmsKeyName with version information and uploads() expect non-versioned information.

google/cloud/storage/blob.py
```diff
@@ -1280,7 +1280,14 @@ def _do_multipart_upload(
         if self.user_project is not None:
             name_value_pairs.append(("userProject", self.user_project))
 
-        if self.kms_key_name is not None:
+        # When cmek is enabled for bucket blob.metadata stores the value of
+        # kmsKeyName with version information whereas upload method expect
+        # information without version so this condition ignores the  kmsKeyName
+        # which has cryptoKeyVersions.
```
frankyn, line 1286: Suggested changes:

> When a Customer Managed Encryption Key is used to encrypt Cloud Storage object at rest, object resource metadata will store the version of the Key Management Service cryptographic material. If a Blob instance with KMS Key metadata set is used to upload a new version of the object then the existing kmsKeyName version value can't be used in the upload request and the client instead ignores it.

PR #172, opened 2020-06-05
fix: set default POST policy scheme to "http"

IlyaFaer:


google/cloud/storage/client.py
```diff
@@ -928,7 +928,7 @@ def generate_signed_post_policy_v4(
         credentials=None,
         virtual_hosted_style=False,
         bucket_bound_hostname=None,
-        scheme=None,
+        scheme="http",
```
IlyaFaer, line 931: It came to my attention that everywhere we're setting `scheme` to `"http"` by default, except POST policies. Doesn't seem to be correct.

PR #182, opened 2020-06-17
feat: rename 'Blob.download_as_{string,bytes}', add 'Blob.download_as_text'

### LINKED ISSUE #178: please provide `download_as_bytes` for blobs
Now that python2.7 support is deprecated, `download_as_string` is never correct.

I doubt it can safely be changed to actually download a string.  But at least is the API provided a `download_as_bytes`, we could avoid the confusing pattern where code looks like it downloaded a string and then must convert it to a string.

HemangChothani:
Leave 'Blob.download_as_string' as a deprecated alias for 'download_as_bytes'.

Fixes #178


google/cloud/storage/blob.py
```diff
@@ -1049,6 +1049,10 @@ def download_as_string(
     ):
         """Download the contents of this blob as a bytes object.
 
+        .. note::
+           The method only supports `python3`, for python2 it returns data
+           as a string.
```
tseaver, line 1054: This note is incorrect, or at least confusing:  the return value under Python2 *will* be bytes, because the native `str` type is bytes.

google/cloud/storage/blob.py
```diff
@@ -1106,6 +1110,77 @@ def download_as_string(
...
+                                        is a live version of the blob.
+
+        :param if_metageneration_match: (Optional) Make the operation conditional on whether the
+                                        blob's current metageneration matches the given value.
+
+        :type if_metageneration_not_match: long
+        :param if_metageneration_not_match: (Optional) Make the operation conditional on whether the
+                                            blob's current metageneration does not match the given value.
+
+        :rtype: str
```
tseaver, line 1164: This should be `unicode` under Python2, rather than `str`, or even just "text".

google/cloud/storage/blob.py
```diff
@@ -1106,6 +1110,77 @@ def download_as_string(
         )
         return string_buffer.getvalue()
 
+    def download_as_string(
```
tseaver, line 1113: `download_as_text` would be a clearer name for this method, given the historic ambiguity of the term "string" in Python.

google/cloud/storage/blob.py
```diff
@@ -1106,6 +1110,77 @@ def download_as_string(
...
+            raw_download=raw_download,
+            if_generation_match=if_generation_match,
+            if_generation_not_match=if_generation_not_match,
+            if_metageneration_match=if_metageneration_match,
+            if_metageneration_not_match=if_metageneration_not_match,
+        )
+        if six.PY2:
+            return data
+        else:
+            return data.decode()
```
tseaver, line 1182: This method should return text, unambiguously, and should use the blob's `content_encoding` property to do the decoding.  Perhaps the method should take an optional `encoding` parameter, allowing the caller to pass through the value for blobs where the `content_encoding` is not set / not fetched from the server.

google/cloud/storage/blob.py
```diff
@@ -1220,11 +1308,15 @@ def download_as_string(
             if_generation_not_match=if_generation_not_match,
             if_metageneration_match=if_metageneration_match,
             if_metageneration_not_match=if_metageneration_not_match,
+            timeout=timeout,
         )
         if six.PY2:
             return data
```
tseaver, line 1314: We should be decoding the bytes to text even on Python 2.

google/cloud/storage/blob.py
```diff
@@ -1165,6 +1161,85 @@ def download_as_string(
         if_generation_not_match=None,
         if_metageneration_match=None,
         if_metageneration_not_match=None,
+        timeout=_DEFAULT_TIMEOUT,
+    ):
+        """Download the contents of this blob as a bytes object.
```
tseaver, line 1166: The summary line should note that this method is deprecated, and be followed by a note, e.g.:

```suggestion
        """(Deprecated) Download the contents of this blob as a bytes object.
        
        .. note::
           Deprecated alias for :meth:`download_as_bytes`.
```

google/cloud/storage/blob.py
```diff
@@ -1165,6 +1161,85 @@ def download_as_string(
...
+            server response. Depending on the retry strategy, a request may be
+            repeated several times using the same timeout each time.
+            Can also be passed as a tuple (connect_timeout, read_timeout).
+            See :meth:`requests.Session.request` documentation for details.
+
+        :rtype: bytes
+        :returns: The data stored in this blob.
+
+        :raises: :class:`google.cloud.exceptions.NotFound`
+        """
```
tseaver, line 1218: Emit a `DeprecationWarning` here.

HemangChothani:
@tseaver  PTAL  whenever have a chance.

google/cloud/storage/blob.py
```diff
@@ -1340,7 +1349,10 @@ def download_as_text(
             timeout=timeout,
         )
         if six.PY2:
```
tseaver, line 1351: Please remove this `if`:  both branches are now identical.

Issue #38, opened 2019-04-22
Proposal: move API-methods to client


tswast:
The Cloud Storage Python client was one of the first google-cloud-python clients, intended to be more reliable and more Pythonic than the google-api-client that it replaced. It is also a 100% hand-written client. Because it predates the auto-generated google-cloud-python clients, the Cloud Storage Python client has inconsistencies with other google-cloud-python clients.

This design proposal is to bring the Cloud Storage client further into alignment with other google-cloud-python clients and make the eventual inclusion of an auto-generated gRPC transport layer less disruptive to users of the Cloud Storage client.

**Design document:**

https://docs.google.com/document/d/1A4FIxThZK_enK7OV9ChY54IPrlpbdHwuhvVEu6iTHxM/edit?usp=sharing

Please share your feedback on the specifics of this as comments in the [design document](https://docs.google.com/document/d/1A4FIxThZK_enK7OV9ChY54IPrlpbdHwuhvVEu6iTHxM/edit?usp=sharing).

CC @crwilcox @frankyn @lbristol88 

**Update**: according to [comment](https://github.com/googleapis/google-cloud-python/pull/8604#pullrequestreview-313722140), we need to update samples for moved methods before deprecating the old ones.

frankyn:
@IlyaFaer I believe this is complete. I'm closing FR.

If there's something that I missed please reopen.

IlyaFaer:
@frankyn, I've delegated the last part to @paul1319
@paul1319, do you have something to push for this issue?

frankyn:
Reopening in case there's something pending.

frankyn:
@IlyaFaer haven't heard back for a bit. What's remaining?

paul1319:
@frankyn as I see Blob class Refactor and Deprecate part.

tseaver:
ISTM that this effort is not complete:  e.g., the client does not have a `get_blob` method, nor any of the notification-related methods for buckets.

This partial refactoring complicates the tests a *lot* -- if it were complete, then bucket / blob unit tests would just use mocked clients and assert that the corresponding method is called correctly, rather than creating mock connections (and real clients, see #416) and grubbing around three layers down.

PR #184, opened 2020-06-18
refactor(storage): move 'list_blobs' implementation from 'Bucket' and  'download_blob_to_file' implementation from 'Blob' to `Client`

paul1319:
Towards #38


tests/unit/test_blob.py
```diff
@@ -43,6 +43,12 @@ def _make_one(*args, **kw):
         blob._properties.update(properties)
         return blob
 
+    @staticmethod
+    def _make_one_client(*args, **kw):
+        from google.cloud.storage.client import Client
+
+        return Client(*args, **kw)
```
tseaver, line 50: It would be a better strategy to return an autospec'ed `Mock` instance, than to depend on the actual `Client` implementation.  E.g.:

```suggestion
        return mock.create_autospec(Client, instance=True, **kw)
```

tests/unit/test_bucket.py
```diff
@@ -380,6 +380,14 @@ def _make_one(self, client=None, name=None, properties=None, user_project=None):
         bucket._properties = properties or {}
         return bucket
 
+    @staticmethod
+    def _make_one_client(*args, **kw):
```
tseaver, line 384: Likewise here: `_make_client`, as well as returning an autospec'ed Mock.

tests/unit/test_client.py
```diff
@@ -993,34 +993,103 @@ def test_create_bucket_w_object_success(self):
...
+        raw_response.request = raw_request.prepare()
+        grmp_response = InvalidResponse(raw_response)
+
+        blob_name = "blob-name"
+        media_link = "http://test.invalid"
         credentials = _make_credentials()
-        client = self._make_one(project=project, credentials=credentials)
-        blob = mock.Mock()
+        client = self._make_one(credentials=credentials)
+        bucket = _Bucket(client)
```
tseaver, line 1012: It doesn't seem like the hand-rolled `_Bucket` shim is useful here:  none of the attributes defined in that class get used.  Could it be just a `mock.Mock`?  Better yet, if we use an autospec mock for the `blob` (as I suggest below) we don't even need to create a bucket at all.

tests/unit/test_client.py
```diff
@@ -993,34 +993,103 @@ def test_create_bucket_w_object_success(self):
...
+    def test_download_blob_to_file_wo_media_link(self):
+        from google.cloud.storage.blob import Blob
+
+        blob_name = "blob-name"
         credentials = _make_credentials()
-        client = self._make_one(project=project, credentials=credentials)
-        blob = mock.Mock()
+        client = self._make_one(credentials=credentials)
+        bucket = _Bucket(client)
+        blob = Blob(blob_name, bucket=bucket)
```
tseaver, line 1036: We really don't need to care about whether the blob returns a media link or another path when testing the `Client` method. It would thus be better to use an autospec'ed mock here, e.g.:

```suggestion
        blob = mock.create_autospec(Blob)
```

and then mock out its `_get_download_url` method (as with the `_do_download` method).

google/cloud/storage/blob.py
```diff
@@ -969,6 +969,7 @@ def download_to_file(
             if_generation_not_match=if_generation_not_match,
             if_metageneration_match=if_metageneration_match,
             if_metageneration_not_match=if_metageneration_not_match,
+            timeout=timeout
```
tseaver, line 972: Lint wants a trailing comma here.

tests/unit/test_bucket.py
```diff
@@ -380,6 +380,14 @@ def _make_one(self, client=None, name=None, properties=None, user_project=None):
         bucket._properties = properties or {}
         return bucket
 
+    @staticmethod
+    def _make_client(*args, **kw):
+        from google.cloud.storage.client import Client
+
+        client = Client(*args, **kw)
```
tseaver, line 387:
```suggestion
        return mock.create_autospec(Client, instance=True, **kw)
```

tests/unit/test_client.py
```diff
@@ -1919,3 +2012,14 @@ def _time_functions_patches():
         "google.cloud.storage.client.get_expiration_seconds_v4", return_value=10
     )
     return dtstamps_patch, now_patch, expire_secs_patch
+
+
+class _Bucket(object):
```
tseaver, line 2017: Delete this class after refactoring above to use a mock rather than an actual `Blob`.

PR #195, opened 2020-06-25
refactor(storage): erase excess return

IlyaFaer:
`patch()` doesn't return anything, so this `return` statement looks little bit confusing


andrewsg:
Thanks!

Issue #108, opened 2020-04-15
Add retry (exponential backoff) for storage api functions


brfulu:
While using the python-storage library, we were getting 503 or 504 errors from time to time when deleting or moving objects.
I looked up the [docs](https://cloud.google.com/storage/docs/exponential-backoff) for cloud storage and came across this:
> Clients should use truncated exponential backoff for all requests to Cloud Storage that return HTTP 5xx and 429 response codes, including uploads and downloads of data or metadata.

At the bottom of the page there is also:
> Examples of backoff implemented in client libraries that you can use with Cloud Storage include:
> - Retrying library for Python.
> - Google Cloud Client Libraries for Node.js can automatically use backoff strategies to retry requests with the autoRetry parameter.

It would be super useful and in my opinion necessary to include retrying functionality into the python-storage library. 
It was also a bit confusing to me that the official docs are recommending a third party retrying library while already having [google.api.core.retry](https://googleapis.dev/python/google-api-core/latest/retry.html) inside google cloud python sdk. 

The `google.api.core.retry` docs mention this:
> Some client library methods apply retry automatically. These methods can accept a retry parameter that allows you to configure the behavior

A potential solution could be to add the retry parameter of type `google.api.core.retry` to public api functions or while creating the storage client. Developers could specify if they want exponential backoff and set the parameters as they like (maximum delay, deadline, etc.)

Java and Node.js libraries already have similar functionality and it seems that this can be useful when cloud storage is facing some issues or high traffic load, or in the case where users are hitting the limits of 1k writes or 5k reads per second and GCS is autoscaling.

If this is already supported in the library, please direct me to some docs, I wasn't able to find it...

brfulu:
Hey everyone, is there any update about this?

frankyn:
I'm working on this and don't have an update yet to share. I will update this issue when I do. Thank you the ping.

simonvanderveldt:
There's a prototype PR for this now https://github.com/googleapis/python-storage/pull/197

PR #197, opened 2020-06-27
[Prototype] Retries for Storage

frankyn:
Prototype based on BigQuery retry implementation in Python: https://github.com/googleapis/python-bigquery/blob/master/google/cloud/bigquery/retry.py

Work towards resolving: #108

Remaining:

- [ ] Idempotent operations are retried *only*
- [ ] Reset Connection Approval
- [ ] Unit Tests
- [ ] Review


google/cloud/storage/_helpers.py
```diff
@@ -134,6 +136,14 @@ def _query_params(self):
             params["userProject"] = self.user_project
         return params
 
+
+    def _call_api(self, client, retry, **kwargs):
+        call = functools.partial(client._connection.api_request, **kwargs)
```
andrewsg, line 141: Is this because the interface for the retry function is not expected to pass in any arguments?

frankyn: Copied it over mainly:
At the moment no, I'm not making it configurable for the first pass and only introducing it to help support/unblock customer friction when using the library.

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,57 @@
...
+"""Retry Strategy logic used across google.cloud.storage requests."""
+
+# Should not be in here and only for prototyping
+import six
+import socket
+import requests
+import urllib3
+
+from google.api_core import exceptions
+from google.api_core import retry
```
tritone, line 23: Neat, so this is what we told customers to use as a workaround already right?

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,57 @@
...
+    """Predicate for determining when to retry."""
+
+    if hasattr(exc, "errors"):
+        if len(exc.errors) == 0:
+            # Check for unstructured error returns, e.g. from GFE
+            return isinstance(exc, _UNSTRUCTURED_RETRYABLE_TYPES)
+        reason = exc.errors[0]["reason"]
+
+        return reason in _RETRYABLE_REASONS
+    else:
```
tritone, line 49: No need for `else` here right? And the nested if statements and uses of isinstance look pretty ugly, I'm sure there's a nicer way to do this.

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,57 @@
...
+def _should_retry(exc):
+    """Predicate for determining when to retry."""
+
+    if hasattr(exc, "errors"):
+        if len(exc.errors) == 0:
+            # Check for unstructured error returns, e.g. from GFE
+            return isinstance(exc, _UNSTRUCTURED_RETRYABLE_TYPES)
+        reason = exc.errors[0]["reason"]
+
+        return reason in _RETRYABLE_REASONS
```
tritone, line 48: Do you not have access to response codes here? Seems kind of a strange workaround to be doing string matching for a 503 for example.

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,57 @@
...
+    else:
+        # Connection Reset
+        if isinstance(exc, requests.exceptions.ConnectionError):
+            if isinstance(exc.args[0], urllib3.exceptions.ProtocolError):
+                if isinstance(exc.args[0].args[1], ConnectionResetError):
+                    return True
+    return False
+
+_DEFAULT_RETRY = retry.Retry(predicate=_should_retry)
```
tritone, line 57: I would be curious to look at other examples of where people have written these predicates.

google/cloud/storage/retry.py
tritone: Also, where are configs around number of attempts or timeouts? Are these values hard-coded into google.api_core.retry?

google/cloud/storage/_helpers.py
```diff
@@ -134,6 +136,14 @@ def _query_params(self):
             params["userProject"] = self.user_project
         return params
 
+
+    def _call_api(self, client, retry, **kwargs):
```
tritone, line 140: Should retry be a kwarg? Do callers explicitly have to pass in None if they don't want a retry? Would be good to clarify here.

simonvanderveldt:
Is there any chance of this being merged soon? We'd very much like to have this functionality available in the library, we currently have to wrap everything in retries ourselves.

simonvanderveldt:
@busunkim96 Why is this closed? Is there a new PR?

andrewsg:
I'll be working on a new one.

andyhky:
Is there any update on this?

We recently encountered no retries for `google.cloud.storage.blob.Blob.reload` https://googleapis.dev/python/storage/latest/blobs.html?#google.cloud.storage.blob.Blob.reload and it would be nice to have this functionality.

frankyn:
Closing this PR as @andrewsg is working on a new PR now. PR should be available soon. Thank you for your patience.

PR #199, opened 2020-07-01
feat: add support for 'Blob.custom_time' and lifecycle rules

### LINKED ISSUE #196: Support customTime metadata and custom timestamp based OLM
Request to support a new `customTime` metadata field and associated custom timestamp based OLM conditions:

```python
bucket.add_lifecycle_delete_rule(
    days_since_custom_time=2, 
    custom_time_before=(datetime.now() + datetime.timedelta(days=10)))

# New metadata field:
blob = bucket.get_blob('object-name')
blob.custom_time = "2020-01-01T00:00:00Z"
blob.patch()
```

HemangChothani:
Fixes #196


google/cloud/storage/blob.py
```diff
@@ -3003,6 +3003,21 @@ def updated(self):
...
+        See https://cloud.google.com/storage/docs/json_api/v1/objects
+
+        :rtype: :class:`datetime.datetime` or ``NoneType``
+        :returns: Datetime object parsed from RFC3339 valid timestamp, or
+                  ``None`` if the blob's resource has not been loaded from
+                  the server (see :meth:`reload`).
+        """
+        value = self._properties.get("customTime")
+        if value is not None:
+            return _rfc3339_to_datetime(value)
```
tseaver, line 3019: This property needs a setter, as well (at least, @jkwlui's example shows setting it).

tseaver:
I have marked this PR as `do not merge` per @jkwlui .

google/cloud/storage/blob.py
```diff
@@ -3018,6 +3019,18 @@ def custom_time(self):
...
+    def custom_time(self, value):
+        """Set the custom time for the object.
+
+        See https://cloud.google.com/storage/docs/json_api/v1/objects
+
+        :type value: :class:`datetime.datetime`
+        :param value: (Optional) Set the custom time of blob. Datetime object parsed
+                                 from RFC3339 valid timestamp.
+        """
+        self._properties["customTime"] = _datetime_to_rfc3339(value)
```
tseaver, line 3203: If `value` is None, the conversion will fail.  Suggest:
```suggestion
        if value is not None:
            value = _datetime_to_rfc3339(value)
            
        self._properties["customTime"] = value
```

tests/unit/test_blob.py
```diff
@@ -3804,6 +3804,17 @@ def test_custom_time(self):
...
+    def test_custom_time_setter(self):
+        from google.cloud._helpers import UTC
+
+        BLOB_NAME = "blob-name"
+        bucket = _Bucket()
+        TIMESTAMP = datetime.datetime(2014, 11, 5, 20, 34, 37, tzinfo=UTC)
+        blob = self._make_one(BLOB_NAME, bucket=bucket)
+        self.assertIsNone(blob.custom_time)
+        blob.custom_time = TIMESTAMP
+        self.assertEqual(blob.custom_time, TIMESTAMP)
```
tseaver, line 4153: Need to test with `value` set to None, as well.

google/cloud/storage/blob.py
```diff
@@ -3018,6 +3019,18 @@ def custom_time(self):
...
             return _rfc3339_to_datetime(value)
 
+    @custom_time.setter
+    def custom_time(self, value):
+        """Set the custom time for the object.
+
+        See https://cloud.google.com/storage/docs/json_api/v1/objects
+
+        :type value: :class:`datetime.datetime`
+        :param value: (Optional) Set the custom time of blob. Datetime object parsed
```
tseaver, line 3200: Document the semantics of `None` value (which should be to clear any existing property value).

tests/unit/test_blob.py
```diff
@@ -4152,6 +4152,14 @@ def test_custom_time_setter(self):
         blob.custom_time = TIMESTAMP
         self.assertEqual(blob.custom_time, TIMESTAMP)
 
+    def test_custom_time_setter_none_value(self):
+        BLOB_NAME = "blob-name"
+        bucket = _Bucket()
+        TIMESTAMP = None
+        blob = self._make_one(BLOB_NAME, bucket=bucket)
+        blob.custom_time = TIMESTAMP
+        self.assertIsNone(blob.custom_time)
```
tseaver, line 4161: This test should pre-populate the properties dict, e.g.:

```python
        TIMESTAMP = datetime.datetime(2014, 11, 5, 20, 34, 37, tzinfo=UTC)
        TIME_CREATED = TIMESTAMP.strftime(_RFC3339_MICROS)
        properties = {"customTime": TIME_CREATED}
        blob = self._make_one(BLOB_NAME, bucket=bucket, properties=properties)
```

HemangChothani:
> Update: The Cloud Storage team has updated type of noncurrent_time_before and custom_time to be a Date format (YYYY-mm-dd) instead of DateTime.

@frankyn That change has been already done.

google/cloud/storage/blob.py
```diff
@@ -3348,6 +3349,35 @@ def updated(self):
...
             return _rfc3339_to_datetime(value)
 
+    @property
+    def custom_time(self):
+        """Retrieve the custom time for the object.
+
+        See https://cloud.google.com/storage/docs/json_api/v1/objects
+
+        :rtype: :class:`datetime.date` or ``NoneType``
+        :returns: Date object parsed from iso8601 valid date, or
```
frankyn, line 3359: This should be dateTime

google/cloud/storage/blob.py
```diff
@@ -3348,6 +3349,36 @@ def updated(self):
...
+        if value is not None:
+            return _rfc3339_to_datetime(value)
+
+    @custom_time.setter
+    def custom_time(self, value):
+        """Set the custom time for the object.
+
+        See https://cloud.google.com/storage/docs/json_api/v1/objects
+
+        :type value: :class:`datetime.datetime` or ``NoneType``
```
frankyn, line 3373: Once set it can't be unset and only changed to a custom datetime in the future. If the custom_time must be unset, you must either perform a rewrite operation or upload the data again.

google/cloud/storage/bucket.py
```diff
@@ -170,6 +170,17 @@ class LifecycleRuleConditions(dict):
...
+    :type days_since_custom_time: int
+    :param days_since_custom_time: (Optional) Apply rule action to items whose number of days
+                                   elapsed since the custom timestamp. This condition is relevant
+                                   only for versioned objects. The value of the field must be a non
+                                   negative integer. If it's zero, the object version will become
+                                   eligible for lifecycle action as soon as it becomes custom.
+
+    :type custom_time_before: :class:`datetime.date`
+    :param custom_time_before: (Optional) Apply rule action to items whose custom time is before this
+                                date. This condition is relevant only for versioned objects.
```
frankyn, line 182: Please include similar text to noncurrent_time_before with ISO8601 and an example.

tests/unit/test_blob.py
frankyn: Please address @tseaver's comment.

HemangChothani: addressed.

HemangChothani: addressed.

google/cloud/storage/blob.py
frankyn: I'd state that setter can only be datetime.datetime to not confuse folks.

google/cloud/storage/bucket.py
```diff
@@ -266,6 +286,18 @@ def number_of_newer_versions(self):
...
+    def days_since_custom_time(self):
+        """Conditon's 'days_since_custom_time' value."""
+        return self.get("daysSinceCustomTime")
+
+    @property
+    def custom_time_before(self):
+        """Conditon's 'custom_time_before' value."""
+        before = self.get("customTimeBefore")
+        if before is not None:
+            return datetime_helpers.from_iso8601_date(before)
```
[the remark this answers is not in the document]
HemangChothani: Not sure about the setter method for all this properties, because we don't have a method for other properties as well which  defined before. Please let me know if needed, so need to create setter method for all the properties of `LifecycleRuleConditions` class.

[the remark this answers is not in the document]
frankyn: Thanks both, looks like setters weren't defined at this level. It will be filed as a feature request for now. It doesn't block customers from using the feature.

PR #204, opened 2020-07-15
feat(blob.py): auto-populate standard headers for non-chunked downloads

### LINKED ISSUE #24: Storage: Capture relevant headers to blob properties during download
Residual from googleapis/google-cloud-python#9003.

@william-silversmith notes that even with `raw_download` enabled, he is unable to detect the `content_type` of a downloaded blob without performing an additional `reload` request, which is prohibitive for his usecase at scale.  E.g.:

```python
blob = bucket.blob( key )
binary = blob.download_as_string(raw_download=True)
if blob.content_encoding == 'gzip':
    return gunzip(binary)
elif blob.content_encoding == 'br':
    return brotli.decompress(binary)
else:
    return binary
```
Potentially even...

```python
if blob.content_type == 'application/json':
    return json.loads(binary.decode('utf8'))
```

william-silversmith:

Fixes #24  🦕

This PR autopopulates the following fields for non-chunked downloads based on the server header response:
```
blob.content_encoding
blob.content_type
blob.cache_control
blob.storage_class
blob.content_language
blob.md5_hash
blob.crc32c
```

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -803,7 +803,7 @@ class Blob(_PropertyMixin):
             #  'X-Goog-Hash': 'crc32c=4gcgLQ==,md5=CS9tHYTtyFntzj7B9nkkJQ==',
             x_goog_hash = response.raw.headers.get("X-Goog-Hash", '')
         except AttributeError:
-            pass
+            x_goog_hash = ''
 
         digests = {}
         for encoded_digest in x_goog_hash.split(","):
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -794,16 +794,13 @@ class Blob(_PropertyMixin):
             :class requests.models.Response
         :param response: The server response from downloading a non-chunked file
         """
-        try:
-            self.content_encoding = response.raw.headers.get("Content-Encoding", None)
-            self.content_type = response.raw.headers.get("Content-Type", None)
-            self.cache_control = response.raw.headers.get("Cache-Control", None)
-            self.storage_class = response.raw.headers.get("X-Goog-Storage-Class", None)
-            self.content_language = response.raw.headers.get("Content-Language", None)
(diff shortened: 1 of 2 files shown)


william-silversmith:
@tseaver would these unit tests be sufficient? I can get more involved if necessary. I had trouble invoking _do_download, it was interacting strangely with "MagicMock" and dying on closing a file stream.

tseaver:
@william-silversmith Thanks, that new test was sufficient.

william-silversmith:
I'm a little confused because 806-808 have to be evaluated for the crc32c and md5_hash tests to pass.

google/cloud/storage/blob.py
```diff
@@ -783,6 +784,34 @@ def _get_download_url(
...
+            :class requests.models.Response
+        :param response: The server response from downloading a non-chunked file
+        """
+        self.content_encoding = response.headers.get("Content-Encoding", None)
+        self.content_type = response.headers.get("Content-Type", None)
+        self.cache_control = response.headers.get("Cache-Control", None)
+        self.storage_class = response.headers.get("X-Goog-Storage-Class", None)
+        self.content_language = response.headers.get("Content-Language", None)
+        #  'X-Goog-Hash': 'crc32c=4gcgLQ==,md5=CS9tHYTtyFntzj7B9nkkJQ==',
+        x_goog_hash = response.headers.get("X-Goog-Hash", '')
```
tseaver, line 803: `black` hates single quotes:
```suggestion
        x_goog_hash = response.headers.get("X-Goog-Hash", "")
```

william-silversmith: Thanks!

tseaver:
@googlebot I consent

william-silversmith:
@googlebot I fixed it.

william-silversmith:
I could try force pushing a double quote update that removes you as an author if that would be okay with you?

tseaver:
@william-silversmith Sure, WFM.

tseaver:
W00t!  Thanks for persevering, @william-silversmith .

william-silversmith:
Excited thank you!

Issue #226, opened 2020-08-04
Systests fail under Python 3.8


tseaver:
From [this Kokoro build]():

```python
______ TestStorageSignURLs.test_create_signed_read_url_v2_w_access_token _______

self = <test_system.TestStorageSignURLs testMethod=test_create_signed_read_url_v2_w_access_token>

    def test_create_signed_read_url_v2_w_access_token(self):
        client = iam_credentials_v1.IAMCredentialsClient()
        service_account_email = Config.CLIENT._credentials.service_account_email
>       name = google.api_core.path_template.expand(
            "projects/{project}/serviceAccounts/{service_account}",
            project="-",
            service_account=service_account_email,
        )
E       AttributeError: module 'google.api_core' has no attribute 'path_template'

tests/system/test_system.py:1283: AttributeError
______ TestStorageSignURLs.test_create_signed_read_url_v4_w_access_token _______

self = <test_system.TestStorageSignURLs testMethod=test_create_signed_read_url_v4_w_access_token>

    def test_create_signed_read_url_v4_w_access_token(self):
        client = iam_credentials_v1.IAMCredentialsClient()
        service_account_email = Config.CLIENT._credentials.service_account_email
>       name = google.api_core.path_template.expand(
            "projects/{project}/serviceAccounts/{service_account}",
            project="-",
            service_account=service_account_email,
        )
E       AttributeError: module 'google.api_core' has no attribute 'path_template'

tests/system/test_system.py:1301: AttributeError
_____________ TestKMSIntegration.test_blob_w_explicit_kms_key_name _____________

self = <test_system.TestKMSIntegration testMethod=test_blob_w_explicit_kms_key_name>

    def setUp(self):
        super(TestKMSIntegration, self).setUp()
        client = kms.KeyManagementServiceClient()
        project = Config.CLIENT.project
        location = self.bucket.location.lower()
        keyring_name = self.KEYRING_NAME
>       purpose = kms.enums.CryptoKey.CryptoKeyPurpose.ENCRYPT_DECRYPT
E       AttributeError: module 'google.cloud.kms' has no attribute 'enums'

tests/system/test_system.py:1861: AttributeError
____________ TestKMSIntegration.test_bucket_w_default_kms_key_name _____________

self = <test_system.TestKMSIntegration testMethod=test_bucket_w_default_kms_key_name>

    def setUp(self):
        super(TestKMSIntegration, self).setUp()
        client = kms.KeyManagementServiceClient()
[...]

PR #227, opened 2020-08-04
testing: fix / skip systests broken by dependency releases

tseaver:
See #226.  This PR doesn't *fix* the issue, because the KMS breakage exists:  it just skips the KMS systests under Python3, until we have dropped support for Python2.


tseaver:
@busunkim96 OK, that works:  #232

PR #234, opened 2020-08-05
init commit

AndrewAXue:

### DIFF
--- a/google/cloud/storage/_opentelemetry_meter.py
+++ b/google/cloud/storage/_opentelemetry_meter.py
@@ -1,0 +1,53 @@
+# Copyright 2014 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Manages OpenTelemetry trace creation and handling"""
+
+try:
+    from opentelemetry import metrics
+    from opentelemetry.sdk.metrics import Counter, MeterProvider
+    from opentelemetry.exporter.cloud_monitoring import (
+        CloudMonitoringMetricsExporter,
+    )
+
+    HAS_OPENTELEMETRY_INSTALLED = True
+except ImportError:
+    HAS_OPENTELEMETRY_INSTALLED = False
+
+instruments_setup = False
+requests_counter = None
+
+def setup_instruments():
+    global requests_counter, instruments_setup
+    meter = metrics.get_meter(__name__)
+    metrics.get_meter_provider().start_pipeline(
+        meter, CloudMonitoringMetricsExporter(), 10
+    )
+    requests_counter = meter.create_metric(
+        name="GCS_request_counter",
+        description="number of requests",
+        unit="1",
+        value_type=int,
+        metric_type=Counter,
+        label_keys=("path"),
+    )
+    instruments_setup = True
+
+
+def telemetry_wrapped_api_request(api_request, **kwargs):
+    if HAS_OPENTELEMETRY_INSTALLED:
+        if not instruments_setup:
+            setup_instruments()
+        requests_counter.add(1, {'path': kwargs['path']})
+    return api_request(**kwargs)
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -24,7 +24,7 @@ import os
 
 from six.moves.urllib.parse import urlsplit
 from google.cloud.storage.constants import _DEFAULT_TIMEOUT
-
+from google.cloud.storage._opentelemetry_meter import telemetry_wrapped_api_request
 
 STORAGE_EMULATOR_ENV_VAR = "STORAGE_EMULATOR_HOST"
 """Environment variable defining host for Storage emulator."""
@@ -198,7 +198,7 @@ class _PropertyMixin(object):
             if_metageneration_match=if_metageneration_match,
             if_metageneration_not_match=if_metageneration_not_match,
         )
-        api_response = client._connection.api_request(
+        api_response = telemetry_wrapped_api_request(client._connection.api_request,
             method="GET",
             path=self.path,
             query_params=query_params,
@@ -299,7 +299,7 @@ class _PropertyMixin(object):
         update_properties = {key: self._properties[key] for key in self._changes}
 
         # Make the API call.
-        api_response = client._connection.api_request(
+        api_response = telemetry_wrapped_api_request(client._connection.api_request,
             method="PATCH",
             path=self.path,
             data=update_properties,
@@ -368,7 +368,7 @@ class _PropertyMixin(object):
             if_metageneration_match=if_metageneration_match,
             if_metageneration_not_match=if_metageneration_not_match,
         )
-        api_response = client._connection.api_request(
+        api_response = telemetry_wrapped_api_request(client._connection.api_request,
             method="PUT",
             path=self.path,
             data=self._properties,
--- a/google/cloud/storage/client.py
+++ b/google/cloud/storage/client.py
@@ -45,6 +45,7 @@ from google.cloud.storage.hmac_key import HMACKeyMetadata
 from google.cloud.storage.acl import BucketACL
 from google.cloud.storage.acl import DefaultObjectACL
 from google.cloud.storage.constants import _DEFAULT_TIMEOUT
+from google.cloud.storage._opentelemetry_meter import telemetry_wrapped_api_request
 
 
 _marker = object()
@@ -254,7 +255,7 @@ class Client(ClientWithProject):
         if project is None:
             project = self.project
         path = "/projects/%s/serviceAccount" % (project,)
-        api_response = self._base_connection.api_request(
+        api_response = telemetry_wrapped_api_request(self._base_connection.api_request,
             method="GET", path=path, timeout=timeout
         )
         return api_response["email_address"]
@@ -524,7 +525,7 @@ class Client(ClientWithProject):
         if location is not None:
             properties["location"] = location
 
-        api_response = self._connection.api_request(
+        api_response = telemetry_wrapped_api_request(self._connection.api_request,
             method="POST",
             path="/b",
             query_params=query_params,
@@ -777,7 +778,7 @@ class Client(ClientWithProject):
         if fields is not None:
             extra_params["fields"] = fields
 
-        api_request = functools.partial(self._connection.api_request, timeout=timeout)
+        api_request = functools.partial(telemetry_wrapped_api_request, self._connection.api_request, timeout=timeout)
 
         return page_iterator.HTTPIterator(
             client=self,
@@ -828,7 +829,7 @@ class Client(ClientWithProject):
         if user_project is not None:
             qs_params["userProject"] = user_project
 
-        api_response = self._connection.api_request(
+        api_response = telemetry_wrapped_api_request(self._connection.api_request,
             method="POST", path=path, query_params=qs_params, timeout=timeout
         )
         metadata = HMACKeyMetadata(self)
@@ -893,7 +894,7 @@ class Client(ClientWithProject):
         if user_project is not None:
             extra_params["userProject"] = user_project
 
-        api_request = functools.partial(self._connection.api_request, timeout=timeout)
+        api_request = functools.partial(telemetry_wrapped_api_request, self._connection.api_request, timeout=timeout)
 
         return page_iterator.HTTPIterator(
             client=self,


google/cloud/storage/_opentelemetry_meter.py
```diff
@@ -0,0 +1,55 @@
...
+    HAS_OPENTELEMETRY_INSTALLED = True
+except ImportError:
+    HAS_OPENTELEMETRY_INSTALLED = False
+
+FUNCTION_NAME_KEY = 'instrumented_function_name'
+
+if HAS_OPENTELEMETRY_INSTALLED:
+    meter = metrics.get_meter(__name__)
+    metrics.get_meter_provider().start_pipeline(
+        meter, CloudMonitoringMetricsExporter(), 15
```
sjbrunst, line 33: What is this 15? Can you use a named parameter here?

google/cloud/storage/_opentelemetry_meter.py
```diff
@@ -0,0 +1,58 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Manages OpenTelemetry metrics creation and handling"""
+
+try:
+    from opentelemetry import metrics
+    from opentelemetry.sdk.metrics import Counter, MeterProvider
```
c24t, line 19: Only a theoretical concern right now since the built in SDK is the _only_ SDK, but in general libraries like this should only depend on the API package so users can bring their own non-standard SDK if they want.

kintel:
It would be nice with a PR description to define what subset of the scope of this project is implemented here. E.g. are you planning to add traces in addition to metrics later, are you planning any follow-up PRs for this integration? A separate design doc would of course be great.

google/cloud/storage/_opentelemetry_meter.py
AndrewAXue: Made it named. This number denotes how frequently the Exporter thread will export things (once every X seconds).

google/cloud/storage/_opentelemetry_meter.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+
+    HAS_OPENTELEMETRY_INSTALLED = True
+except ImportError:
+    HAS_OPENTELEMETRY_INSTALLED = False
+
+FUNCTION_NAME_KEY = 'instrumented_function_name'
+
+if HAS_OPENTELEMETRY_INSTALLED:
+    metrics.set_meter_provider(MeterProvider(stateful=False))
```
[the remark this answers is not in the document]
AndrewAXue: Changed. The user is now in charge of creating a meter provider (the "opt-in" mechanism).

google/cloud/storage/_opentelemetry_meter.py
```diff
@@ -0,0 +1,58 @@
...
+    meter = metrics.get_meter(__name__)
+    metrics.get_meter_provider().start_pipeline(
+        meter, CloudMonitoringMetricsExporter(), 15
+    )
+    requests_counter = meter.create_metric(
+        name="GCS_request_counter",
+        description="number of requests",
+        unit="1",
+        value_type=int,
+        metric_type=Counter,
```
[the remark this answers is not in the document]
AndrewAXue: Discussed offline, for now this needs to be done.

google/cloud/storage/_opentelemetry_meter.py
AndrewAXue: Discussed offline. For now we will actually need the sdk to be imported.

README_OPENTELEMETRY.rst
```diff
@@ -0,0 +1,40 @@
...
+    # VM instacne that belongs to the project "cloudtop:prod" which you likely
+    # don't have read access for
+    #==============================================================================
+
+    # If you don't use resources, remove resource=resources
+    metrics.set_meter_provider(MeterProvider(stateful=False, resource=resources))
+
+    # Normal code below. This is your application / script
+
+Check out captured metrics at `Cloud Monitoring <https://pantheon.corp.google.com/monitoring/metrics-explorer>`_.
```
crwilcox, line 40: change to external link

google/cloud/storage/blob.py
```diff
@@ -888,7 +890,7 @@ def _do_download(
             )
 
             while not download.finished:
-                download.consume_next_chunk(transport, timeout=timeout)
+                telemetry_wrapped_api_request(download.consume_next_chunk, transport, timeout=timeout, instrumented_function_name='GET /b/{}/o/{}'.format(self.bucket.name, self.name))
```
crwilcox, line 893: NIT: this line may have issue with linting due to length, could this be broken down into shorter statements by chance?

tests/unit/test_client.py
```diff
@@ -1114,7 +1129,7 @@ def test_list_blobs_w_all_arguments_and_user_project(self):
             blobs = list(iterator)
 
             self.assertEqual(blobs, [])
-            connection.api_request.assert_called_once_with(
+            connection.api_request.args[0].assert_called_once_with(
```
crwilcox, line 1132: This test seems to indicate that with telemetry the return type is different than previous. Is this the case with and without telemetry enabled, or is it possible the type of api_request could be different?

google/cloud/storage/blob.py
```diff
@@ -1667,7 +1669,7 @@ def _do_resumable_upload(
         )
 
         while not upload.finished:
-            response = upload.transmit_next_chunk(transport, timeout=timeout)
+            response = telemetry_wrapped_api_request(upload.transmit_next_chunk, transport, timeout=timeout, instrumented_function_name='upload')
```
crwilcox, line 1672: could you explain why in some cases we call the wrapped function and provide args, but in other cases we are using a partial? It seems a bit strange to me that both varieties are being used, but I think I may be missing a subtle detail.

README_OPENTELEMETRY.rst
```diff
@@ -0,0 +1,40 @@
+Metrics with OpenTelemetry
+==================================
+This library uses `OpenTelemetry <https://opentelemetry.io/>`_ to record metric information on API requests.
+For information on the benefits and utility of tracing, see the `Cloud Monitoring docs <https://cloud.google.com/monitoring/docs>`_.
+
+We first need to install opentelemetry:
+
+.. code-block:: sh
+
+    pip install opentelemetry-api
```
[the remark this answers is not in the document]
AndrewAXue: Yeah so opentelemetry is split into an api and sdk, and we'll need both (unless the user decides to make their own SDK). The opentelemetry-exporter-google-cloud is the package containining the Cloud Monitoring Exporter which we'll need as well. opentelemetry-tools-google-cloud is the package that contains the Cloud Resource Detector, which I suppose is optional. I'll mark it as such.

google/cloud/storage/blob.py
AndrewAXue: linted

tests/unit/test_client.py
AndrewAXue: Yeah, so with or without telemetry we will do 
`value.api_request = functools.partial(telemetry_wrapped_api_request, value.api_request)`
Its just that with telemetry enabled, `telemetry_wrapped_api_request` will actually do stuff, whereas if telemetry is not enable its just a wrapper that calls the underlying api_request. In either case we need to call `args[0]` on the partial to get to the underlying api_request.

google/cloud/storage/blob.py
AndrewAXue: Most client functions will eventually call `client._connection.api_request`. The exceptions to this are `_do_download` and `upload` which will, instead of using `client._connection.api_request`, pass in the raw transport and use another unrelated method. I'm a bit fuzzy on the details maybe @frankyn can explain better.

tests/unit/test_client.py
crwilcox: As we add type-hints I wonder if this will surface in an odd way.

README_OPENTELEMETRY.rst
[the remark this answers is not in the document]
crwilcox: If you can relay, this is a large number of dependencies to add to a project. I think it may more than double the dependencies for this library.

google/cloud/storage/_opentelemetry_meter.py
```diff
@@ -21,32 +21,40 @@
...
+if OPENTELEMETRY_READY:
+
     meter = metrics.get_meter(__name__)
-    metrics.get_meter_provider().start_pipeline(
+    meter_provider.start_pipeline(
         meter, CloudMonitoringMetricsExporter(), 15
     )
     requests_counter = meter.create_metric(
-        name="GCS_request_counter",
+        name="other_demo",
```
crwilcox, line 37: is this an intentional change?

AndrewAXue: whoops, reverted

google/cloud/storage/_opentelemetry_meter.py
```diff
@@ -0,0 +1,63 @@
...
+    OPENTELEMETRY_READY = not isinstance(meter_provider, metrics.DefaultMeterProvider)
+except ImportError:
+    OPENTELEMETRY_READY = False
+
+FUNCTION_NAME_KEY = "instrumented_function_name"
+
+if OPENTELEMETRY_READY:
+
+    meter = metrics.get_meter(__name__)
+    meter_provider.start_pipeline(meter, CloudMonitoringMetricsExporter(), interval=15)
```
kintel, line 35: This is likely out of scope for this PR, but would be valuable to address, perhaps in the design doc:
What if the GCS customer in question does not want to export directly for Cloud Monitoring?
-> they may be using a 3rd party monitoring tool, or they may want to export to the OpenTelemetry collector (for performance or security reasons).

Does (or should) OpenTelemetry offer any ways of doing this?

AndrewAXue: So because in this case, we do always want to export to Cloud Monitoring we'll keep this code. The user can just add in their own code
`meter_provider.start_pipeline(meter, OtherExporter(), interval=15)`
To make it export to another backend in addition to Cloud Monitoring.

kintel: Where would users add that code? This is a private initializer inside this library. I wouldn't expect users to want to fork and modify this library. I believe exporting to the collector will be very common, especially in security-hardened deployments where you don't want components to be able to make arbitrary network connection to the internet.

google/cloud/storage/_opentelemetry_meter.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    def telemetry_wrapped_api_request(api_request, *args, **kwargs):
+        instrumented_labels = {}
+        if FUNCTION_NAME_KEY in kwargs:
+            instrumented_labels["function_name"] = kwargs[FUNCTION_NAME_KEY]
+            kwargs.pop(FUNCTION_NAME_KEY)
+        elif "path" in kwargs:
+            instrumented_labels["function_name"] = "{} {}".format(
+                kwargs["method"], kwargs["path"]
```
kintel, line 51: Here we're assuming that `method` is always available. This seem to hold for all current direct calls to api_client, but there is no explicit `method` argument to cases where a HTTPIterator is used. It's possible that this gets added behind the scenes.

Related to this, it looks like this integration isn't tested for the cases where an HTTPIterator is used to wrap an API request.

google/cloud/storage/blob.py
```diff
@@ -869,7 +870,12 @@ def _do_download(
             download = klass(
                 download_url, stream=file_obj, headers=headers, start=start, end=end
             )
-            response = download.consume(transport, timeout=timeout)
+            response = telemetry_wrapped_api_request(
```
kintel, line 873: I'm not familiar with this code, but it isn't obvious to me that the blob telemetry wrappers are tested.

google/cloud/storage/_opentelemetry_meter.py
AndrewAXue: Sorry, to be clear it would look like
```
# Imports
meter_provider = MeterProvider(stateful=False)
metrics.set_meter_provider(meter_provider)
meter_provider.start_pipeline(otherexporter()...)

# user's logic
```
Note that the only new line is `meter_provider.start_pipeline(otherexporter()...)`

tests/unit/test_client.py
kintel: Another way of accomplishing this could be to mock Connection.api_request, and assert that it gets called, instead of relying on the exact structure of the wrapper object, but then you'd probably also want to assert that the underlying meter got touched.

google/cloud/storage/_opentelemetry_meter.py
kintel: Summary from off-line conversation: The current implementation is good: This is a Cloud Monitoring-only integration and the users opt in by installing the Cloud Monitoring exporter themselves. The larger question is whether we should allow the customer to configure their own exporter. That is not supported by the OpenTelemetry API and thus out of scope for this PR.

README_OPENTELEMETRY.rst
```diff
@@ -0,0 +1,45 @@
+Metrics with OpenTelemetry
+==================================
+This library uses `OpenTelemetry <https://opentelemetry.io/>`_ to record metric information on API requests.
+For information on the benefits and utility of tracing, see the `Cloud Monitoring docs <https://cloud.google.com/monitoring/docs>`_.
```
sjbrunst, line 4: nit: this PR adds metrics but not tracing yet; is "tracing" the right word here?

README_OPENTELEMETRY.rst
```diff
@@ -0,0 +1,45 @@
...
+
+    #==============================================================================
+    # OPTIONAL: These lines of code will detect + scrape resource info from the env/metadata.
+    # This info will then be passed down the line and added to metric info
+    from opentelemetry.tools.resource_detector import GoogleCloudResourceDetector
+    resources = GoogleCloudResourceDetector().detect()
+
+    # Note that this will NOT work if you are on a cloudtop, as you are on a
+    # VM instance that belongs to the project "cloudtop:prod" which you likely
+    # don't have Cloud Monitoring write access to
```
sjbrunst, line 37:
```suggestion
    # don't have Cloud Monitoring write access to.
```

README_OPENTELEMETRY.rst
```diff
@@ -0,0 +1,45 @@
...
+.. code-block:: sh
+
+    pip install opentelemetry-tools-google-cloud
+
+Then, at the beginning of your application create a MeterProvider. Example:
+
+.. code:: python
+
+    # "opt-in" for opentelemtry metrics capturing.
+    # Requires you to have installed opentelemetry packages and called set_meter_provider
```
sjbrunst, line 25:
```suggestion
    # Requires you to have installed opentelemetry packages and called set_meter_provider.
```

README_OPENTELEMETRY.rst
```diff
@@ -0,0 +1,45 @@
...
+    from opentelemetry import metrics
+    from opentelemetry.sdk.metrics import MeterProvider
+
+    #==============================================================================
+    # OPTIONAL: These lines of code will detect + scrape resource info from the env/metadata.
+    # This info will then be passed down the line and added to metric info
```
sjbrunst, line 31:
```suggestion
    # This info will then be passed down the line and added to metric info.
```

README_OPENTELEMETRY.rst
```diff
@@ -0,0 +1,45 @@
...
+
+.. code:: python
+
+    # "opt-in" for opentelemtry metrics capturing.
+    # Requires you to have installed opentelemetry packages and called set_meter_provider
+    from opentelemetry import metrics
+    from opentelemetry.sdk.metrics import MeterProvider
+
+    #==============================================================================
+    # OPTIONAL: These lines of code will detect + scrape resource info from the env/metadata.
```
sjbrunst, line 30: We can probably shorten this a bit
```suggestion
    # OPTIONAL: These lines of code will scrape resource info from the env/metadata.
```

README_OPENTELEMETRY.rst
```diff
@@ -0,0 +1,45 @@
...
+    # This info will then be passed down the line and added to metric info
+    from opentelemetry.tools.resource_detector import GoogleCloudResourceDetector
+    resources = GoogleCloudResourceDetector().detect()
+
+    # Note that this will NOT work if you are on a cloudtop, as you are on a
+    # VM instance that belongs to the project "cloudtop:prod" which you likely
+    # don't have Cloud Monitoring write access to
+    #==============================================================================
+
+    # If you don't want to capture resources, remove resource=resources
```
sjbrunst, line 40:
```suggestion
    # If you don't want to capture resources, remove 'resource=resources'.
```

README_OPENTELEMETRY.rst
```diff
@@ -0,0 +1,45 @@
...
+    metrics.set_meter_provider(MeterProvider(stateful=False, resource=resources))
+
+    # Normal code below. This is your application / script
```
sjbrunst, line 43:
```suggestion
    # Normal code below. This is your application / script.
```

google/cloud/storage/_opentelemetry_meter.py
```diff
@@ -0,0 +1,63 @@
...
+
+try:
+    from opentelemetry import metrics
+    from opentelemetry.sdk.metrics import Counter
+    from opentelemetry.exporter.cloud_monitoring import CloudMonitoringMetricsExporter
+
+    meter_provider = metrics.get_meter_provider()
+
+    # If opentelemetry packages have been installed and a MeterProvider setup
+    #   we are ready to use opentelemetry
```
sjbrunst, line 25:
```suggestion
    #   then we are ready to use opentelemetry.
```

google/cloud/storage/blob.py
AndrewAXue: added test for this

google/cloud/storage/_opentelemetry_meter.py
AndrewAXue: added this test

PR #235, opened 2020-08-07
feat: error message return from api

### LINKED ISSUE #224: Except with error messages returned from the API
**Is your feature request related to a problem? Please describe.**

Currently the implementation will raise an exception if the answer is not 200. https://github.com/googleapis/google-resumable-media-python/blob/master/google/resumable_media/_helpers.py#L70-L98

The current implementation is terrible for development / problem triaging in production systems because it shadows the real problem making no use of the error message explaining the issue.

 **Describe the solution you'd like**

To show the messages the API return

 **Describe alternatives you've considered**

Introspecting the stack to get the message out

HemangChothani:
Fixes #224  🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -3427,6 +3427,9 @@ def _raise_from_invalid_response(error):
              to the failed status code
     """
     response = error.response
+
+    # The 'response.text' gives the actual reason of error, where 'error' gives
+    # the message of expected status code.
     if response.text:
         error_message = response.text + ": " + str(error)
     else:
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -804,7 +804,7 @@ class Blob(_PropertyMixin):
 
         digests = {}
         for encoded_digest in x_goog_hash.split(","):
-            match = re.match(r"(crc32c|md5)=([\w\d]+)==", encoded_digest)
+            match = re.match(r"(crc32c|md5)=([\w\d/]+={0,3})", encoded_digest)
             if match:
                 method, digest = match.groups()
                 digests[method] = digest
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -3433,7 +3433,7 @@ def _raise_from_invalid_response(error):
     if response.text:
         error_message = response.text + ": " + str(error)
     else:
-        error_message = "unknown error: " + str(error)
+        error_message = str(error)
 
     message = u"{method} {url}: {error}".format(
         method=response.request.method, url=response.request.url, error=error_message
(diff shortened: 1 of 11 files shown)


tseaver:
Nit:  please drop the `(apiname)` part of the prefix for your commit summaries.  That convention made sense in the mono-repository, but post-repo-split, it is redundant and cluttery.

google/cloud/storage/blob.py
```diff
@@ -3253,7 +3253,10 @@ def _raise_from_invalid_response(error):
              to the failed status code
     """
     response = error.response
-    error_message = str(error)
+    if response.text:
+        error_message = response.text + ": " + str(error)
+    else:
+        error_message = "unknown error: " + str(error)
```
tseaver, line 3259: I'd prefer not to add the prefix here.  @frankyn WDYT?

frankyn: Could you clarify the difference between response.text and error (add comments I suspect others will ask this question when reviewing the code in the future). 

I also think repsonse.text could be added to the line below when constructing the `message` value instead of this conditional.

HemangChothani: @frankyn  I have added condition here to match the method which is responsible for other http responses and errors declared in `pytho-api-core.exceptions.from_http_response`  you can find here https://github.com/googleapis/python-api-core/blob/622931721ce34839d630aa1e974c7d8f47b5d25e/google/api_core/exceptions.py#L390-L411  

`response.text: ` gives the actual reason of error
`str(error): ` gives ('Request failed with status code', 404(Error Code), 'Expected one of', <HTTPStatus.OK: 200>, <HTTPStatus.PARTIAL_CONTENT: 206>)

I think here we can remove `str(error)` as it's not meaning full for the users, also not included in other http methods from python-api-core as mention above.

HemangChothani: @frankyn  I think it's necessary to add actual reason in error message, Ex: in the https://github.com/googleapis/python-storage/issues/249  issue user share a stack trace of error but it's hard to debug as actual reason isn't there in message.

frankyn: I think I need to pull in @andrewsg on this one when he has a moment.

Andrew do you have guidance on raising the actual error that occurred in this case?

andrewsg: If I understand correctly, the problem here is the same as I've observed elsewhere where the API responds with human-readable text which is not converted into a specific error message by our client library. Prepending this human-readable text does make sense though I'm neutral on whether we should include the "unknown error" prepending in the `else` case.

We should make sure that the traceback still includes the original function call that caused the error.

frankyn: Thanks @andrewsg.

I don't see the value of `unknown error:` and could be confusing otherwise. Please only preppend `response.text` when it's available. Remove `unknown error:`.

PR #238, opened 2020-08-11
fix: extract hashes correctly during download

HemangChothani:
Fixes #237  🦕

Issue #237, opened 2020-08-10
Multiple upload and downloads fail


AndrewAXue:
#### Environment details

  - OS type and version: Linux rodete
  - Python version: Python 3.8.5
  - pip version: pip 20.0.2
  - `google-cloud-storage` version: 1.30.0

#### Steps to reproduce

Run:

```python
from google.cloud import storage
client = storage.Client()
bucket = client.get_bucket('fishbucket1')
blob2 = bucket.blob('file')
blob2.upload_from_string('hi')
blob2.download_as_string()
blob2.upload_from_string('hi')
```

#### Stack trace
```
Traceback (most recent call last):
  File "/usr/local/google/home/aaxue/Documents/google/temp/lib/python3.8/site-packages/google/cloud/storage/blob.py", line 1910, in upload_from_file
    created_json = self._do_upload(
  File "/usr/local/google/home/aaxue/Documents/google/temp/lib/python3.8/site-packages/google/cloud/storage/blob.py", line 1758, in _do_upload
    response = self._do_multipart_upload(
  File "/usr/local/google/home/aaxue/Documents/google/temp/lib/python3.8/site-packages/google/cloud/storage/blob.py", line 1404, in _do_multipart_upload
    response = upload.transmit(
  File "/usr/local/google/home/aaxue/Documents/google/temp/lib/python3.8/site-packages/google/resumable_media/requests/upload.py", line 148, in transmit
    self._process_response(response)
  File "/usr/local/google/home/aaxue/Documents/google/temp/lib/python3.8/site-packages/google/resumable_media/_upload.py", line 116, in _process_response
    _helpers.require_status_code(response, (http_client.OK,), self._get_status_code)
  File "/usr/local/google/home/aaxue/Documents/google/temp/lib/python3.8/site-packages/google/resumable_media/_helpers.py", line 106, in require_status_code
    raise common.InvalidResponse(
google.resumable_media.common.InvalidResponse: ('Request failed with status code', 400, 'Expected one of', <HTTPStatus.OK: 200>)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "t", line 7, in <module>
    blob2.upload_from_string('hi')
  File "/usr/local/google/home/aaxue/Documents/google/temp/lib/python3.8/site-packages/google/cloud/storage/blob.py", line 2097, in upload_from_string
    self.upload_from_file(
[...]

PR #240, opened 2020-08-12
fix(blob): extract padding chars from hashes

william-silversmith:
I didn't previously understand that the "==" at the end of the hash
isn't merely a delimiter, but a padding character for base64 encoding.
It's not strictly necessary to have it, but you have to add it back in
to perform a base64 decode. Base64 encoded strings are padded out to
a multiple of four characters.

https://en.wikipedia.org/wiki/Base64#Output_padding


william-silversmith:
Looks like #238 and #237 are related.

tseaver:
Thanks for the fix!  Closing because #238 is already reviewed, and adds a system test.

Issue #190, opened 2020-06-23
test_system.TestKMSIntegration.test_bucket_w_default_kms_key_name failed


IlyaFaer:
Got a flaky error on kokoro checks. Failed only once, on Python 2.7 session. Trying to reproduce locally - no results so far.

```python
____________ TestKMSIntegration.test_bucket_w_default_kms_key_name _____________

self = <test_system.TestKMSIntegration testMethod=test_bucket_w_default_kms_key_name>

    def test_bucket_w_default_kms_key_name(self):
        BLOB_NAME = "default-kms-key-name"
        OVERRIDE_BLOB_NAME = "override-default-kms-key-name"
        ALT_BLOB_NAME = "alt-default-kms-key-name"
        CLEARTEXT_BLOB_NAME = "cleartext"

        file_data = self.FILES["simple"]

        with open(file_data["path"], "rb") as _file_data:
            contents = _file_data.read()

        kms_key_name = self._kms_key_name()
        self.bucket.default_kms_key_name = kms_key_name
        self.bucket.patch()
        self.assertEqual(self.bucket.default_kms_key_name, kms_key_name)

        defaulted_blob = self.bucket.blob(BLOB_NAME)
        defaulted_blob.upload_from_filename(file_data["path"])
        self.case_blobs_to_delete.append(defaulted_blob)

        self.assertEqual(defaulted_blob.download_as_string(), contents)
        # We don't know the current version of the key.
        self.assertTrue(defaulted_blob.kms_key_name.startswith(kms_key_name))

        alt_kms_key_name = self._kms_key_name(self.ALT_KEY_NAME)

        override_blob = self.bucket.blob(
            OVERRIDE_BLOB_NAME, kms_key_name=alt_kms_key_name
        )
        override_blob.upload_from_filename(file_data["path"])
        self.case_blobs_to_delete.append(override_blob)

        self.assertEqual(override_blob.download_as_string(), contents)
        # We don't know the current version of the key.
        self.assertTrue(override_blob.kms_key_name.startswith(alt_kms_key_name))

        self.bucket.default_kms_key_name = alt_kms_key_name
        self.bucket.patch()

        alt_blob = self.bucket.blob(ALT_BLOB_NAME)
        alt_blob.upload_from_filename(file_data["path"])
        self.case_blobs_to_delete.append(alt_blob)

        self.assertEqual(alt_blob.download_as_string(), contents)
        # We don't know the current version of the key.
[...]

PR #245, opened 2020-08-14
test: add retry for bucket cmek enabled system test

### LINKED ISSUE #239: 'TestKMSIntegration.test_upload_new_blob_w_bucket_cmek_enabled' flakes w/ EC error
From [this Kokoro build failure](https://source.cloud.google.com/results/invocations/27fb932e-3234-4aac-a3a2-3715a0b6e76e/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-storage%2Fpresubmit%2Fpresubmit/log):

```python
________ TestKMSIntegration.test_upload_new_blob_w_bucket_cmek_enabled _________

self = <test_system.TestKMSIntegration testMethod=test_upload_new_blob_w_bucket_cmek_enabled>

    def test_upload_new_blob_w_bucket_cmek_enabled(self):
        blob_name = "test-blob"
        payload = b"DEADBEEF"
        alt_payload = b"NEWDEADBEEF"

        kms_key_name = self._kms_key_name()
        self.bucket.default_kms_key_name = kms_key_name
        self.bucket.patch()
        self.assertEqual(self.bucket.default_kms_key_name, kms_key_name)

        blob = self.bucket.blob(blob_name)
        blob.upload_from_string(payload)
        # We don't know the current version of the key.
>       self.assertTrue(blob.kms_key_name.startswith(kms_key_name))
E       AttributeError: 'NoneType' object has no attribute 'startswith'

tests/system/test_system.py:2029: AttributeError
```

Similar to #190.  In this case, reloading the blob after a failure would likely catch up to the expected state.

HemangChothani:
Fixes #239 

Tried to reproduce locally but no results so far, so adding a retry for the test to avoid flaky failures.

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -2013,6 +2013,7 @@ class TestKMSIntegration(TestStorageFiles):
 
         self.assertEqual(dest.download_as_bytes(), source_data)
 
+    @RetryErrors(unittest.TestCase.failureException)
     def test_upload_new_blob_w_bucket_cmek_enabled(self):
         blob_name = "test-blob"
         payload = b"DEADBEEF"
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -2013,7 +2013,6 @@ class TestKMSIntegration(TestStorageFiles):
 
         self.assertEqual(dest.download_as_bytes(), source_data)
 
-    @RetryErrors(unittest.TestCase.failureException)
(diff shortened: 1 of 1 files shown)


HemangChothani:
@tseaver  If this solution is good enough then can we also change in  https://github.com/googleapis/python-storage/issues/190, they had fixed it with retrying the whole test  for `test_bucket_w_default_kms_key_name`  system test.

PR #250, opened 2020-08-19
docs: add docs signed_url expiration take default utc

### LINKED ISSUE #244: generate_signed_post_policy_v4() improperly handles expiration timestamps with tzinfo=None
The `generate_signed_post_policy_v4()` workflow, using `get_expiration_seconds_v4()` (`get_expiration_seconds_v2()` may also be affected) improperly calculates the expiration seconds if a datetime without tzinfo is used.

For instance, Python's `datetime.datetime.now()` produces a datetime with `tzinfo=None`, but in local time. The `get_expiration_seconds_v4()` method replaces that "None" timezone with a UTC tzinfo, which results in an inaccurate timestamp that was generated for local time but is now presented as UTC.

There is no way to safely handle `tzinfo=None` user input in Python. The library should either document that `tzinfo=None` will be treated as either local time or UTC (current behavior, UTC) or else it should reject timestamps without tzinfo for safety.

The system test `TestV4POSTPolicies.test_get_signed_policy_v4()` will also fail when run in a timezone west of the equator and should be modified along with the above code. Warning: it is possible Kokoro is not running in a timezone west of the equator and will not manifest this error.

HemangChothani:
Fixes #244

### DIFF
--- a/google/cloud/storage/_signing.py
+++ b/google/cloud/storage/_signing.py
@@ -92,8 +92,8 @@ def get_expiration_seconds_v2(expiration):
 
     :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
     :param expiration: Point in time when the signed URL should expire. If
-                       expiration in datetime, then need to pass timezone
-                       with it. Default value of ``tzinfo`` is ``UTC``.
+                       a ``datetime`` instance is passed without an explicit
+                       ``tzinfo`` set,  it will be converted to ``UTC``.
 
     :raises: :exc:`TypeError` when expiration is not a valid type.
 
@@ -126,8 +126,8 @@ def get_expiration_seconds_v4(expiration):
 
     :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
     :param expiration: Point in time when the signed URL should expire. If
-                       expiration in datetime, then need to pass timezone
-                       with it. Default value of ``tzinfo`` is ``UTC``.
+                       a ``datetime`` instance is passed without an explicit
+                       ``tzinfo`` set,  it will be converted to ``UTC``.
 
     :raises: :exc:`TypeError` when expiration is not a valid type.
     :raises: :exc:`ValueError` when expiration is too large.
@@ -304,8 +304,8 @@ def generate_signed_url_v2(
 
     :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
     :param expiration: Point in time when the signed URL should expire. If
-                       expiration in datetime, then need to pass timezone with
-                       it. Default value of ``tzinfo`` is ``UTC``.
+                       a ``datetime`` instance is passed without an explicit
+                       ``tzinfo`` set,  it will be converted to ``UTC``.
 
     :type api_access_endpoint: str
     :param api_access_endpoint: (Optional) URI base. Defaults to empty string.
@@ -468,8 +468,8 @@ def generate_signed_url_v4(
 
     :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
     :param expiration: Point in time when the signed URL should expire. If
-                       expiration in datetime, then need to pass timezone with
-                       it. Default value of ``tzinfo`` is ``UTC``.
+                       a ``datetime`` instance is passed without an explicit
+                       ``tzinfo`` set,  it will be converted to ``UTC``.
 
     :type api_access_endpoint: str
     :param api_access_endpoint: (Optional) URI base. Defaults to
--- a/google/cloud/storage/_signing.py
+++ b/google/cloud/storage/_signing.py
@@ -91,7 +91,9 @@ def get_expiration_seconds_v2(expiration):
     """Convert 'expiration' to a number of seconds in the future.
 
     :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
-    :param expiration: Point in time when the signed URL should expire.
+    :param expiration: Point in time when the signed URL should expire. If
+                       expiration in datetime, then need to pass timezone
+                       with it. Default value of ``tzinfo`` is ``UTC``.
 
     :raises: :exc:`TypeError` when expiration is not a valid type.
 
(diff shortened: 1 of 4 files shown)


google/cloud/storage/_signing.py
```diff
@@ -91,7 +91,9 @@ def get_expiration_seconds_v2(expiration):
     """Convert 'expiration' to a number of seconds in the future.
 
     :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
-    :param expiration: Point in time when the signed URL should expire.
+    :param expiration: Point in time when the signed URL should expire. If
+                       expiration in datetime, then need to pass timezone
+                       with it. Default value of ``tzinfo`` is ``UTC``.
```
tseaver, line 96: The default value of the `tzinfo` argument to `datetime.datetime` is None, resulting in a "naive" `datetime` instance:  the comment should state clearly that we forcibly convert such instances to `UTC`.  E.g.:
```python
    :param expiration: Point in time when the signed URL should expire. If
                       a `datetime` instance is passed without an explicit
                      `tzinfo` set,  it will be converted to ``UTC``.
```

andrewsg: Rather than "it will be converted to" let's use "it will be assumed to be". This is a huge footgun; a user who sees the term "converted" will assume that we will do the right thing in the default case, whereas in fact we will almost always do the wrong thing.

andrewsg:
Cancel that, I just saw that the test issue is resolved in #251. Thanks for that! Leaving review on "changes requested" for the docstring update as per my comment; lgtm other than that.

google/cloud/storage/_signing.py
```diff
@@ -92,8 +92,8 @@ def get_expiration_seconds_v2(expiration):
 
     :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
     :param expiration: Point in time when the signed URL should expire. If
-                       expiration in datetime, then need to pass timezone
-                       with it. Default value of ``tzinfo`` is ``UTC``.
+                       a ``datetime`` instance is passed without an explicit
+                       ``tzinfo`` set,  it will be converted to ``UTC``.
```
tseaver: Per @andrewsg 
```suggestion
                       ``tzinfo`` set,  it will be assumed to be ``UTC``.
```

google/cloud/storage/_signing.py
```diff
@@ -126,8 +126,8 @@ def get_expiration_seconds_v4(expiration):
...
```
tseaver: Per @andrewsg 
```suggestion
                       ``tzinfo`` set,  it will be assumed to be ``UTC``.
```

google/cloud/storage/_signing.py
```diff
@@ -304,8 +304,8 @@ def generate_signed_url_v2(
 
     :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
     :param expiration: Point in time when the signed URL should expire. If
-                       expiration in datetime, then need to pass timezone with
-                       it. Default value of ``tzinfo`` is ``UTC``.
+                       a ``datetime`` instance is passed without an explicit
+                       ``tzinfo`` set,  it will be converted to ``UTC``.
```
tseaver: Per @andrewsg 
```suggestion
                       ``tzinfo`` set,  it will be assumed to be ``UTC``.
```

google/cloud/storage/_signing.py
```diff
@@ -468,8 +468,8 @@ def generate_signed_url_v4(
...
```
tseaver: Per @andrewsg 
```suggestion
                       ``tzinfo`` set,  it will be assumed to be ``UTC``.
```

google/cloud/storage/blob.py
```diff
@@ -412,8 +412,8 @@ def generate_signed_url(
 
         :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
         :param expiration: Point in time when the signed URL should expire. If
-                           expiration in datetime, then need to pass timezone
-                           with it. Default value of ``tzinfo`` is ``UTC``.
+                           a ``datetime`` instance is passed without an explicit
+                           ``tzinfo`` set,  it will be converted to ``UTC``.
```
tseaver: Per @andrewsg 
```suggestion
                       ``tzinfo`` set,  it will be assumed to be ``UTC``.
```

google/cloud/storage/bucket.py
```diff
@@ -2956,8 +2956,8 @@ def generate_signed_url(
 
         :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
         :param expiration: Point in time when the signed URL should expire. If
-                           expiration in datetime, then need to pass timezone
-                           with it. Default value of ``tzinfo`` is ``UTC``.
+                           a ``datetime`` instance is passed without an explicit
+                           ``tzinfo`` set,  it will be converted to ``UTC``.
```
tseaver: Per @andrewsg 
```suggestion
                       ``tzinfo`` set,  it will be assumed to be ``UTC``.
```

google/cloud/storage/client.py
```diff
@@ -962,9 +962,9 @@ def generate_signed_post_policy_v4(
         :param blob_name: Object name.
 
         :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
-        :param expiration: Policy expiration time. If expiration in datetime,
-                           then need to pass timezone with it. Default value of
-                           ``tzinfo`` is ``UTC``.
+        :param expiration: Policy expiration time. If a ``datetime`` instance is
+                           passed without an explicit ``tzinfo`` set,  it will be
+                           converted to ``UTC``.
```
tseaver: Per @andrewsg 
```suggestion
                       ``tzinfo`` set,  it will be assumed to be ``UTC``.
```

tseaver:
@HemangChothani I would merge my suggestions to address the concern expressed by @andrewsg , but you have the "maintainers can push changes" option turned off.

google/cloud/storage/client.py
HemangChothani: All the changes suggested by @andrewsg has been already pushed. I have verified all the files.

tseaver: @HemangChothani see [this comment by @andrewsg ](https://github.com/googleapis/python-storage/pull/250#discussion_r475949117):

> Rather than "it will be converted to" let's use "it will be assumed to be". This is a huge footgun; a user who sees the term "converted" will assume that we will do the right thing in the default case, whereas in fact we will almost always do the wrong thing.

google/cloud/storage/_signing.py
```diff
@@ -91,7 +91,9 @@ def get_expiration_seconds_v2(expiration):
     """Convert 'expiration' to a number of seconds in the future.
 
     :type expiration: Union[Integer, datetime.datetime, datetime.timedelta]
-    :param expiration: Point in time when the signed URL should expire.
+    :param expiration: Point in time when the signed URL should expire. If
+                       a ``datetime`` instance is passed without an explicit
+                       ``tzinfo`` set,  it will be assumed to be ``UTC``.
```
HemangChothani, line 96: I think i have added all the place but don't why it's not reflecting in the `requested cheanges`, but in the file it's there.

PR #258, opened 2020-08-27
fix(blob): base64 includes additional characters

william-silversmith:
Hashes were not being parsed correctly. I forgot that base64 includes the "+" and "/" characters. 

https://en.wikipedia.org/wiki/Base64

image


tseaver (approved):

william-silversmith:
Thank you again Tres! Sorry for the boo boo.

PR #273, opened 2020-09-16
chore: restrict pubsub to  < 2.0

### LINKED ISSUE #272: System test failed for python3.8
- system test failed related to pubsub and due to new version of pusub

HemangChothani:
Fixes #272 

For the passing current scenario restrict version in `nox.py` file and for the future added in synth.py file.

`google-cloud-pubsub==2.0.0` version has dropped the support of python2, so once storage will drop the support following changes needs to be done:

-  replace create_topic(self.topic_path) to create_topic(name=self.topic_path)

- replace get_iam_policy(self.topic_path) to get_iam_policy(request={"resource": self.topic_path})

- replace set_iam_policy(self.topic_path, policy) to set_iam_policy(request={"resource": self.topic_path, "policy": policy})

- replace delete_topic(self.topic_path) to delete_topic(request={"topic":self.topic_path})

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -124,7 +124,7 @@ def system(session):
         "pytest",
         "google-cloud-testutils",
         "google-cloud-iam",
-        "google-cloud-pubsub",
+        "google-cloud-pubsub < 2.0.0",
         "google-cloud-kms < 2.0dev",
     )
     session.install("-e", ".")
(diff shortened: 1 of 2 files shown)


plamut (approved):
This looks good, the dependency should indeed be pinned to exclude the next major version, as major versions are generally expected to contain breaking changes.

PR #287, opened 2020-10-02
feat: retry API calls with exponential backoff

### LINKED ISSUE #108: Add retry (exponential backoff) for storage api functions

andrewsg:
Fixes #108

### DIFF
--- a/google/cloud/storage/_http.py
+++ b/google/cloud/storage/_http.py
@@ -56,7 +56,7 @@ class Connection(_http.JSONConnection):
             # If this is a ConditionalRetryPolicy, check conditions.
             try:
                 retry = retry.get_retry_policy_if_conditions_met(**kwargs)
-            except AttributeError:
+            except AttributeError:  # This is not a ConditionalRetryPolicy.
                 pass
             if retry:
                 call = retry(call)
--- a/google/cloud/storage/_http.py
+++ b/google/cloud/storage/_http.py
@@ -49,7 +49,7 @@ class Connection(_http.JSONConnection):
     API_URL_TEMPLATE = "{api_base_url}/storage/{api_version}{path}"
     """A template for the URL of a particular API call."""
 
-    def api_request(self, *args, retry=None, **kwargs):
+    def api_request(self, retry=None, *args, **kwargs):
         call = functools.partial(super(Connection, self).api_request, *args, **kwargs)
         if retry:
             # If this is a ConditionalRetryPolicy, check conditions.
--- a/google/cloud/storage/_http.py
+++ b/google/cloud/storage/_http.py
@@ -55,9 +55,10 @@ class Connection(_http.JSONConnection):
         if retry:
             # If this is a ConditionalRetryPolicy, check conditions.
             try:
-                retry = retry.get_retry_policy_if_conditions_met(kwargs)
+                retry = retry.get_retry_policy_if_conditions_met(**kwargs)
             except AttributeError:
                 pass
-            call = retry(call)
+            if retry:
+                call = retry(call)
         return call()
 
--- a/google/cloud/storage/_http.py
+++ b/google/cloud/storage/_http.py
@@ -49,7 +49,8 @@ class Connection(_http.JSONConnection):
     API_URL_TEMPLATE = "{api_base_url}/storage/{api_version}{path}"
     """A template for the URL of a particular API call."""
 
-    def api_request(self, retry=None, *args, **kwargs):
+    def api_request(self, *args, **kwargs):
+        retry = kwargs.pop("retry", None)
         call = functools.partial(super(Connection, self).api_request, *args, **kwargs)
         if retry:
             # If this is a ConditionalRetryPolicy, check conditions.
--- a/google/cloud/storage/_http.py
+++ b/google/cloud/storage/_http.py
@@ -14,9 +14,12 @@
 
 """Create / interact with Google Cloud Storage connections."""
 
+import functools
+
 from google.cloud import _http
 
 from google.cloud.storage import __version__
+from google.cloud.storage import retry
 
 
 class Connection(_http.JSONConnection):
@@ -46,3 +49,15 @@ class Connection(_http.JSONConnection):
 
     API_URL_TEMPLATE = "{api_base_url}/storage/{api_version}{path}"
     """A template for the URL of a particular API call."""
+
+    def api_request(self, *args, retry=None, **kwargs):
+        call = functools.partial(super(Connection, self).api_request, *args, **kwargs)
+        if retry:
+            # If this is a ConditionalRetryPolicy, check conditions.
+            try:
+                retry = retry.get_retry_policy_if_conditions_met(kwargs)
+            except AttributeError:
+                pass
+            call = retry(call)
+        return call()
+
--- a/google/cloud/storage/_http.py
+++ b/google/cloud/storage/_http.py
@@ -61,4 +61,3 @@ class Connection(_http.JSONConnection):
             if retry:
                 call = retry(call)
         return call()
-
--- a/google/cloud/storage/_http.py
+++ b/google/cloud/storage/_http.py
@@ -19,7 +19,6 @@ import functools
 from google.cloud import _http
 
 from google.cloud.storage import __version__
-from google.cloud.storage import retry
 
 
 class Connection(_http.JSONConnection):
--- a/google/cloud/storage/retry.py
+++ b/google/cloud/storage/retry.py
@@ -28,6 +28,7 @@ _UNSTRUCTURED_RETRYABLE_TYPES = (
     exceptions.BadGateway,
 )
 
+
 # FIXME: needs to be brought in line with doc outlining all retriable error codes
 # FIXME: add tests
 def _should_retry(exc):
--- a/google/cloud/storage/retry.py
+++ b/google/cloud/storage/retry.py
@@ -12,36 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import requests
+
 from google.api_core import exceptions
 from google.api_core import retry
 
 import json
 
 
-_RETRYABLE_REASONS = frozenset(
-    ["rateLimitExceeded", "backendError", "internalError", "badGateway"]
+_RETRYABLE_TYPES = (
+    exceptions.TooManyRequests,  # 429
+    exceptions.InternalServerError,  # 500
+    exceptions.BadGateway,  # 502
+    exceptions.ServiceUnavailable,  # 503
+    exceptions.GatewayTimeout,  # 504
+    requests.ConnectionError,
 )
 
-_UNSTRUCTURED_RETRYABLE_TYPES = (
-    exceptions.TooManyRequests,
-    exceptions.InternalServerError,
-    exceptions.BadGateway,
-)
 
-
-# FIXME: needs to be brought in line with doc outlining all retriable error codes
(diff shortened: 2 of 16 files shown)


tests/unit/test__http.py
```diff
@@ -81,3 +81,115 @@ def test_build_api_url_w_extra_query_params(self):
...
+        import requests
+
+        http = mock.create_autospec(requests.Session, instance=True)
+        client = mock.Mock(_http=http, spec=["_http"])
+
+        # Some of this is unnecessary if the test succeeds, but we'll leave it
+        # to ensure a failure produces a less confusing error message.
+        conn = self._make_one(client)
+        response = requests.Response()
+        response.status_code = 200
```
frankyn, line 153: I think we discussed but re-raising:

Can you validate that response.status_code 503 will call retry?

frankyn (changes requested):
Thank you for your patience, it took a bit more time than I expected to review the PR.

andrewsg:
@tritone FYI

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,102 @@
...
+
+# FIXME: needs to be brought in line with doc outlining all retriable error codes
+# FIXME: add tests once above is done
+def _should_retry(exc):
+    """Predicate for determining when to retry."""
+    if not hasattr(exc, "errors"):
+        return False
+
+    if len(exc.errors) == 0:
+        # Check for unstructured error returns, e.g. from GFE
```
crwilcox, line 40: unabbreviate for public understanding?

tests/unit/test__http.py
```diff
@@ -81,3 +81,115 @@ def test_build_api_url_w_extra_query_params(self):
...
+
+        http = mock.create_autospec(requests.Session, instance=True)
+        client = mock.Mock(_http=http, spec=["_http"])
+
+        conn = self._make_one(client)
+        response = requests.Response()
+        response.status_code = 200
+        data = b"brent-spiner"
+        response._content = data
+        http.request.return_value = response
```
crwilcox, line 96: this test is missing the comment the other tests have. There may be a way to refactor this also, but unsure if it is warranted.

google/cloud/storage/_http.py
```diff
@@ -46,3 +48,15 @@ def __init__(self, client, client_info=None, api_endpoint=DEFAULT_API_ENDPOINT):
...
     API_URL_TEMPLATE = "{api_base_url}/storage/{api_version}{path}"
     """A template for the URL of a particular API call."""
+
+    def api_request(self, *args, retry=None, **kwargs):
+        call = functools.partial(super(Connection, self).api_request, *args, **kwargs)
+        if retry:
+            # If this is a ConditionalRetryPolicy, check conditions.
+            try:
+                retry = retry.get_retry_policy_if_conditions_met(**kwargs)
+            except AttributeError:
```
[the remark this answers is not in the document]
andrewsg: Duck typing: if we get an AttributeError here it is because the retry policy passed in was a straight run-of-the-mill retry policy, and not a ConditionalRetryPolicy. We could just do an isinstance check, but duck typing is usually preferred in Python. I'll document this in a comment.

tests/unit/test__http.py
andrewsg: Which comment is missing?

google/cloud/storage/retry.py
andrewsg: Removed (this was part of google-api-core; my hands-on testing suggests the entire section is unneeded for GCS)

tests/unit/test__http.py
andrewsg: Now validated.

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,102 @@
...
+    reason = exc.errors[0]["reason"]
+    return reason in _RETRYABLE_REASONS
+
+
+DEFAULT_RETRY = retry.Retry(predicate=_should_retry)
+"""The default retry object.
+
+To modify the default retry behavior, call a ``with_XXX`` method
+on ``DEFAULT_RETRY``. For example, to change the deadline to 30 seconds,
+pass ``retry=DEFAULT_RETRY.with_deadline(30)``.
```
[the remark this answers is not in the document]
andrewsg: Done.

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,102 @@
...
+
+_UNSTRUCTURED_RETRYABLE_TYPES = (
+    exceptions.TooManyRequests,
+    exceptions.InternalServerError,
+    exceptions.BadGateway,
+)
+
+
+# FIXME: needs to be brought in line with doc outlining all retriable error codes
+# FIXME: add tests once above is done
```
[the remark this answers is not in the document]
andrewsg: Done.

andrewsg:
@crwilcox @frankyn PTAL. Changes include:

- Retry predicate simplified; my testing suggests the google-api-core nuanced inspection of error messages is not needed for GCS.
- Retry predicate now includes all HTTP error codes listed in retry design, except 408
-- 408 is mysteriously missing from google-api-core exceptions and I have no way to test what happens if we actually get one (testbench was no help here)
- Retry predicate now includes ConnectionError which encompasses failure to find host and broken connections of all types. LMK if we need to narrow it down; we have fairly rich and detailed info on the specific ConnectionError subtype which we can use here.

Responded to all other comments as well.

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,105 @@
...
+import json
+
+
+_RETRYABLE_TYPES = (
+    exceptions.TooManyRequests,  # 429
+    exceptions.InternalServerError,  # 500
+    exceptions.BadGateway,  # 502
+    exceptions.ServiceUnavailable,  # 503
+    exceptions.GatewayTimeout,  # 504
+    requests.ConnectionError,
```
frankyn, line 29: @tritone this is a change in what we discussed around retrying ConnectionErrors in general. Discussed this with @andrewsg w.r.t retrying unable to connect errors and for now implementation is moving forward with this solution until it's baked into the api_core libraries. Rationale, there's a default timeout enabled that would timeout the retry compared to Golang which does not.

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,105 @@
...
+    """Predicate for determining when to retry."""
+    if isinstance(exc, _RETRYABLE_TYPES):
+        return True
+    elif isinstance(exc, exceptions.GoogleAPICallError):
+        return exc.code in _ADDITIONAL_RETRYABLE_STATUS_CODES
+    else:
+        return False
+
+
+DEFAULT_RETRY = retry.Retry(predicate=_should_retry)
```
frankyn, line 46: This is readable from the design, but could you document that this is retrying based on retryable error codes above?

frankyn, line 46: One additional comment, I know DEFAULT_RETRY was picked up from BigQuery but the word "default" doesn't mean much. It might help just to document that it retries retriable error codes.

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,105 @@
...
+"""The default retry object.
+
+To modify the default retry behavior, call a ``with_XXX`` method
+on ``DEFAULT_RETRY``. For example, to change the deadline to 30 seconds,
+pass ``retry=DEFAULT_RETRY.with_deadline(30)``. See google-api-core reference
+(https://googleapis.dev/python/google-api-core/latest/retry.html) for details.
+"""
+
+
+class ConditionalRetryPolicy(object):
```
frankyn, line 56: Please include documentation for using this ConditionalRetryPolicy helper class for future us.

frankyn (changes requested):
Generally LGTM, clean design with minimal impact on the code base!

Have documentation comment requests

google/cloud/storage/retry.py
tritone: Got it, this seems fine to me re: the fact that there is a default timeout in python. So the idea is that down the road, we can add an error in api_core which catches "connection reset by peer" more specifically?

google/cloud/storage/retry.py
```diff
@@ -0,0 +1,105 @@
...
+        return False
+
+
+DEFAULT_RETRY = retry.Retry(predicate=_should_retry)
+"""The default retry object.
+
+To modify the default retry behavior, call a ``with_XXX`` method
+on ``DEFAULT_RETRY``. For example, to change the deadline to 30 seconds,
+pass ``retry=DEFAULT_RETRY.with_deadline(30)``. See google-api-core reference
+(https://googleapis.dev/python/google-api-core/latest/retry.html) for details.
```
tritone, line 56: Would it be worth having a short sample snippet here rather than explaining and inlining? Might make it easier to understand what's necessary.

andrewsg:
@frankyn Added documentation as requested, PTAL

tritone (commented):
Couple small comments, generally looks really good!

google/cloud/storage/retry.py
frankyn: that's correct, because right now requests.ConnectionError is tied to a specific implementation, e.g. requests and should be wrapped by api_core / cloud_core for it to be used generally outside of just the GCS library.

google/cloud/storage/retry.py
andrewsg: I decided against this since a useful snippet here would have to show how to actually inject the modified retry via user code, and that feature is out of scope for this PR (will require changing the signature of every method in the library).

google/cloud/storage/retry.py
andrewsg: I'm not totally convinced that retrying for a little while by default if there is an interruption due to network outage is not the right thing to do anyways, but yes, the idea is that instead of tapping into requests here we would have a custom api_core exception like we do with other errors.

frankyn (approved):

NoahDragon:
Just wonder if the blob.update() also has the retry logic with this feature. Is there any way to specify the retry number?

Issue #292, opened 2020-10-13
Add test pass, classifiers for Python 3.9


crwilcox:
Python 3.9 classifiers and tests should be added.

https://github.com/googleapis/python-storage/blob/master/setup.py#L84
https://github.com/googleapis/python-storage/blob/master/setup.py#L93
https://github.com/googleapis/python-storage/blob/master/noxfile.py#L30

crwilcox:
Similarly, 3.5 should be dropped. it may make sense to drop 2.7 as well.

PR #295, opened 2020-10-15
feat: add classifiers for python3.9 and remove for python3.5

HemangChothani:
Towards #292 

- nox.py file is generated by synthtool, so didn't include changes here.


andrewsg:
Will this refuse to install on Python 2.7 or will it just no longer be recommended?

tseaver:
@andrewsg The following will refuse to install on 2.7, at least with any recent setuptools:
```python
    python_requires=">=3.6",
```

This *definitely* needs to get merged as a breaking change.

crwilcox:
I'd recommend waiting to remove Python 2 until other breaking changes. At this time supporting Python 2 isn't providing any drag, but I think once it is we can remove it. @andrewsg @frankyn do we know of any changes that are also breaking to bundle?

noxfile.py
```diff
@@ -27,8 +27,8 @@
 BLACK_PATHS = ["docs", "google", "tests", "noxfile.py", "setup.py"]
 
 DEFAULT_PYTHON_VERSION = "3.8"
-SYSTEM_TEST_PYTHON_VERSIONS = ["2.7", "3.8"]
-UNIT_TEST_PYTHON_VERSIONS = ["2.7", "3.5", "3.6", "3.7", "3.8"]
+SYSTEM_TEST_PYTHON_VERSIONS = ["3.8"]
+UNIT_TEST_PYTHON_VERSIONS = ["3.6", "3.7", "3.8", "3.9"]
```
tseaver, line 31: Your comment says that you weren't including these changes, which should be driven by synthtool.

setup.py
```diff
@@ -75,13 +75,11 @@
         "Intended Audience :: Developers",
         "License :: OSI Approved :: Apache Software License",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 2.7",
```
tseaver, line 79: Dropping support for Python 2.7 is definitely a breaking change.  ISTM that #313 (where `synthtool` adds 3.9 support to `noxfile.py`, without dropping 3.7) suggests that we aren't yet ready to do that.  @crwilcox WDYT?

setup.py
```diff
@@ -90,7 +88,7 @@
     namespace_packages=namespaces,
     install_requires=dependencies,
     extras_require=extras,
-    python_requires=">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*",
+    python_requires=">=3.6",
```
tseaver, line 91: Given we decide dropping 2.7 is premature:
```suggestion
    python_requires=">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*",
```

noxfile.py
tseaver: #313 indeed does update `noxfile.py` via synth.

tseaver: Fixed when resolving conflict with #313.

setup.py
tseaver: I see that @crwilcox actually [commented already](https://github.com/googleapis/python-storage/pull/295#issuecomment-725695994) that we shouldn't drop 2.7 support.

tseaver (approved):

PR #298, opened 2020-10-19
fix: preserve metadata value

### LINKED ISSUE #293: Blob.metadata not preserved when uploading new file content
`blob.metadata` is not preserved when uploading new file content on an existing Blob instance with `blob.upload_from_string(...)` unless you first reset it with `blob.metadata = blob.metadata`.

I suspect this bug / surprising behaviour is the result of the somewhat awkward mapping of the REST API onto mutable objects. It feels like setting metadata should really be a parameter of the blob upload methods instead of a mutable field on the instance.

#### Environment details

  - OS type and version:

Both locally on my Mac and within Google Cloud Functions

  - Python version: `python --version`

Python 3.8.5

  - pip version: `pip --version`

pip 20.1.1

  - `google-cloud-storage` version: `pip show google-cloud-storage`

google-cloud-storage==1.31.2

#### Steps to reproduce

  See code example.

#### Code example

```python
>>> b = bucket.blob('testme1')
>>> b.metadata = {'test': 'me'}
>>> b.upload_from_string('foo')
>>> b.metadata
{'test': 'me'}
>>> b.upload_from_string('bar')
>>> b.metadata
>>> # Note how b.metadata is not blank

>>> b = bucket.blob('testme2')
>>> b.metadata = {'test': 'me'}
>>> b.upload_from_string('foo')
>>> b.metadata
{'test': 'me'}
>>> b.metadata = b.metadata # <-- workaround
>>> b.upload_from_string('bar')
>>> b.metadata
{'test': 'me'}
>>> # Note how b.metadata is now preserved
```

HemangChothani:
Fixes #293

### DIFF
--- a/tests/unit/test_blob.py
+++ b/tests/unit/test_blob.py
@@ -1828,12 +1828,17 @@ class Test_Blob(unittest.TestCase):
         if_metageneration_not_match=None,
         kms_key_name=None,
         timeout=None,
+        metadata=None,
     ):
         from six.moves.urllib.parse import urlencode
 
         bucket = _Bucket(name="w00t", user_project=user_project)
         blob = self._make_one(u"blob-name", bucket=bucket, kms_key_name=kms_key_name)
         self.assertIsNone(blob.chunk_size)
+        if metadata:
+            self.assertIsNone(blob.metadata)
+            blob._properties["metadata"] = metadata
+            self.assertEqual(len(blob._changes), 0)
 
         # Create mocks to be checked for doing transport.
         transport = self._mock_transport(http_client.OK, {})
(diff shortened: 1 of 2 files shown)


frankyn:
@andrewsg PTAL when you have a moment.

tests/unit/test_blob.py
```diff
@@ -1906,10 +1911,14 @@ def _do_multipart_success(
 
         upload_url += "?" + urlencode(qs_params)
 
+        blob_data = b'{"name": "blob-name"}\r\n'
+        if metadata:
+            blob_data = b'{"name": "blob-name", "metadata": {"test": "test"}}\r\n'
```
tseaver, line 1916: This needs to use the passed-in `metadata` argument, e.g.:
```suggestion
            blob_data = b'{"name": "blob-name", "metadata": %r}\r\n' % (metadata,)
```

tests/unit/test_blob.py
```diff
@@ -1906,10 +1911,14 @@ def _do_multipart_success(
...
+            self.assertEqual(blob._changes, set(["metadata"]))
         payload = (
             b"--==0==\r\n"
             + b"content-type: application/json; charset=UTF-8\r\n\r\n"
-            + b'{"name": "blob-name"}\r\n'
+            + blob_data
             + b"--==0==\r\n"
             + b"content-type: application/xml\r\n\r\n"
             + data_read
```
tseaver, line 1924: This test needs to assert that `metadata` has been pushed to the blob's changes, too.

tseaver (changes requested):

tests/unit/test_blob.py
HemangChothani: I think assert for blob._changes is already there.
https://github.com/googleapis/python-storage/blob/cfefaaa3ef2b5e66bb9b60e9917095de39e40bce/tests/unit/test_blob.py#L1917

HemangChothani:
@tseaver PTAL!

tseaver (approved):

Issue #294, opened 2020-10-13
There should be a built-in way to list prefixes (directories) in a bucket


lrowe:
The REST API provides a way to list prefixes (directories) in a bucket but this does not seem to be supported by the client library.

According to #192 the iterator returned by `list_blobs` has a `prefixes` field which is filled in as you iterate over the blob. This should be better documented, since it is only mentioned in passing:

> include_trailing_delimiter (boolean) – (Optional) If true, objects that end in exactly one instance of delimiter will have their metadata included in items in addition to prefixes.

https://googleapis.dev/python/storage/latest/client.html#google.cloud.storage.client.Client.list_blobs

As well as a method to list prefixes alone, it would be helpful to have an iterator that returned both prefixes and objects in order for produce ordered listings.

Workaround found on Stack Overflow. https://stackoverflow.com/a/59008580

```python
def list_prefixes(client, bucket_name, prefix, delimiter):
    # Adapted from https://stackoverflow.com/a/59008580
    from google.api_core import page_iterator
    return page_iterator.HTTPIterator(
        client=client,
        api_request=client._connection.api_request,
        path=f"/b/{bucket_name}/o",
        items_key="prefixes",
        item_to_value=lambda iterator, item: item,
        extra_params={
            "projection": "noAcl",
            "prefix": prefix,
            "delimiter": delimiter,
        },
    )
```

frankyn:
Hi @lrowe,
Thanks for filing the issue.

Would it help to document an example instead pydocs? Here's an example:

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/storage/cloud-client/storage_list_files_with_prefix.py#L49-L65

lrowe:
@frankyn An example would be helpful too, but I think the `prefixes` property should be included the API docs since they are expected to be exhaustive.

Perhaps the `:returns:` section of `list_blobs` could be expanded to mention the special prefixes property of the iterator, see [list_blobs docstring](https://github.com/googleapis/python-storage/blob/master/google/cloud/storage/bucket.py#L1208-L1210):
```
        :rtype: :class:`~google.api_core.page_iterator.Iterator`
        :returns: Iterator of all :class:`~google.cloud.storage.blob.Blob`
                  in this bucket matching the arguments.
```

lrowe:
I do worry though that the current list_blobs Iterator.prefixes set might be problematic when there are a very large number of prefixes since it would grow unbounded.

frankyn:
[...]

PR #300, opened 2020-10-21
feat: add list prefix method

HemangChothani:
Fixes https://github.com/googleapis/python-storage/issues/294

### DIFF
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -1296,7 +1296,7 @@ class Bucket(_PropertyMixin):
         client=None,
         timeout=_DEFAULT_TIMEOUT,
     ):
-        """Return an iterator used to find directories(prefixes) in the bucket.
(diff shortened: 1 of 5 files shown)


google/cloud/storage/bucket.py
```diff
@@ -1263,6 +1285,102 @@ def list_blobs(
         iterator.prefixes = set()
         return iterator
 
+    def list_prefixes(
```
frankyn, line 1288: Pending feedback from issue filer. I'm not convinced this is the right path forward.

lrowe:
Looking a little more deeply at the page_iterator and Storage [Objects: list](https://cloud.google.com/storage/docs/json_api/v1/objects/list) JSON API it seems both items and prefixes are always returned in the same response, so adding `list_prefixes` may encourage API consumers to make twice as many Storage Objects list requests as necessary in cases where they want both prefixes and objects.

Perhaps the more general solution would be to create an listing method that returns an iterator of pages, with each page having both an iterator of prefixes and an iterator or blobs.

PR #310, opened 2020-11-11
tests: avoid depending on hash order

### LINKED ISSUE #309: New test for 'Blob._do_multipart_upload' w/ metadata flakes
The [flaky test](https://source.cloud.google.com/results/invocations/26bf2df0-2ddb-4699-bc41-2430500326e9/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-storage%2Fpresubmit%2Fpresubmit/log) was [introduced in #298](https://github.com/googleapis/python-storage/pull/298/files#diff-e4a098aa380a5eea289442e5395c3af9905e5208d2e89ed56e1bea1f8ee3ebeaR1916)

The issue is that the code which constructs the expected payload for the request is flaky given Python's [hash randomization](https://docs.python.org/3/reference/datamodel.html#object.__hash__):

```python
        blob_data = b'{"name": "blob-name"}\r\n'
        if metadata:
            blob_data = (
                b'{"name": "blob-name", "metadata": '
                + json.dumps(metadata).encode("utf-8")
                + b"}\r\n"
            )
            self.assertEqual(blob._changes, set(["metadata"]))
        payload = (
            b"--==0==\r\n"
            + b"content-type: application/json; charset=UTF-8\r\n\r\n"
            + b'{"name": "blob-name"}\r\n'
            + blob_data
            + b"--==0==\r\n"
            + b"content-type: application/xml\r\n\r\n"
            + data_read
```

tseaver:
Closes #309

### DIFF
--- a/tests/unit/test_blob.py
+++ b/tests/unit/test_blob.py
@@ -1932,19 +1932,15 @@ class Test_Blob(unittest.TestCase):
 
         upload_url += "?" + urlencode(qs_params)
 
-        blob_data = b'{"name": "blob-name"}\r\n'
+        blob_data = {"name": "blob-name"}
         if metadata:
-            blob_data = (
-                b'{"name": "blob-name", "metadata": '
-                + json.dumps(metadata).encode("utf-8")
-                + b"}\r\n"
-            )
+            blob_data["metadata"] = metadata
             self.assertEqual(blob._changes, set(["metadata"]))
         payload = (
             b"--==0==\r\n"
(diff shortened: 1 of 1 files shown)


tswast (approved):
LGTM, though I'm curious if this would still be flakey on versions of Python where key names are ordered by time of entry.

tseaver:
@tswast The only flake I saw was on Python 3.5.

PR #311, opened 2020-11-12
refactor: move methods to client class

### LINKED ISSUE #38: Proposal: move API-methods to client

HemangChothani:
Fixes #38


andrewsg (approved):
LGTM

andrewsg:
@tswast In the bug you said, "we need to update samples for moved methods before deprecating the old ones." Is this change okay as-is, or do we need to make this change without the deprecation warning first, then update samples, then add the deprecation warning?

tswast:
If it helps you find the samples that need to be updated, I'm okay merging this PR first. It's just a bad look if we have samples on cloud.google.com that show deprecation warnings.

andrewsg:
I agree. But we can't update the sample without pushing part of this change, so @HemangChothani Let's omit the deprecation warning for now and open an issue to update the samples, then add the deprecation warning in that order.

HemangChothani:
@andrewsg Opend an issue https://github.com/GoogleCloudPlatform/python-docs-samples/issues/5031 to update the samples.

HemangChothani:
@andrewsg  I have opend an issue to update sample  https://github.com/GoogleCloudPlatform/python-docs-samples/issues/5031 in python-docs-sample repo , but i didn't find any sample needs to update, should i open issue in `python-storage` repo to update snippets and system-tests for `list_blobs` and `download_to_file` methods

frankyn:
@andrewsg and @HemangChothani, I replied to the issue. We are only missing updates for  download_to_filename.

HemangChothani:
@frankyn we have moved ` download_to_file` into the client class, ` download_to_filename` is still in the blob class so i think there is no need to update any sample.

 ` blob.download_to_filename`  is calling `blob.download_to_file` method which is deprecated now, so  i will update soon and direct call the method `client.download_to_file` method.

Also update the method call from `bucket.list_blobs` to `client.list_blobs`.

frankyn:
Why was only download_to_file updated but not download_to_filename?

HemangChothani:
Document mentioned in issue https://github.com/googleapis/python-storage/issues/38, found `download_to_file` and `list_blobs` methods need to move implementation from blob class to client class and deprecate older one.

PR #308, opened 2020-11-11
tests: add explicit coverage for retry members

tseaver:
See also: #283 

This PR adds explicit coverage for each member of `google.cloud.storage.retry`.  Together with PR #284, should return this package to 100% unit test coverage.

PR #316, opened 2020-11-16
fix: retry uploads only conditionally

andrewsg:


google/cloud/storage/blob.py
```diff
@@ -2058,6 +2060,15 @@ def _do_upload(
                   **only** response in the multipart case and it will be the
                   **final** response in the resumable case.
         """
+        if if_metageneration_match is None and num_retries is None:
```
tritone, line 2063: I think we should update the docstring as well to make it clear that setting num_retries can be used to override the idempotency requirement (assuming that's the intention), to make sure that users know the implications.

tests/unit/test_blob.py
```diff
@@ -2577,6 +2577,12 @@ def _do_upload_helper(
             if_metageneration_not_match,
             **timeout_kwarg
         )
+
+        # Adjust num_retries expectations to reflect the conditional default in
+        # _do_upload()
+        if num_retries is None and if_metageneration_match is None:
```
tritone, line 2583: Can you comment on to what extent the conditional is covered by test cases?

tseaver: @tritone the `cover` session under `nox ` requires (since the merge of #308) 100% coverage (branch and statements).

tritone: Ah yeah that's right, good point.

andrewsg: Branch coverage of the code I've added is 100% according to the coverage tool. We are exercising scenarios where if_metageneration_match = true and num_retries is not None.

google/cloud/storage/blob.py
andrewsg: Yes, that's the intention. Will do.

tseaver (approved):

PR #321, opened 2020-11-22
fix custom_time setter bug

tamir-jether:
the current setter doesn't update the _changes field. using _patch_property solves that.


tseaver (commented):
@tamir-jether Thanks very much for the patch!  I happened to review #323 first, and so will end up merging it in place of this one.

PR #323, opened 2020-11-22
fix: add change in 'Blob.custom_time' to changes

### LINKED ISSUE #322: 'Blob.custom_time' setter doesnt update '_changes'
i think the `custom_time setter` code should be:

```python
        self._patch_property('customTime', value)
```

instead of 

```python
        self._properties["customTime"] = value
```

haim0n:
This way '_PropertyMixin.patch' will send them to server.

Fixes #322 🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -3613,7 +3613,7 @@ class Blob(_PropertyMixin):
         if value is not None:
             value = _datetime_to_rfc3339(value)
 
-        self._properties["customTime"] = value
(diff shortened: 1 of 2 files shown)


tseaver (changes requested):
Thanks for the patch.  Please add unit test assertions (as I did in PR #325, which I am closing in favor of this one).

haim0n:
> Thanks for the patch. Please add unit test assertions (as I did in PR #325, which I am closing in favor of this one).

done

tseaver (approved):

PR #325, opened 2020-11-24
fix: update 'custom_time' setter to record change

tseaver:
Also, add test assertions for 'metadata' setter for the same issue.

Closes #323


tseaver:
PR #323 already has the fix.  I have asked the submitter to add the assertions from here.

PR #326, opened 2020-11-24
fix: fall back to 'charset' of 'content_type' in 'download_as_text' 

### LINKED ISSUE #319: Blob.download_as_text does not decode properly
`Blob.download_as_text` tries to use the `content-encoding` header to decode the bytes. In most cases that value is `gzip`, even though the bytes were already decompressed at that point. In other cases, e.g `text/plain; charset=utf-8`, the value does not make sense to Python's `bytes.decode`.

```
  File "/storage/bazel-cache/_bazel_kornholi/9f066b43468ef9bfd3c6a621a4515622/execroot/__main__/bazel-out/k8-opt/bin/foo.runfiles/pypi__google_cloud_storage_1_33_0/google/cloud/storage/blob.py", line 1424, in download_as_text
    return data.decode(self.content_encoding)
LookupError: unknown encoding: gzip
```

I don't think we can be smarter here than passing through the `encoding` kwarg which defaults to utf-8.

tseaver:
Explicit 'encoding' overrides the fallback.

Use the 'charset' param of 'content_type', rather than 'content_encoding',
which isn't going to be a Unicode -> bytes encoding.

Closes #319.

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -152,8 +152,8 @@ class Blob(_PropertyMixin):
         contents.
 
     :type generation: long
-    :param generation: (Optional) If present, selects a specific revision of
-                       this object.
+    :param generation:
+        (Optional) If present, selects a specific revision of this object.
     """
 
     _chunk_size = None  # Default value for each instance.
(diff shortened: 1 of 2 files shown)


tseaver:
Note for reviewers:  the first commit (a006df3) is the meat of the PR.  The second one (3685129) is just rewrapping long / overindented docstring params for readability within the source.

tswast (approved):
LGTM for Python readability. Thanks for adding clear tests for this case (no explicit encode, but with a charset in the content-type header)

PR #331, opened 2020-12-01
feat: make retry parameter public and added in other methods

### LINKED ISSUE #315: Add a default retry parameter to all public methods
All public storage method should have a default retry parametr defined so that clients don't have to explicitly use the `@retry.Retry` decorator to handle transient issues, like 503 error code.

The solution could look similar to the implementation in the [google-cloud-bigquery](https://github.com/googleapis/python-bigquery/blob/master/google/cloud/bigquery/client.py#L830) where retry is implemented in every public method and it is set to DEFAULT_RETRY by default.

HemangChothani:
Fixes #315

### DIFF
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -145,6 +145,7 @@ class _PropertyMixin(object):
         if_generation_not_match=None,
         if_metageneration_match=None,
         if_metageneration_not_match=None,
+        retry=DEFAULT_RETRY,
     ):
         """Reload properties from Cloud Storage.
 
@@ -187,6 +188,9 @@ class _PropertyMixin(object):
         :type if_metageneration_not_match: long
         :param if_metageneration_not_match: (Optional) Make the operation conditional on whether the
                                             blob's current metageneration does not match the given value.
+
+        :type retry: google.api_core.retry.Retry
+        :param retry: (Optional) How to retry the RPC.
         """
         client = self._require_client(client)
         query_params = self._query_params
@@ -207,7 +211,7 @@ class _PropertyMixin(object):
             headers=self._encryption_headers(),
             _target_object=self,
             timeout=timeout,
-            retry=DEFAULT_RETRY,
+            retry=retry,
         )
         self._set_properties(api_response)
 
@@ -247,6 +251,7 @@ class _PropertyMixin(object):
         if_generation_not_match=None,
         if_metageneration_match=None,
         if_metageneration_not_match=None,
+        retry=DEFAULT_RETRY_IF_METAGENERATION_SPECIFIED,
     ):
         """Sends all changed properties in a PATCH request.
 
@@ -286,6 +291,9 @@ class _PropertyMixin(object):
         :type if_metageneration_not_match: long
         :param if_metageneration_not_match: (Optional) Make the operation conditional on whether the
                                             blob's current metageneration does not match the given value.
+
+        :type retry: google.api_core.retry.Retry
+        :param retry: (Optional) How to retry the RPC.
         """
         client = self._require_client(client)
         query_params = self._query_params
@@ -309,7 +317,7 @@ class _PropertyMixin(object):
             query_params=query_params,
             _target_object=self,
             timeout=timeout,
-            retry=DEFAULT_RETRY_IF_METAGENERATION_SPECIFIED,
+            retry=retry,
         )
         self._set_properties(api_response)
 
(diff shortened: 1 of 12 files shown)


andrewsg:
Thanks! This is a major change to the UI in line with our medium-term plan for retry support, so I (author of underlying retry logic) will review thoroughly soon.

google/cloud/storage/_helpers.py
```diff
@@ -187,6 +188,9 @@ def reload(
         :type if_metageneration_not_match: long
         :param if_metageneration_not_match: (Optional) Make the operation conditional on whether the
                                             blob's current metageneration does not match the given value.
+
+        :type retry: google.api_core.retry.Retry
+        :param retry: (Optional) How to retry the RPC.
```
andrewsg, line 193: In order for users to use this, they'll need an explanation of what kinds of objects to pass in and how to modify them. "See retry.py" might be enough if retry.py's comments and docstrings are comprehensive.

google/cloud/storage/acl.py
```diff
@@ -430,7 +431,7 @@ def _require_client(self, client):
             client = self.client
         return client
 
-    def reload(self, client=None, timeout=_DEFAULT_TIMEOUT):
+    def reload(self, client=None, timeout=_DEFAULT_TIMEOUT, retry=DEFAULT_RETRY):
```
andrewsg, line 434: Here in the ACL file you're not just exposing retries in the function signature but actually adding a new default, right? @frankyn ACL commands aren't in the consolidated retry strategy doc, do we want retries enabled on them?

andrewsg (changes requested):

google/cloud/storage/_helpers.py
```diff
@@ -190,7 +190,9 @@ def reload(
                                             blob's current metageneration does not match the given value.
 
         :type retry: google.api_core.retry.Retry
```
andrewsg, line 192: Technically either google.api_core.retry.Retry or google.cloud.storage.retry.ConditionalRetryPolicy are supported here, and in some cases ConditionalRetryPolicy is the default so despite the complexity we should document it here.

google/cloud/storage/_helpers.py
```diff
@@ -190,7 +190,9 @@ def reload(
...
-        :param retry: (Optional) How to retry the RPC.
+        :param retry: (Optional) How to retry the RPC. To modify the default retry behavior,
+            create a new retry object modeled after this one by calling it a ``with_XXX`` method.
+            See: https://googleapis.dev/python/google-api-core/latest/retry.html for details.
```
andrewsg, line 195: (Optional) How to retry the RPC. A `None` value will disable retries. A `google.api_core.retry.Retry` value will enable retries, and the object will define retriable response codes and errors and configure backoff and timeout options.

A `google.cloud.storage.retry.ConditionalRetryPolicy` value wraps a Retry object and activates it only if certain conditions are met. This class exists to provide safe defaults for RPC calls that are not technically safe to retry normally (due to potential data duplication or other side-effects) but become safe to retry if a condition such as `if_metageneration_match` is set.

See the google.cloud.storage.retry source code and docstrings for information on retry types and how to configure them.

andrewsg (changes requested):

google/cloud/storage/acl.py
```diff
@@ -456,13 +471,19 @@ def reload(self, client=None, timeout=_DEFAULT_TIMEOUT):
...
+            query_params=query_params,
+            timeout=timeout,
+            retry=retry,
         )
         self.loaded = True
         for entry in found.get("items", ()):
             self.add_entity(self.entity_from_dict(entry))
 
-    def _save(self, acl, predefined, client, timeout=_DEFAULT_TIMEOUT):
+    def _save(
```
andrewsg, line 484: We've investigated and found that ACL save retries don't support the concurrency control we'd need to have full faith in retries here. Please go ahead and remove them. Thanks!

andrewsg (changes requested):

google/cloud/storage/acl.py
frankyn: ACLs don't have optimistic concurrency control, therefore we would need to update implementation to rely on storage.objects.patch and storage.buckets.patch instead of ACL specific APIs. 

Let's remove ACL retries from this PR for now and open a tracking issue.

andrewsg (approved):
LGTM. Thanks for your efforts here!

andrewsg:
@HemangChothani are we clear to merge?

HemangChothani:
@andrewsg Yes.

PR #333, opened 2020-12-01
fix: avoid triggering global logging config

### LINKED ISSUE #332: logging.info() triggers basicConfig()
When I use `download_as_string()` method or similar, `logging.root` logger is being activated by calling `logging.info()` on line 
https://github.com/googleapis/python-storage/blob/63ff23387f5873c609490be8e58d69ba34a10a5e/google/cloud/storage/blob.py#L926. It causes other loggers to be logged by the root logger which is unexpected as some apps require a single logger in strict format.

#### Environment details

  - OS type and version: ubuntu 20.04
  - Python version: `3.8.6`
  - pip version: `20.2.4`
  - `google-cloud-storage` version: `1.33.0`

#### Steps to reproduce

  1. set up your own logger for any logger
  2. call `logger.info("test")` (nothing should happen)
  3. run Client.from_service_account_json().get_bucket(...).blob(..., chunk_size=1024*1024).download_as_string()
  4. call `logger.info("test")` one more time
  5. log in basic format should appear on stdout (unexpected)

### How to fix it ###
you need to use your own logger (e.g. using `logging.getLogger(__name__)` instead of root logger.

tseaver:
Closes #332

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -122,6 +122,8 @@ _CHUNKED_DOWNLOAD_CHECKSUM_MESSAGE = (
 _DEFAULT_CHUNKSIZE = 104857600  # 1024 * 1024 B * 100 = 100 MB
 _MAX_MULTIPART_SIZE = 8388608  # 8 MB
 
+_logger = logging.getLogger(__name__)
+
 
 class Blob(_PropertyMixin):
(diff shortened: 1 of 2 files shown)


andrewsg (approved):

andrewsg:
Thanks!

PR #337, opened 2020-12-04
chore: add constraints file

busunkim96:
Add constraints file to test lower bounds.

These files will not be used until the noxfile is changed in https://github.com/googleapis/synthtool/pull/869.


andrewsg:
Why the blank files for other py versions?

tseaver (dismissed):

tswast:
@andrewsg Re: "Why the blank files for other py versions?"

I added these in BigQuery because then we can always use a constraints file for every install step in the noxfile. An empty constraints file is a no-op.

These can also be customized to support other import versions besides the earliest, such as the latest/earliest version of a dependency package to support Python 3.7, for example. I had to do this in pandas-gbq to avoid installing numpy from source.

tseaver (approved):

PR #340, opened 2020-12-10
fix: fix conditional retry handling of camelCase query params

andrewsg:
A customer reported difficulty with if_metageneration_match conditional retries. It turned out to be confusion in the retry code over snake case vs. camel case query params; this will resolve that issue.

### DIFF
--- a/google/cloud/storage/retry.py
+++ b/google/cloud/storage/retry.py
@@ -99,13 +99,13 @@ class ConditionalRetryPolicy(object):
 def is_generation_specified(query_params):
     """Return True if generation or if_generation_match is specified."""
     generation = query_params.get("generation") is not None
-    if_generation_match = query_params.get("if_generation_match") is not None
+    if_generation_match = query_params.get("ifGenerationMatch") is not None
     return generation or if_generation_match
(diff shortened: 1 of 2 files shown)


frankyn (approved):

tritone (commented):
Belated comment on this-- does this seem to indicate that we're missing an integration test case?

Also @frankyn are these parameter keys case-sensitive?

andrewsg:
@tritone Yes, I thought the same thing. We discussed integration coverage for this area in the initial commit, but found that there was really no good way to test the retry logic without mocking server responses in our systems tests. Rather than do that, which would be a change in policy on mocks in system tests, we decided to wait for automated conformance testing using the test bench which is supposedly coming in the future.

I'm open to revisiting that decision; this bug was nasty to track down and would have been easy with good integration tests here. What do you think?

If I recall correctly (I haven't tested this on GCS, this is my experience from other libraries) the API is flexible on how it accepts parameters, and both camelCase and snake_case are accepted, hence my confusion here. I don't know whether case is enforced on camel case; I'd assume it is not.

PR #356, opened 2021-01-04
fix: allow signed url version v4 without signed credentials

### LINKED ISSUE #355: Impossible to use signed url v4 and the Service Account Credential APIs
Months ago, I solved an issue to a Stack overflow user by explaining [how to generate a signed URL without a service account JSON fey file](https://stackoverflow.com/questions/64234214/how-to-generate-a-blob-signed-url-in-google-cloud-run/64245028#64245028). I also [release an article on this](https://medium.com/google-cloud/service-account-credentials-api-a-solution-to-different-issues-dc7434037115)

Today, a comment of a user told me that is doesn't work. They open a new [Stack Overflow question](https://stackoverflow.com/questions/65568916/generating-cloud-storage-signed-url-from-google-cloud-function-without-using-exp/65569603#65569603) and in their question, there is the v4 mentioned.

I dug into the code and yes, the v2 and the v4 haven't the same behavior: the v4 check if the credential can [generate a signed URL by itself](https://github.com/googleapis/python-storage/blob/9d5bb5622f17a0143fa6670406596e53d0068286/google/cloud/storage/_signing.py#L536), that implies the [useless check later](https://github.com/googleapis/python-storage/blob/9d5bb5622f17a0143fa6670406596e53d0068286/google/cloud/storage/_signing.py#L623) that allows to use Service Account Credential API, instead of the service account private key to generate the signature.

#### Environment details

  - Linux (Cloud Function and Cloud Run)
  - Python version: `python --version` 3.7 and 3.8
  - `google-cloud-storage` version: latest (1.35.0)

#### Steps to reproduce

  1. Working code on Cloud Run (for example) *Replace YOUR_BUCKET by a valid bucket*

```
import os
from flask import Flask, request

app = Flask(__name__)

@app.route('/', methods=['GET', 'POST'])
def sign_url():
    from datetime import datetime, timedelta

    import google.auth

    credentials, project_id = google.auth.default()

    # Perform a refresh request to get the access token of the current credentials (Else, it's None)
    from google.auth.transport import requests
    r = requests.Request()
    credentials.refresh(r)

    client = storage.Client()
    bucket = client.get_bucket('YOUR_BUCKET')
    blob = bucket.get_blob('name.csv')

    service_account_email = credentials.service_account_email

    url = blob.generate_signed_url(service_account_email=service_account_email, access_token=credentials.token,version="v2", method="PUT",expiration=timedelta(minutes=120),content_type="application/octet-stream")
    return url, 200

if __name__ == "__main__":
    app.run(host='0.0.0.0', port=int(os.environ.get('PORT', 8080)))

```

  2.  Broken code on Cloud Run (for example) *Replace YOUR_BUCKET by a valid bucket*

```
import os
from flask import Flask, request

app = Flask(__name__)

@app.route('/', methods=['GET', 'POST'])
def sign_url():
    from datetime import datetime, timedelta

    import google.auth

    credentials, project_id = google.auth.default()

    # Perform a refresh request to get the access token of the current credentials (Else, it's None)
    from google.auth.transport import requests
    r = requests.Request()
    credentials.refresh(r)

    client = storage.Client()
    bucket = client.get_bucket('YOUR_BUCKET')
    blob = bucket.get_blob('name.csv')

    service_account_email = credentials.service_account_email

    url = blob.generate_signed_url(service_account_email=service_account_email, access_token=credentials.token,version="v4", method="PUT",expiration=timedelta(minutes=120),content_type="application/octet-stream")
    return url, 200

if __name__ == "__main__":
    app.run(host='0.0.0.0', port=int(os.environ.get('PORT', 8080)))

```

*Only the version (v2/v4) has changed*

Fix proposal is coming.

guillaumeblaquiere:
Fixes #355 🦕

I used the service_account_email passed in method parameter as "signer" instead of taking it in the credentials.signer_email method. That allows to not use `google.auth.credentials.Signing` type as credential parameter and to use the Service Account credential API instead.

I preserved the checks in case of `access_token` or `service_account_email` set to nil, to preserve the efficiency and the existing behavior.

----

I also updated tests. Successful with python 3.8.

Open to comment and to improve this PR if needed


google/cloud/storage/_signing.py
```diff
@@ -542,7 +541,11 @@ def generate_signed_url_v4(
         request_timestamp = _request_timestamp
         datestamp = _request_timestamp[:8]
 
-    client_email = credentials.signer_email
+    client_email = service_account_email
+    if not access_token or not service_account_email:
+        ensure_signed_credentials(credentials)
+        client_email = credentials.signer_email
```
tseaver, line 547: Please update the docstring above to indicate that `credentials` must proviide `signer_email` only if `service_account_email` and `access_token` are not passed.

Also, this new branch (where `credentials` is ignored because `service_account_email` is passed) will require a separate unit test below.

tests/unit/test__signing.py
```diff
@@ -655,16 +655,16 @@ def test_w_custom_query_parameters_w_none_value(self):
 
     def test_with_access_token(self):
```
tseaver, line 656: We need to test both the case where `access_token` is passed explicitly, but `service_account_email` is not (falling back to `client.signer_email`), *and* we need to test where both values are passed explicitly.  We also need assertions that the correct values are passed through to the `_sign_message` mock.

tseaver (dismissed):

cajoek:
I really want this!

tests/unit/test__signing.py
guillaumeblaquiere: Done, working path tested, not the exception cases. I can do it if required

google/cloud/storage/_signing.py
guillaumeblaquiere: Docstring added, and new test cases added.

guillaumeblaquiere:
Sorry for the delay, the week was awful!

guillaumeblaquiere:
@tseaver any review possible on this?

frankyn (approved):
Thanks @guillaumeblaquiere LGTM

Issue #357, opened 2021-01-07
API Gateway timeout


nitishxp:
I recently updated my google-cloud-storage to **1.35.0** in my production system after that I am randomly getting this error
I am running my code on Cloud Functions

`gcs.get_bucket(bucket_name) File "/env/local/lib/python3.7/site-packages/google/cloud/storage/client.py", line 400, in get_bucket retry=retry, File "/env/local/lib/python3.7/site-packages/google/cloud/storage/bucket.py", line 1002, in reload retry=retry, File "/env/local/lib/python3.7/site-packages/google/cloud/storage/_helpers.py", line 225, in reload retry=retry, File "/env/local/lib/python3.7/site-packages/google/cloud/storage/_http.py", line 63, in api_request return call() File "/env/local/lib/python3.7/site-packages/google/cloud/_http.py", line 438, in api_request raise exceptions.from_http_response(response) google.api_core.exceptions.GatewayTimeout: 504 GET https://storage.googleapis.com/storage/v1/b/bucket_name?projection=noAcl&prettyPrint=false:cs.get_bucket(project + '-stage') File "/env/local/lib/python3.7/site-packages/google/cloud/storage/client.py", line 400, in get_bucket retry=retry, File "/env/local/lib/python3.7/site-packages/google/cloud/storage/bucket.py", line 1002, in reload retry=retry, File "/env/local/lib/python3.7/site-packages/google/cloud/storage/_helpers.py", line 225, in reload retry=retry, File "/env/local/lib/python3.7/site-packages/google/cloud/storage/_http.py", line 63, in api_request return call() File "/env/local/lib/python3.7/site-packages/google/cloud/_http.py", line 438, in api_request raise exceptions.from_http_response(response) google.api_core.exceptions.GatewayTimeout: 504 GET https://storage.googleapis.com/storage/v1/b/bucket_name?projection=noAcl&prettyPrint=false:`

andrewsg:
It looks like the refactor in 1.35.0 accidentally changed the default retry behavior for get_bucket. Thank you for your timely report. I'll fix this but in the meantime you can enable retries on get_bucket by calling it as:
```
from google.cloud.storage import retry
gcs.get_bucket(bucket_name, retry=retry.DEFAULT_RETRY)
```

nitishxp:
Thank you @andrewsg when I can expect the fix to be released?

nitishxp:
@andrewsg 

[...]

PR #358, opened 2021-01-07
fix: Amend default retry behavior for bucket operations on client

andrewsg:
Fixes: #357 

The bug exposed a hole in our testing of the new retry defaults, so this PR also adds some simple tests to verify defaults are correct. It doesn't try to comprehensively validate all defaults throughout the library, but the existence of this bug suggests we might want to audit those to make sure.

### DIFF
--- a/google/cloud/storage/client.py
+++ b/google/cloud/storage/client.py
@@ -52,7 +52,6 @@ from google.cloud.storage.acl import BucketACL
 from google.cloud.storage.acl import DefaultObjectACL
 from google.cloud.storage.constants import _DEFAULT_TIMEOUT
 from google.cloud.storage.retry import DEFAULT_RETRY
-from google.cloud.storage.retry import DEFAULT_RETRY_IF_METAGENERATION_SPECIFIED
 
 
 _marker = object()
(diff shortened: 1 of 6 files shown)


google/cloud/storage/client.py
```diff
@@ -52,7 +52,6 @@
 from google.cloud.storage.acl import DefaultObjectACL
 from google.cloud.storage.constants import _DEFAULT_TIMEOUT
 from google.cloud.storage.retry import DEFAULT_RETRY
-from google.cloud.storage.retry import DEFAULT_RETRY_IF_METAGENERATION_SPECIFIED
```
tseaver, line 55: If `DEFAULT_RETRY_IF_METAGENERATION_SPECIFIED` is only supposed to be used for "write" operations, then we need to update other places, too:  `Bucket.get_blob`, `Bucket.exists`, `Blob.exists` should all use `DEFAULT_RETRY`.

tests/unit/test_client.py
```diff
@@ -562,6 +562,29 @@ def test_get_bucket_with_object_hit(self):
...
+        PROJECT = "PROJECT"
+        CREDENTIALS = _make_credentials()
+        client = self._make_one(project=PROJECT, credentials=CREDENTIALS)
+
+        bucket_name = "bucket-name"
+        bucket_obj = Bucket(client, bucket_name)
+
+        with mock.patch.object(Connection, "api_request") as req:
+            client.get_bucket(bucket_obj)
+            req.assert_called_once_with(
```
tseaver, line 578: It is clearer to move the assertion back out of the `with` scope.

Also, ISTM we need to test that an explicitly-passed `retry` argument is honored.

andrewsg:
@tseaver PTAL

tritone (approved):

frankyn (approved):

PR #367, opened 2021-01-27
feat: add mtls support

arithmetic1728:
https://google.aip.dev/auth/4114
googlers see go/mtls-python-cloud-core-clients for more details.

Part of the mtls feature is implemented in https://github.com/googleapis/python-cloud-core/pull/75, and will be released as version 1.16.0.

This PR adds the mtls feature to storage client lib. Note that:
(1) if the python-cloud-core version is < 1.16.0, this PR does nothing, it is backward compatible and won't break any current users.
(2) if the user sets `GOOGLE_API_USE_CLIENT_CERTIFICATE` env var to "true" to trigger mtls, then the PR checks python-cloud-core version. It throws an exception asking the user to bump the version, if the version < 1.16.0. So probably it is a good idea to release python-cloud-core 1.16.0 before merging this PR.
(3) the unit tests work for both python-cloud-core versions, so unit test shouldn't break after the upgrading in the future.
(4) for mtls testing (running internally), we need to skip a couple of system tests:
 - kms/pubsub tests: because the version used doesn't support mtls
 - any tests using service account credentials: because mtls only works with user credentials


google/cloud/storage/blob.py
```diff
@@ -829,9 +829,12 @@ def _get_download_url(
         """
         name_value_pairs = []
         if self.media_link is None:
-            base_url = _DOWNLOAD_URL_TEMPLATE.format(
-                hostname=client._connection.API_BASE_URL, path=self.path
+            hostname = (
+                client._connection.API_BASE_URL
+                if not hasattr(client._connection, "get_api_base_url_for_mtls")
```
andrewsg, line 834: Is this something we can also simplify by requiring google-cloud-core 1.6.0 in setup.py? (also same idiom below)

tests/unit/test__http.py
```diff
@@ -213,3 +215,16 @@ def test_api_request_conditional_retry_failed(self):
...
+
+        conn = self._make_one(client, api_endpoint=None)
+        self.assertEqual(conn.ALLOW_AUTO_SWITCH_TO_MTLS_URL, True)
+        self.assertEqual(conn.API_BASE_URL, "https://storage.googleapis.com")
+        self.assertEqual(conn.API_BASE_MTLS_URL, "https://storage.mtls.googleapis.com")
+
+        conn = self._make_one(client, api_endpoint="http://foo")
+        self.assertEqual(conn.ALLOW_AUTO_SWITCH_TO_MTLS_URL, False)
+        self.assertEqual(conn.API_BASE_URL, "http://foo")
+        self.assertEqual(conn.API_BASE_MTLS_URL, "https://storage.mtls.googleapis.com")
```
frankyn, line 230: Can you test the call to `client._connection.get_api_base_url_for_mtls` without using a mock?

andrewsg (changes requested):

tests/system/test_system.py
```diff
@@ -2000,6 +2032,9 @@ def _kms_key_name(self, key_name=None):
     @classmethod
     def setUpClass(cls):
         super(TestKMSIntegration, cls).setUpClass()
+        if Config.TESTING_MTLS:
+            raise unittest.SkipTest("Skip kms tests for mTLS testing")
```
frankyn, line 2036: Are you skipping this test because KMS and similarly Pubsub are not yet ready?

tests/system/test_system.py
```diff
@@ -2453,6 +2488,15 @@ def test_ubla_set_unset_preserves_acls(self):
...
 
 class TestV4POSTPolicies(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestV4POSTPolicies, cls).setUpClass()
+        if (
+            type(Config.CLIENT._credentials)
+            is not google.oauth2.service_account.Credentials
+        ):
+            raise unittest.SkipTest("These tests require a service account credential")
```
frankyn, line 2498: Do these tests run successfully with a given service account with mTLS?

frankyn (changes requested):

google/cloud/storage/_http.py
```diff
@@ -15,12 +15,20 @@
...
+import pkg_resources
 
 from google.cloud import _http
 
 from google.cloud.storage import __version__
 
 
+if os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE") == "true":  # pragma: NO COVER
+    release = pkg_resources.get_distribution("google-cloud-core").parsed_version
+    if release < pkg_resources.parse_version("1.6.0"):
```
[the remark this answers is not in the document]
arithmetic1728: There is a requirement not to update google-cloud-core until the version sits long enough to be considered stable. We did similar thing for big query as well. https://github.com/googleapis/python-bigquery/pull/492

google/cloud/storage/_http.py
```diff
@@ -15,12 +15,20 @@
...
 import functools
+import os
+import pkg_resources
 
 from google.cloud import _http
 
 from google.cloud.storage import __version__
 
 
+if os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE") == "true":  # pragma: NO COVER
```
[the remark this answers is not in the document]
arithmetic1728: it must be exactly "true".

google/cloud/storage/blob.py
arithmetic1728: cannot upgrade it yet as stated above

tests/system/test_system.py
arithmetic1728: service account is not supported for mtls, only user credentials is supported

tests/unit/test__http.py
arithmetic1728: The reason is that `get_api_base_url_for_mtls` is only available for python-cloud-core >= 1.6.0 which we cannot upgrade it for now. We use mock here so the unit test won't break when we upgrade it in the future. Moreover, it is already well tested in python-cloud-core.

tests/system/test_system.py
arithmetic1728: The current kms and pubsub version used in system test are not the latest. New major version supports mtls (already tested and onboard to our testing framework), but it is for python 3 only, so we cannot upgrade it since it will break python 2.7 system test. Moreover, since the new version uses a new set of api, it will be a lot of work to upgrade the test code as well.

google/cloud/storage/_http.py
```diff
@@ -29,13 +37,19 @@ class Connection(_http.JSONConnection):
 
     :type client_info: :class:`~google.api_core.client_info.ClientInfo`
     :param client_info: (Optional) instance used to generate user agent.
+
+    :type api_endpoint: str or None
```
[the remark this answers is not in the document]
arithmetic1728: mTLS will be enabled if `GOOGLE_API_USE_CLIENT_CERITIFICATE` is set to "true". Yes, just added this sentence to the doc.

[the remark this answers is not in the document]
arithmetic1728: done

google/cloud/storage/blob.py
frankyn: Could you introduce a single helper internal helper method to reduce duplication of this check and clarify that it should be cleaned up once version 1.6.0 is stable?

tests/system/test_system.py
frankyn: Could you call this out as a comment in tests for posterity, this will be easily missed otherwise. Thank you for clarifying!

tests/system/test_system.py
frankyn: Could you add a comment that it's being skipped for this reason?

frankyn (changes requested):

tests/system/test_system.py
arithmetic1728: done

tests/system/test_system.py
arithmetic1728: done

google/cloud/storage/blob.py
arithmetic1728: done

frankyn (approved):

andrewsg (approved):

PR #374, opened 2021-02-08
fix: set custom_time on uploads

### LINKED ISSUE #372: Blob's custom_time is lost on upload.
#### Environment details

  - OS type and version: Debian Buster
  - Python version: Python 3.7.3
  - pip version: pip 18.1 from /usr/lib/python3/dist-packages/pip (python 3.7)
  - `google-cloud-storage` version: 1.35.1

#### Steps to reproduce

  1. Create a blob.
  2. Set `custom_time`.
  3. Upload the blob .
  4. Try to retrieve blob's `custom_time`.
  5. Expected result is a `datatime`, but the actual result is `None`.

If blob's `custom_time` is set before uploading, it is just lost. My understanding is that the metadata could be uploaded with the object's content at the same time without needing to patch it afterwards.

https://cloud.google.com/storage/docs/uploading-objects#rest-upload-objects

#### Code example

```python
import datetime

bucket_name = "my-test-bucket"
blob_name = "test_blob"

client = storage.Client()
bucket = client.get_bucket(bucket_name)

blob = bucket.blob(blob_name)
blob.custom_time = datetime.datetime.now(datetime.timezone.utc)
print("Before upload {}".format(blob.custom_time))
blob.upload_from_string("test")
blob.reload()
print("After upload {}".format(blob.custom_time))
blob.custom_time = datetime.datetime.now(datetime.timezone.utc)
blob.patch()
blob.reload()
print("After patching {}".format(blob.custom_time))
```

#### Stack trace
```
Before upload 2021-02-05 10:24:08.405494+00:00
After upload None
After patching 2021-02-05 10:24:08.814159+00:00
```

tritone:
The custom_time property was incorrectly not included as a field
that could be set on object upload.

Fixes #372

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -96,6 +96,7 @@ _WRITABLE_FIELDS = (
     "contentLanguage",
     _CONTENT_TYPE_FIELD,
     "crc32c",
+    "customTime",
     "md5Hash",
     "metadata",
     "name",
@@ -1530,6 +1531,7 @@ class Blob(_PropertyMixin):
         * ``contentLanguage``
         * ``contentType``
         * ``crc32c``
+        * ``customTime``
         * ``md5Hash``
         * ``metadata``
         * ``name``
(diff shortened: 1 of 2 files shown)


andrewsg (approved):
Thanks Chris!

PR #385, opened 2021-02-22
feat: add blob.open() for file-like I/O

### LINKED ISSUE #29: Add an API method to give us a streaming file object
It doesn't look like there's a way to get a streaming download from google storage in the Python API. We have `download_to_file` , `download_to_string`, and `download_to_filename`, but I don't see anything that returns a file-like object that can be streamed. This is a disadvantage for many file types which can usefully be processed as they download.

Can a method like this be added?

andrewsg:
Fixes #29

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -30,6 +30,7 @@ import cgi
 import copy
 import hashlib
 from io import BytesIO
+from io import TextIOWrapper
 import logging
 import mimetypes
 import os
@@ -78,6 +79,8 @@ from google.cloud.storage.constants import STANDARD_STORAGE_CLASS
 from google.cloud.storage.retry import DEFAULT_RETRY
 from google.cloud.storage.retry import DEFAULT_RETRY_IF_ETAG_IN_JSON
 from google.cloud.storage.retry import DEFAULT_RETRY_IF_GENERATION_SPECIFIED
+from google.cloud.storage.fileio import BlobReader
+from google.cloud.storage.fileio import BlobWriter
 
 
 _API_ACCESS_ENDPOINT = "https://storage.googleapis.com"
@@ -3405,6 +3408,38 @@ class Blob(_PropertyMixin):
                 if_source_metageneration_not_match=if_source_metageneration_not_match,
                 timeout=timeout,
                 retry=retry,
+            )
+
+    def open(
+        self,
+        mode="r",
+        chunk_size=None,
+        encoding=None,
+        errors=None,
+        newline=None,
+        **kwargs
+    ):
+        if mode == "rb":
+            return BlobReader(self, chunk_size=chunk_size, **kwargs)
+        elif mode == "wb":
+            return BlobWriter(self, chunk_size=chunk_size, **kwargs)
+        elif mode in ("r", "rt"):
+            return TextIOWrapper(
+                BlobReader(self, chunk_size=chunk_size, **kwargs),
+                encoding=encoding,
+                errors=errors,
+                newline=newline,
+            )
+        elif mode in ("w", "wt"):
+            return TextIOWrapper(
+                BlobWriter(self, chunk_size=chunk_size, text_mode=True, **kwargs),
+                encoding=encoding,
+                errors=errors,
+                newline=newline,
+            )
+        else:
+            raise NotImplementedError(
+                "Supported modes strings are 'r', 'rb', 'rt', 'w', 'wb', and 'wt' only."
             )
 
     cache_control = _scalar_property("cacheControl")
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -3426,6 +3426,13 @@ class Blob(_PropertyMixin):
         This method can be used as a context manager, just like Python's
         built-in 'open()' function.
 
+        While reading, as with other read methods, if blob.generation is not set
+        the most recent blob generation will be used. Because the file-like IO
+        reader downloads progressively in chunks, this could result in data from
+        multiple versions being mixed together. If this is a concern, use
+        either bucket.get_blob(), or blob.reload(), which will download the
+        latest generation number and set it.
+
         :type mode: str
         :param mode:
             (Optional) A mode string, as per standard Python `open()` semantics.The first
@@ -3485,7 +3492,7 @@ class Blob(_PropertyMixin):
             >>> client = storage.Client()
             >>> bucket = client.bucket("bucket-name")
 
-            >>> blob = bucket.blob("blob-name.txt")
+            >>> blob = bucket.get_blob("blob-name.txt")
             >>> with blob.open("rt") as f:
             >>>     print(f.read())
 
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -3431,7 +3431,8 @@ class Blob(_PropertyMixin):
         reader downloads progressively in chunks, this could result in data from
         multiple versions being mixed together. If this is a concern, use
         either bucket.get_blob(), or blob.reload(), which will download the
-        latest generation number and set it.
+        latest generation number and set it; or, if the generation is known, set
+        it manually, for instance with bucket.blob(generation=123456).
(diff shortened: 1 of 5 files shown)


andrewsg:
Docstrings are WIP and system tests should be expanded before merging.

google/cloud/storage/blob.py
```diff
@@ -3407,6 +3410,38 @@ def update_storage_class(
                 retry=retry,
             )
 
+    def open(
```
frankyn, line 3413: Do you think it would be more readable to just do first class methods for each BlobReader/Writer combo? 

For example `def open_blob_reader()`

google/cloud/storage/fileio.py
```diff
@@ -0,0 +1,327 @@
...
+
+        # Initialize the upload if necessary.
+        if not self._upload_and_transport:
+            self._initiate_upload()
+
+        upload, transport = self._upload_and_transport
+
+        # Upload chunks. The SlidingBuffer class will manage seek position.
+        for _ in range(num_chunks):
+            upload.transmit_next_chunk(transport)
```
frankyn, line 215: Do you know if partial success occurs does transmit_next_chunk handle the retry from the remote offset?

google/cloud/storage/fileio.py
```diff
@@ -0,0 +1,327 @@
...
+        self._buffer.truncate(0)
+        self._buffer.write(leftover)
+        self._buffer.seek(0)
+
+    def tell(self):
+        """Report how many bytes have been read from the buffer in total."""
+        return self._cursor
+
+    def seek(self, pos, whence=None):
+        raise io.UnsupportedOperation("seek() is not supported for this class.")
```
frankyn, line 310: If there's a partial upload of a chunk then this would be needed to retry correctly.

frankyn (changes requested):
Added some feedback that I want to talk through in implementation. Otherwise this is looking good. I'll review the design document as well.

google/cloud/storage/blob.py
andrewsg: It would be more readable, but this is (regrettably) the standard open() interface shared by all Python file-like I/O.

frankyn: thanks for clarifying.

google/cloud/storage/fileio.py
andrewsg: Fixed. Seek() backwards is supported as long as the buffer has not yet been cleared (which only happens on a successful chunk transmit).

google/cloud/storage/fileio.py
andrewsg: Define "partial success"?

andrewsg:
Added docstrings, seek functionality in the sliding buffer, and other features. PTAL. Remaining questions: can we set up a system test to test end-to-end the error-on-chunk-upload-and-retry behavior? And, even though blob.download methods do not populate the blob generation, can we fetch that generation information from download metadata somehow in order to implement generation lock? (alternative: force blob.reload() before d/l if generation is not populated - race condition possible).

google/cloud/storage/fileio.py
frankyn: GCS may receive a subset of bytes from client upload, leaving the client and server next byte offset unaligned. A request to get current GCS offset can be made per Step [Checking the status of a resumable upload](https://cloud.google.com/storage/docs/performing-resumable-uploads#status-check).

If the client doesn't attempt to recover from this state, the inconsistent byte alignment will cause a 400 error which is not recoverable.

frankyn:
Remaining questions: 

1. Can we set up a system test to test end-to-end the error-on-chunk-upload-and-retry behavior? 
- Not without GCS emulator. You'd be better off mocking out the API request at this time.

2. Even though blob.download methods do not populate the blob generation, can we fetch that generation information from download metadata somehow in order to implement generation lock?  (alternative: force blob.reload() before d/l if generation is not populated - race condition possible).
- You can get generation information from response headers. A similar issue came up in Node.js (for checksums) and [here's example of responses](https://github.com/googleapis/nodejs-storage/pull/1419#discussion_r594670157) specifically the header is named `X-goog-generation:`.

google/cloud/storage/blob.py
```diff
@@ -3407,6 +3412,115 @@ def update_storage_class(
...
+        errors=None,
+        newline=None,
+        **kwargs
+    ):
+        r"""Create a file handler for file-like I/O to or from this blob.
+
+        This method can be used as a context manager, just like Python's
+        built-in 'open()' function.
+
+        :type mode: str
```
frankyn, line 3429: I think `mode` should be a required parameter to not confuse devs.

frankyn (changes requested):
Over LGTM, I have a few nits. 

Also retrying on mismatched offset may be better handled / addressed in resumable media than in this wrapper so we could table it for now and open tracking issues in resumable media repo. WDYT?

google/cloud/storage/fileio.py
andrewsg: Got it. With the changes in this recent update, sliding buffer implements seek() so this feature is as supported as it's going to be in python-storage. If we want further support we need to flesh it out in resumable media.

google/cloud/storage/blob.py
```diff
@@ -3407,6 +3412,115 @@ def update_storage_class(
...
+            If fewer bytes than the chunk_size are requested, the remainder is
+            buffered. For writes, the maximum number of bytes to buffer before
+            sending data to the server, and the size of each request when data
+            is sent. Writes are implemented as a "resumable upload", so
+            chunk_size for writes must be exactly a multiple of 256KiB as with
+            other resumable uploads. The default is 40 MiB.
+
+        :type encoding: str
+        :param encoding:
+            For text mode only, the name of the encoding that the stream will
```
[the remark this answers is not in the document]
andrewsg: That should be optional. I'll mark it.

google/cloud/storage/blob.py
andrewsg: This follows the custom of open() defaulting to read used in the Python standard library. However, it's not that important that we follow that custom in this case. Given that, do you still think we should change it to required?

frankyn: Oh, I didn't realize that open() in Python defaulted to reads. I feel like we should be opinionated because it turns into one of two types which has different implications. I'm open to leaving as-is, but would lean towards being required given the underlying complexities.

google/cloud/storage/fileio.py
frankyn: Let's do that for recovering in a mismatch (address it in resumable media repo). Could you file a tracking bug?

google/cloud/storage/blob.py
```diff
@@ -3426,6 +3426,13 @@ def open(
         This method can be used as a context manager, just like Python's
         built-in 'open()' function.
 
+        While reading, as with other read methods, if blob.generation is not set
+        the most recent blob generation will be used. Because the file-like IO
+        reader downloads progressively in chunks, this could result in data from
+        multiple versions being mixed together. If this is a concern, use
+        either bucket.get_blob(), or blob.reload(), which will download the
+        latest generation number and set it.
```
frankyn, line 3434: Suggestion: Alternatively pass in the value when calling `bucket.blob(generation=123456)` which does not perform a request if the generation is available locally.

google/cloud/storage/blob.py
```diff
@@ -3485,7 +3492,7 @@ def open(
             >>> client = storage.Client()
             >>> bucket = client.bucket("bucket-name")
 
-            >>> blob = bucket.blob("blob-name.txt")
+            >>> blob = bucket.get_blob("blob-name.txt")
```
frankyn, line 3495: Suggestion: Add a comment stating:
```
bucket.get_blob() is retrieving object metadata which will pin generation to the latest value to prevent a race-condition during download.
```

frankyn (changes requested):

frankyn (approved):
LGTM! :ship:

PR #402, opened 2021-03-30
fix: Ensure consistency check in test runs even if expected error occurs

andrewsg:


cojenco (approved):
LGTM, thanks!

tests/unit/test_fileio.py
```diff
@@ -388,7 +388,7 @@ def test_seek(self):
         with self.assertRaises(ValueError):
             pos = buff.tell()
             buff.seek(len(TEST_BINARY_DATA) + 1)
-            self.assertEqual(pos, buff.tell())
+        self.assertEqual(pos, buff.tell())
```
frankyn, line 391: what happened here? is there more context

cojenco: Line 390 raises the exception in which the destination byte is not contained in buffer, so the test never hits Line 391. This change ensures all assertions are run.

frankyn (approved):

PR #412, opened 2021-04-16
chore: migrate to owl bot

parthea:


CONTRIBUTING.rst
```diff
@@ -21,8 +21,8 @@ In order to add a feature:
 - The feature must be documented in both the API and narrative
   documentation.
 
-- The feature must work fully on the following CPython versions:  2.7,
-  3.5, 3.6, 3.7 and 3.8 on both UNIX and Windows.
+- The feature must work fully on the following CPython versions:
```
busunkim96, line 24: I think this library still supports 2.7 so the CONTRIBUTING.rst may need to be excluded temporarily.

setup.py
```diff
@@ -17,6 +17,20 @@
 
 import setuptools
 
+# Disable version normalization performed by setuptools.setup()
```
busunkim96, line 20: Is this intentionally being rolled into this PR?

parthea: This line is already in master. I think the branch is out of date.

parthea: I'm going to retry the migration.

tseaver (commented):
LGTM with the exception of `CONTRIBUTING.rst`, as pointed out by @busunkim96

setup.py
parthea: This issue was due to https://github.com/googleapis/repo-automation-bots/issues/1752. I'm going to force push the branch.

parthea:
I'm going to force push this branch again due to issue googleapis/repo-automation-bots#1752. I'll also address the review comments at the same time.

noxfile.py
```diff
@@ -129,14 +154,28 @@ def system(session):
         "google-cloud-iam",
         "google-cloud-pubsub < 2.0.0",
         "google-cloud-kms < 2.0dev",
+        "-c",
+        constraints_path,
```
tseaver, line 158: Ugh, this change is now in conflict with #433, which works around the breakage incurred by `googleapis-common-protos` dropping Python 2.7 support.  Likely we need to tweak `synth.py` to preserve that change.

tseaver (changes requested):

noxfile.py
parthea: Done.

CONTRIBUTING.rst
parthea: Done.

parthea:
@tseaver  PTAL

tseaver (approved):

PR #430, opened 2021-05-05
tests: fix unit tests broken by dependency drift

tseaver:
Pin `googleapis-common-protos` FBO python2.7

Closes #429

Fix 'TestClient.test_ctor_wo_project', broken by https://github.com/googleapis/python-cloud-core/pull/51

Closes #427

PR #431, opened 2021-05-05
refactor: add / use 'Client._get_resource' method

tseaver:
Use an explicit helper client method for `GET` requests, rather than manipulating client's private `_connection.api_request`.  As a benefit, tests get *way* clearer.

Toward #38

~~Based on top of the branch from #430.  I will rebase when that PR merges.~~

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -2828,13 +2828,12 @@ class Blob(_PropertyMixin):
         if requested_policy_version is not None:
             query_params["optionsRequestedPolicyVersion"] = requested_policy_version
 
-        info = client._connection.api_request(
-            method="GET",
-            path="%s/iam" % (self.path,),
+        info = client._get_path(
+            "%s/iam" % (self.path,),
             query_params=query_params,
-            _target_object=None,
             timeout=timeout,
             retry=retry,
+            _target_object=None,
         )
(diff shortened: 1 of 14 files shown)


tseaver:
I rebased this PR on top of #431, in order to avoid having the python 2.7 system test install fail.  I will rebase to master once #431 is merged.

tseaver:
Flaky systest reported in #435.  Rerunning tests.

google/cloud/storage/blob.py
```diff
@@ -2829,13 +2828,12 @@ def get_iam_policy(
         if requested_policy_version is not None:
             query_params["optionsRequestedPolicyVersion"] = requested_policy_version
 
-        info = client._connection.api_request(
-            method="GET",
-            path="%s/iam" % (self.path,),
+        info = client._get_resource(
+            "%s/iam" % (self.path,),
```
chrisrossi, line 2832: This is more of just a style thing, but in modern Python, should we be preferring `str.format` or `f` strings if Python 2 support has been dropped?

chrisrossi (approved):
LGTM!

google/cloud/storage/blob.py
tseaver: Once we drop Python2, I would definitely prefer f-strings.

PR #437, opened 2021-05-07
fix: replace python lifecycle action parsing ValueError with warning

cojenco:
Replace python Lifecycle action parsing ValueError with a user warning in cases where the client attempts to parse unexpected fields. This will help future proof this feature by preventing runtime errors for parsing unsupported actions.

### DIFF
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -2362,7 +2362,7 @@ class Bucket(_PropertyMixin):
                 yield LifecycleRuleSetStorageClass.from_api_repr(rule)
             else:
                 warnings.warn(
-                    "Unknown lifecycle rule by the client: {}. Please upgrade your client.".format(
+                    "Unknown lifecycle rule type received: {}. Please upgrade to the latest version of google-cloud-storage.".format(
                         rule
                     ),
                     UserWarning,
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -2362,7 +2362,11 @@ class Bucket(_PropertyMixin):
                 yield LifecycleRuleSetStorageClass.from_api_repr(rule)
             else:
                 warnings.warn(
-                    "Unknown lifecycle rule: {}".format(rule), UserWarning, stacklevel=1
+                    "Unknown lifecycle rule by the client: {}. Please upgrade your client.".format(
+                        rule
+                    ),
+                    UserWarning,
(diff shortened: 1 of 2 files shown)


andrewsg (approved):
Thanks!

google/cloud/storage/bucket.py
```diff
@@ -2361,7 +2361,9 @@ def lifecycle_rules(self):
             elif action_type == "SetStorageClass":
                 yield LifecycleRuleSetStorageClass.from_api_repr(rule)
             else:
-                raise ValueError("Unknown lifecycle rule: {}".format(rule))
+                warnings.warn(
```
tritone, line 2364: Does it cause issues if nothing is yielded here?

cojenco: Let me double check on that, thanks! Any insight on this @andrewsg?

andrewsg: I considered that in my review but decided it looked to me like because it's in a for loop, yielding nothing will behave identically to no rule at all, which is probably acceptable in this case. I'm not familiar with how people use this feature, however.

tritone: Gotcha, I guess it will just continue to the next rule in the list and/or end the generator if nothing is left.

I was discussing this issue with someone who pointed me to this code in the Java Bigtable client: https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/EncryptionInfo.java#L48 . I like the message there-- could we also have the warning suggest upgrading to a newer client version?

cojenco: To the point of future proof, I think suggesting upgrading to a newer client version would make sense. Will incorporate that in!

google/cloud/storage/bucket.py
```diff
@@ -2361,7 +2361,13 @@ def lifecycle_rules(self):
...
+                    "Unknown lifecycle rule by the client: {}. Please upgrade your client.".format(
```
tritone, line 2365: I would say something like Unknown lifecycle rule type received: {}. Please upgrade to the latest version of google-cloud-storage.

How does that sound?

tritone (approved):

google/cloud/storage/bucket.py
cojenco: That sounds much better. Thanks for the review!

PR #439, opened 2021-05-08
chore: add library type to .repo-metadata.json

parthea:


.repo-metadata.json
```diff
@@ -6,6 +6,7 @@
   "issue_tracker": "https://issuetracker.google.com/savedsearches/559782",
   "release_level": "ga",
   "language": "python",
+  "library_type": "GAPIC_MANUAL",
```
tseaver, line 9: I'm not sure what that constant means, but `google-cloud-storage` does not use any gRPC at all, nor do we report any GAPIC-related headers (see https://github.com/googleapis/google-cloud-python/pull/9549/files).

busunkim96: Hi Tres,

These are labels to classify libraries. I copied the table from go/library-types (internal doc). I also think this is GAPIC_MANUAL, but please let us know if you think this falls into a different category.

<meta charset="utf-8"><br /><div dir="ltr" style="margin-left:0pt;" align="left">

Enum / Type | Description | Example package
-- | -- | --
UNSPECIFIED | # Default, applied by DRIFT, if library type is not set (no change from existing schema) | N/A
REST | # A generated Apiary ‘client’ | Google API Node.js Client
GAPIC_AUTO | # Purely auto-generated GAPIC ‘client’ | Python Client for the Cloud Tasks API
GAPIC_COMBO | # GAPIC ‘client’ augmented with handwritten code/helpers | Python Client of the Cloud Vision API
GAPIC_MANUAL | # Handwritten, modern ‘client’ with GAPIC-like surface | Google.Cloud.PubSub.V1 (.NET)
AUTH | # Authentication libraries for each language | Python
AGENT | # Service/API instrumentation tools/decorators | Debugger Node.js
INTEGRATION | # e.g. ORM, framework plug-in, or extension | Cloud Spanner support for Django (Python)
CORE | # Common helpers e.g. GAX, common dependencies | Google Cloud Shared Dependencies (Java), Cloud IAM protos and grpc services (Ruby),Core Library for Google Client Libraries (Python)
OTHER | # e.g. API emulator, internal tool, API-specific common libraries (e.g. across admin/data libraries for one API) | Bigtable Emulator,Sloth

</div>

tseaver: The Python storage client doesn't really fit any of those categories:  it is entirely hand-written, served over REST, with no GAPIC or GAPIC-like features that I know of.  Have a look at the [`synth.py`](https://github.com/googleapis/python-storage/blob/2532d506b44fc1ef0fa0a996822d29e7459c465a/synth.py#L1-L40), which is used only for `templated_files` support.

If that list is canonical, then I guess `GAPIC_MANUAL` is arguably the closest, but it still isn't close.

busunkim96 (approved):

PR #445, opened 2021-05-17
fix: add ConnectionError to default retry

### LINKED ISSUE #426: Add Python ConnectionError in list of retry
Hi, Could you add 
https://docs.python.org/3/library/exceptions.html#ConnectionError just like https://github.com/googleapis/python-bigquery/blob/be3c49a72f0e04de4055f5ca7a99f821c2c8f240/google/cloud/bigquery/retry.py#L26 in list of retry

This ConnectionError is seperate from requests.ConnectionError

cojenco:
This adds python built-in [ConnectionError](https://docs.python.org/3/library/exceptions.html#ConnectionError) to default retryable types. 
ConnectionError was recently added in the [BigQuery library](https://github.com/googleapis/python-bigquery/pull/571) to allow retries. 

Fixes #426 🦕


google/cloud/storage/retry.py
```diff
@@ -22,6 +22,7 @@
 
 
 _RETRYABLE_TYPES = (
+    ConnectionError,
```
tritone, line 25: ConnectionError has a bunch of subclasses: https://docs.python.org/3/library/exceptions.html#ConnectionError . Just wanted to confirm that we consider all of these retryable?

Also, is there a test where we should add this as a test case?

tritone (commented):
thanks!

google/cloud/storage/retry.py
tseaver: One or more tests should be added to the `Test_should_retry` class in `tests/unit/test_retry.py` (see the [BigQuery PR](https://github.com/googleapis/python-bigquery/pull/571/files) for examples).

Also, that PR makes `requests.exceptions.ConnectionError` an explicit retryable type.  The `requests` version does *not* derive from the stdlib type:

```python
$ .nox/unit-3-9/bin/python
Python 3.9.0 (default, Apr 20 2021, 11:50:03) 
[GCC 9.3.0] on linux
>>> import requests.exceptions
>>> requests.exceptions.ConnectionError
<class 'requests.exceptions.ConnectionError'>
>>> requests.exceptions.ConnectionError.__mro__
(<class 'requests.exceptions.ConnectionError'>, <class 'requests.exceptions.RequestException'>, <class 'OSError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>)
```

tseaver (changes requested):

google/cloud/storage/retry.py
tseaver: Note that `ConnectionError` is not a stdlib type in Python 2.7:

```python
$ .nox/unit-2-7/bin/python
Python 2.7.17 (default, Apr 20 2021, 14:27:04) 
[GCC 9.3.0] on linux2
>>> ConnectionError
  File "<stdin>", line 1, in <module>
NameError: name 'ConnectionError' is not defined
```

We need to do something like:

```python
try:
    _RETRYABLE_STDLIB_TYPES = {ConnectionError,)
except NameError:
    _RETRYABLE_STDLIB_TYPES = {}

```
and then, below:
```python
RETRYABLE_TYPES = _RETRYABLE_STDLIB_TYPES + (
    api_exceptions.TooManyRequests,  # 429
    ...
)

**Review by tseaver — CHANGES_REQUESTED** [review]

File: google/cloud/storage/retry.py (hunk 1 above)
Comment (cojenco) [reply google/cloud/storage/retry.py:25]: @tritone @tseaver Thanks for reviewing! Looking at the python3 docs and [previous discussion in the Big Query](https://github.com/googleapis/python-bigquery/issues/434), ConnectionError and its subclasses are retryable. From users feedback, this error occurs when there are multiple Cloud Functions running and/or subsequent reinvocation of Cloud Functions. The [retry test](https://github.com/googleapis/python-storage/blob/3fb5086d54ebe487e54cb5a254c0899c91faaf8e/tests/unit/test_retry.py#L40) is set up with a loop that tests each default retry. I'll double check to make the changes and add one or more specific tests. Thanks!

Comment (cojenco) [reply google/cloud/storage/retry.py:25]: @tseaver I noticed that `ConnectionError` is not a stdlib type in Python 2.7 upon submission. Changed PR to draft to implement changes. This is exactly what I need. Thanks so much for the pointers!

File: tests/unit/test_retry.py [hunk 2]
```diff
@@ -59,6 +59,16 @@ def test_w_requests_connection_error(self):
...
+    def test_w_stdlib_connection_error(self):
+        from google.cloud.storage import retry
+
+        try:
+            exc = ConnectionError()
+            self.assertTrue(self._call_fut(exc))
+            self.assertTrue(ConnectionError in retry._RETRYABLE_TYPES)
+        except NameError:
+            pass
+
```
tseaver, line 71: Let's mark this testcase as skipped on Python2.  Also, we don't really need to make the assertion about the error being in the list, so e.g.:
```suggestion
    @unittest.skipUnless(
        "ConnectionError" in __builtins__.__dict__,
        "No builtin 'ConnectionError' in Python 2",
    )
    def test_w_stdlib_connection_error(self):
        exc = ConnectionError()
        self.assertTrue(self._call_fut(exc))
```

tseaver (changes requested):

tests/unit/test_retry.py
```diff
@@ -59,6 +59,16 @@ def test_w_requests_connection_error(self):
         exc = ValueError("testing")
```
tritone, line 59: Side note, I don't understand what this test is doing-- seems to raise a ValueError but the title implies requests.ConnectionError?

tests/unit/test_retry.py
tritone: +1 to this.

tests/unit/test_retry.py
tseaver: I saw that, too.  `git blame` says I introduced it in #308.  @cojenco Can you add a commit to this PR renaming that test to `test_miss_w_stdlib_error`?

tseaver:
@googlebot I fixed it.

cojenco:
@googlebot I fixed it.

tseaver:
@cojenco I don't know what the CLAbot is on about.  As a Googler, I believe you can manually clear the `cla: no` label and set `cla: yes` instead.

cojenco:
Thanks @tseaver! Manually added the cla label. I'm looking at my configs. So when I commited the suggestion directly on Github, the email logged was a <users.noreply.github.com> account.

tseaver:
Ugh, I tested that syntax at the Python command prompt, where `__buitins__` is a module.  Per the [Python docs for the `builtins` module](https://docs.python.org/3/library/builtins.html), we should probably not rely on it being *either* the module *or* its dict, and instead move to something like:

```python
try:
    ConnectionError
except NameError:
    _HAS_CONNECTION_ERROR = False
else:
    _HAS_CONNECTION_ERROR = True
```

and then use that flag to control the skip:

```python
    @unittest.skipUnless(_HAS_CONNECTION_ERROR, "...")
```

Since my suggestion didn't work out, *and* messed up the CLAbot, maybe it would be best to rebase away 
bcfe2be  and start fresh.

tseaver (approved):

tritone (approved):

PR #447, opened 2021-05-20
feat: media operation retries can be configured using the same interface as with non-media operation

andrewsg:
All media operation calls (downloads and uploads) can be configured with Retry objects and ConditionalRetryPolicy objects, nearly identically to non-media operations. This is accomplished by converting the Retry object to a google-resumable-media-python library RetryStrategy object at the point of entry to that library.

Custom predicates of Retry objects (for instance set with Retry(predicate=...)) are not supported for media operations; they will be replaced with a media-operation-specific predicate.

This change is backwards-compatible for users of public methods using num_retries arguments to configure uploads; num_retries continue to be supported but the deprecation warning remains in effect. They will be fully removed and replaced with Retry objects in the future.

With this change, the default parameters for a media operations retry changes to be uniform with non-media operation retries. Specifically, **the retry deadline for media operation retries becomes 120 seconds** unless otherwise configured.


andrewsg:
Tests and coverage are complete here, but docstrings are still WIP. It should be ready to be reviewed except for the docstrings. I'll mark as DNS to ensure it's not submitted without documentation complete.

google/cloud/storage/_helpers.py
```diff
@@ -563,3 +570,37 @@ def _bucket_bound_hostname_url(host, scheme=None):
...
+    :rtype: google.resumable_media.RetryStrategy
+    :returns: A RetryStrategy with all applicable attributes copied from input,
+              or a RetryStrategy with max_retries set to 0 if None was input.
+    """
+
+    if retry is not None and num_retries is not None:
+        raise ValueError("num_retries and retry arguments are mutually exclusive")
+
+    elif retry is not None:
+        return resumable_media.RetryStrategy(
```
frankyn, line 597: It looks like error types are not translatable in this case?

google/cloud/storage/blob.py
```diff
@@ -958,6 +970,7 @@ def _do_download(
                 end=end,
                 checksum=checksum,
             )
+            download._retry_strategy = retry_strategy
```
frankyn, line 973: IIUC, this is accessing a private variable maiintained in ResumableMedia, should resumable media expose a mutable setting instead?

google/cloud/storage/blob.py
```diff
@@ -943,7 +937,25 @@ def _do_download(
...
+            This private method does not accept ConditionalRetryPolicy values
+            because the information necessary to evaluate the policy is instead
+            evaluated in client.download_blob_to_file().
+
+            See the retry.py source code and docstrings in this package
+            (google.cloud.storage.retry) for information on retry types and how
+            to configure them.
         """
+
+        retry_strategy = _api_core_retry_to_resumable_media_retry(retry)
```
frankyn, line 957: Would it be reasonable to expose a configurable option for mutating which errors are retried in ResumableMedia package or a pass in a lambda?

frankyn (changes requested):

google/cloud/storage/_helpers.py
```diff
@@ -563,3 +570,37 @@ def _bucket_bound_hostname_url(host, scheme=None):
         return host
 
     return "{scheme}://{host}/".format(scheme=scheme, host=host)
+
+
+def _api_core_retry_to_resumable_media_retry(retry, num_retries=None):
+    """Convert google.api.core.Retry to google.resumable_media.RetryStrategy.
+
+    Custom predicates are not translated.
```
tritone, line 578: Is this clear to the end user?

google/cloud/storage/blob.py
```diff
@@ -943,7 +937,25 @@ def _do_download(
             downloads where chunk_size is set) an INFO-level log will be
             emitted. Supported values are "md5", "crc32c" and None. The default
             is "md5".
+
+        :type retry: google.api_core.retry.Retry
+        :param retry: (Optional) How to retry the RPC. A None value will disable
+            retries. A google.api_core.retry.Retry value will enable retries,
```
tritone, line 943: Does this change the default behavior to be no retries? I would expect the default value for this retry strategy to be some reasonable standard policy rather than None. Or is this okay here because the method is not one that end users will call directly?

google/cloud/storage/blob.py
```diff
@@ -2162,14 +2189,18 @@ def _do_upload(
...
-            # this library to solve this problem.
-            num_retries = 0
+
+        # Handle ConditionalRetryPolicy.
+        if isinstance(retry, ConditionalRetryPolicy):
+            # Conditional retries are designed for non-media calls, which change
+            # arguments into query_params dictionaries. Media operations work
+            # differently, so here we make a "fake" query_params to feed to the
+            # ConditionalRetryPolicy.
+            query_params = {
```
tritone, line 2199: If `if_generation_match` and `if_metageneration_match` are not set, will the values in this dict be 0 or `None`? I'm a little confused because the docstring indicates that the param is optional but the type is  listed as`long`.

google/cloud/storage/_helpers.py
andrewsg: I'll have this warning in every single public method docstring, so with that it should be clear.

google/cloud/storage/_helpers.py
andrewsg: You mean custom predicates? Yes, those are not translatable. It'll be documented wherever possible.

google/cloud/storage/blob.py
andrewsg: That last bit is correct; private methods such as this one don't have a retry default, but public methods do and they pass them through.

google/cloud/storage/blob.py
andrewsg: I did think about that, but because both error codes and exceptions are covered, they're covered in slightly different ways and exceptions are mutated through the libraries we're using in the transport layer, I felt it was not practical for this PR. We would need to think about a more comprehensive unification of code or else an abstraction of the exceptions and error codes for that to work.

google/cloud/storage/blob.py
andrewsg: Yeah, I noticed that. This is the private variable we were already manipulating in our code (that was how num_retries was implemented). I don't know the history behind that. In the interest of keeping this PR small I chose not to refactor that. If we are confident about our interface we could potentially expose this in the Download and Upload constructors instead.

google/cloud/storage/blob.py
andrewsg: They will be None. Optional in the docstring means it may be either the declared type or None.

google/cloud/storage/blob.py
```diff
@@ -2162,14 +2189,18 @@ def _do_upload(
...
-            # Uploads are only idempotent (safe to retry) if
-            # if_metageneration_match is set. If it is not set, the default
-            # num_retries should be 0. Note: Because retry logic for uploads is
-            # provided by the google-resumable-media-python package, it doesn't
-            # use the ConditionalRetryStrategy class used in other API calls in
-            # this library to solve this problem.
-            num_retries = 0
+
+        # Handle ConditionalRetryPolicy.
+        if isinstance(retry, ConditionalRetryPolicy):
```
cojenco, line 2194: Will there be cases where `if_metageneration_match is not None `and `retry=None`? Or is it not a concern to this private method? IIUC, it seems like `if_metageneration_match` will be picked up as query params in the upload url eventually, but at the same time we will have retry as `resumable_media.RetryStrategy(max_retries=0)`? Will the conditional retry still be triggered?

andrewsg: Yes, `if_metageneration_match` not being `None` and `retry` being `None` is a valid state here. `query_params` inside this conditional _only_ exists for the purpose of evaluating a `ConditionalRetryPolicy`. In any scenario where there is no `ConditionalRetryPolicy`, this code will not fire, but `if_generation_match` and `if_metageneration_match` will still be consumed properly and added to the request later in the process.

andrewsg:
Currently working through some absolutely brutal test merge conflicts, will ask for further review when done.

andrewsg:
@frankyn @tritone PTAL. Docstrings added, all comments responded to, merge conflict resolved.

@tseaver As discussed in chat, also PTAL due to significant potential for error on my part in resolving blob method -> client method-related merge conflicts.

tests/unit/test_fileio.py
```diff
@@ -37,7 +39,12 @@ def test_attributes(self):
         self.assertTrue(reader.seekable())
         self.assertTrue(reader.readable())
         self.assertFalse(reader.writable())
-        self.assertEqual(256, reader._chunk_size)
+        self.assertEqual(reader._chunk_size, 256)
+        self.assertEqual(reader._retry, DEFAULT_RETRY)
+
```
tseaver, line 44: I would generally split this into two test cases, e.g. `test_ctor_w_defaults` and `test_ctor_w_explicit`).

tests/unit/test_fileio.py
```diff
@@ -64,22 +71,22 @@ def read_from_fake_data(start=0, end=None, **_):
...
 
         # Read all remaining data.
         self.assertEqual(reader.read(), TEST_BINARY_DATA[28:])
         self.assertEqual(blob.download_as_bytes.call_count, 4)
         blob.download_as_bytes.assert_called_with(
-            start=28, end=None, checksum=None, **download_kwargs
+            start=28, end=None, checksum=None, retry=DEFAULT_RETRY, **download_kwargs
         )
 
         reader.close()
```
tseaver, line 92: Do we need a testcase which passes a non-default `retry` to reader, and verifies that it is passed through?

tests/unit/test_fileio.py
```diff
@@ -209,7 +223,11 @@ def test_attributes(self):
         self.assertFalse(writer.seekable())
         self.assertFalse(writer.readable())
         self.assertTrue(writer.writable())
-        self.assertEqual(256 * 1024, writer._chunk_size)
+        self.assertEqual(writer._chunk_size, 256 * 1024)
+
```
tseaver, line 227: I would likewise split this testcase.

tests/unit/test_fileio.py
```diff
@@ -261,6 +279,7 @@ def test_write(self):
             None,
             NUM_RETRIES,
             chunk_size=chunk_size,
+            retry=None,
             **upload_kwargs
         )
         upload.transmit_next_chunk.assert_called_with(transport)
```
tseaver, line 285: Do we need a testcase for `write` that assures that a non-default `retry` is passed through?

tests/unit/test_client.py
```diff
@@ -1619,46 +1560,9 @@ def test_list_blobs_w_all_arguments_and_user_project(self):
...
-        self.assertEqual("%s://%s" % (scheme, netloc), client._connection.API_BASE_URL)
-        self.assertEqual(
-            path, "/".join(["", "storage", client._connection.API_VERSION, "b"])
-        )
-        parms = dict(urlparse.parse_qsl(qs))
-        self.assertEqual(parms["project"], PROJECT)
-        self.assertEqual(parms["projection"], "noAcl")
-
-        with pytest.raises(ValueError, match="URI scheme must be gs"):
-            client.download_blob_to_file("http://bucket_name/path/to/object", file_obj)
```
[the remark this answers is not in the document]
andrewsg: Yes, I believe it happened during the automerge. The linter alerted me to a number of test cases that were duplicated word-for-word with the same name.

google/cloud/storage/blob.py
```diff
@@ -3371,9 +3658,12 @@ def open(
...
             supported: "if_generation_match", "if_generation_not_match",
-            "if_metageneration_match", "if_metageneration_not_match", "timeout".
-            For uploads only, the following additional arguments are supported:
-            "content_type", "num_retries", "predefined_acl", "checksum".
+            "if_metageneration_match", "if_metageneration_not_match", "timeout",
+            "retry". For uploads only, the following additional arguments are
+            supported: "content_type", "num_retries", "predefined_acl",
+            "checksum". "num_retries" is supported for backwards-compatibility
+            reasons only; please use "retry" with a Retry object or
+            ConditionalRetryPolicy instead.
```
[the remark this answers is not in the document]
andrewsg: OK, good to know. Do you think I should change anything here to make the merge easier for that PR?

tests/unit/test__helpers.py
```diff
@@ -632,17 +632,6 @@ def test_none(self):
...
-    def __init__(self, *responses):
-        self._responses = responses
-        self._requested = []
-
-    def api_request(self, **kw):
-        self._requested.append(kw)
-        response, self._responses = self._responses[0], self._responses[1:]
-        return response
-
-
```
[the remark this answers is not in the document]
andrewsg: Yeah I was surprised, but the cover tool pointed out that nothing is using it anywhere at all.

google/cloud/storage/client.py
```diff
@@ -1259,6 +1259,7 @@ def list_blobs(
             max_results=max_results,
             extra_params=extra_params,
             page_start=_blobs_page_start,
+            timeout=timeout,
```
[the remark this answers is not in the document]
andrewsg: Good catch!!

tests/unit/test_client.py
```diff
@@ -1432,18 +1434,135 @@ def test_download_blob_to_file_with_uri(self):
...
+
+    def test_download_blob_to_file_w_conditional_retry_fail(self):
+        self._download_blob_to_file_helper(
+            use_chunks=True,
+            raw_download=True,
+            retry=DEFAULT_RETRY_IF_GENERATION_SPECIFIED,
+            expect_condition_fail=True,
+        )
+
+    def test_list_blobs(self):
```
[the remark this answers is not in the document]
andrewsg: Removed.

tests/unit/test_client.py
```diff
@@ -1432,18 +1434,135 @@ def test_download_blob_to_file_with_uri(self):
...
+            self.assertEqual(blobs, [])
+            connection.api_request.assert_called_once_with(
+                method="GET",
+                path="/b/%s/o" % BUCKET_NAME,
+                query_params={"projection": "noAcl"},
+                timeout=self._get_default_timeout(),
+                retry=DEFAULT_RETRY,
+            )
+
+    def test_list_blobs_w_all_arguments_and_user_project(self):
```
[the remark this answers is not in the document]
andrewsg: Removed.

tests/unit/test_client.py
```diff
@@ -1809,6 +1726,7 @@ def test_list_blobs_w_explicit_w_user_project(self):
             max_results=expected_max_results,
             extra_params=expected_extra_params,
             page_start=expected_page_start,
+            timeout=timeout,
```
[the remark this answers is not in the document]
andrewsg: That's... strange... in my source code it already does. I don't understand why it's not showing up on the diff. I will double-check after the next push.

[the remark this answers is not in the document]
andrewsg: Scratch that, found the issue. Fixed. Thanks!

tests/unit/test_fileio.py
andrewsg: OK, added test as `test_forced_default_retry`

tests/unit/test_fileio.py
andrewsg: Added as `test_retry_passed_through`

tests/unit/test_fileio.py
andrewsg: done

google/cloud/storage/blob.py
```diff
@@ -1077,6 +1092,28 @@ def download_to_file(
...
+
+            See the retry.py source code and docstrings in this package
+            (google.cloud.storage.retry) for information on retry types and how
+            to configure them.
+
+            Media operations (downloads and uploads) do not support non-default
+            predicates in a Retry object. The default will always be used. Other
+            configuration changes for Retry objects such as delays and deadlines
+            are respected.
+
```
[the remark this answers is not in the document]
andrewsg: OK, adding as a todo after this is merged

tests/unit/test_client.py
```diff
@@ -1735,6 +1651,7 @@ def test_list_blobs_w_defaults_w_bucket_obj(self):
             max_results=expected_max_results,
             extra_params=expected_extra_params,
             page_start=expected_page_start,
+            timeout=self._get_default_timeout(),
```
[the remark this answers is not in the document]
andrewsg: Done

google/cloud/storage/blob.py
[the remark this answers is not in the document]
tseaver: I think it is my turn to be in the barrel for that one. :)

andrewsg:
@tseaver PTAL. Thank you for your attention, I have much more confidence in the merge now.

google/cloud/storage/blob.py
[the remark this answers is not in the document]
tseaver: If you want to open an issue and assign it to me, that would be fine.

tseaver (approved):

frankyn (approved):
LGTM, thanks Andrew!

PR #449, opened 2021-05-24
fix: drop gcloud-python header from user agent

### LINKED ISSUE #25: Storage, Core: Once x-goog-api-client headers, gccl, is used for storage, remove gcloud-python
internal bug: b/143493862

The Python Client has a patch https://github.com/googleapis/google-cloud-python/pull/9548 to temporarily collect metrics until we address a pipeline issue. This is a temporary fix and should be removed once that is done.

### LINKED ISSUE #15990: Once x-goog-api-client headers, gccl, is used for storage, remove gcloud-python
internal bug: b/143493862

The Python Client has a patch https://github.com/googleapis/google-cloud-python/pull/9548 to temporarily collect metrics until we address a pipeline issue. This is a temporary fix and should be removed once that is done.

tseaver:
This reverts commit dc817e9d0b94a84ef5ed713bfe300b6d00ec15da, added in https://github.com/googleapis/google-cloud-python/pull/9551

Closes googleapis/google-cloud-python#15990.

Note:  do not merge until we have confirmation that the GCS back-end is using `gccl` for header processing.


parthea:
Blocked by b/188672632

google/cloud/storage/_http.py
```diff
@@ -53,11 +53,6 @@ def __init__(self, client, client_info=None, api_endpoint=None):
         self.ALLOW_AUTO_SWITCH_TO_MTLS_URL = api_endpoint is None
         self._client_info.client_library_version = __version__
 
-        # TODO: When metrics all use gccl, this should be removed #9552
-        if self._client_info.user_agent is None:  # pragma: no branch
-            self._client_info.user_agent = ""
-        self._client_info.user_agent += " gcloud-python/{} ".format(__version__)
```
parthea, line 59: When this code is removed, will the `user-agent` header be blank?

frankyn: Hi @parthea, this change was made because tracking does not support x-goog-api-client header yet. If we were to remove this now, we would now see version numbers internally.

PR #450, opened 2021-05-25
tests: add retry conformance test framework and test always idempotent operations

cojenco:
Retry strategies are included in the client library when it's safe to do so. These retry strategies are complex and need automated integration tests to ensure they work and continue to work in the future. 

This adds retry strategy conformance tests that are expected to run against a local GCS API emulator and validate correct behavior around retryable errors, idempotency considerations and preconditions. This PR includes the tests and schema json for:
- Scenario 1 - Libraries retry all idempotent operations for retriable error codes
- Scenario 2 - For conditionally idempotent operations with retryable error codes, libraries retry when precondition is present

Tests are loaded by the json directly. An individual test case is generated for each scenario ID, instruction set, method, and method_invocation_mapping_library_method). Test names are displayed as `test-S{scenario_id}-{method}-{client-library-method-name}`, for example, `test-S1-storage.buckets.insert-bucket_create`


noxfile.py
```diff
@@ -29,6 +29,9 @@
 DEFAULT_PYTHON_VERSION = "3.8"
 SYSTEM_TEST_PYTHON_VERSIONS = ["2.7", "3.8"]
 UNIT_TEST_PYTHON_VERSIONS = ["2.7", "3.6", "3.7", "3.8", "3.9"]
+CONFORMANCE_TEST_PYTHON_VERSIONS = ["2.7", "3.8"]
+
+_DEFAULT_STORAGE_HOST = "https://storage.googleapis.com"
```
tseaver, line 34: Changes to `noxfile.py` need to be preserved by adding appropriate `s.replace` entries to `owlbot.py`:  otherwise, they will vanish in the next bot-generated PR.

tseaver: Also, I reckon we can avoid running the new conformance tests on Python 2, unless there is a compelling reason to believe they would fail differently.

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,449 @@
...
+
+########################################################################################################################################
+### Helper Methods for Populating Resources ############################################################################################
+########################################################################################################################################
+
+
+def _populate_resource_bucket(client, resources):
+    bucket = client.bucket(uuid.uuid4().hex)
+    client.create_bucket(bucket)
+    resources["bucket"] = bucket
```
tseaver, line 271: How are the created resources being cleaned up?  The cleanup needs to be resilient against unexpected failures, both during test runs and while tearing down other resources.

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,449 @@
...
+
+
+def _populate_resources(client, json_resource):
+    resources = {}
+
+    for r in json_resource:
+        func = resource_mapping[r]
+        func(client, resources)
+
+    return resources
```
tseaver, line 315: Is there any reason why we shouldn't just create the bucket, blob, notification, and HMAC key at the beginning of the test run, rather than for each testcase?

tests/conformance/test_conformance.py
tritone: These tests are run against an emulator; when it shuts down at the end of the test session, all created resources disappear.

tests/conformance/test_conformance.py
tritone: Individual test cases may modify/delete the resource. Since it's quick and cheap to create these resources in the emulator, there's not really a good reason to try to deduplicate in an ad-hoc way.

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,449 @@
...
+        bucket.rename_blob(object, new_name, if_generation_match=0)
+    else:
+        bucket.rename_blob(object, new_name)
+
+
+def rewrite_blob(client, _preconditions, bucket, object):
+    new_blob = client.bucket(bucket.name).blob(uuid.uuid4().hex)
+    new_blob.metadata = {"foo": "bar"}
+    if _preconditions:
+        new_blob.rewrite(object, if_generation_match=0)
```
tritone, line 215: Should this use generation = 0 or the generation from the existing object?

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,449 @@
...
+
+def _populate_resource_hmackey(client, resources):
+    hmac_key, secret = client.create_hmac_key(
+        service_account_email=_CONF_TEST_SERVICE_ACCOUNT_EMAIL,
+        project_id=_CONF_TEST_PROJECT_ID,
+    )
+    resources["hmac_key"] = hmac_key
+
+
+resource_mapping = {
```
tritone, line 300: Did you try using pytest fixtures for these? Curious about the pros/cons of that.

tritone (commented):
Coming along well! A few comments.

noxfile.py
cojenco: Thanks @tseaver for the comments. I agree we can remove Python 2 from the conformance tests. Will look into the owlbot user guide and add the entries.

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,449 @@
...
+
+########################################################################################################################################
+### Run Conformance Tests for Retry Strategy ###########################################################################################
+########################################################################################################################################
+
+
+@pytest.mark.parametrize("test_data", _CONFORMANCE_TESTS)
+def test_conformance_retry_strategy(test_data):
+    host = _API_ACCESS_ENDPOINT
+    if host == _DEFAULT_STORAGE_HOST:
```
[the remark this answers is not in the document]
cojenco: I found tests using the utility method from `_helpers.py` but yes I think it's more straightforward to check for the emulator env var directly. Updated, thanks!

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,449 @@
...
+    headers = {
+        "Content-Type": "application/json",
+    }
+    data_dict = {"instructions": {method_name: instructions}}
+    data = json.dumps(data_dict)
+    r = requests.post(preflight_post_uri, headers=headers, data=data)
+    return r.json()
+
+
+def _check_retry_test(host, id):
```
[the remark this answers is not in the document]
cojenco: Thanks for the review @tritone! Currently this method is purely getting the retry test (status) by the unique ID. 

Indeed maybe it should... I will try that out. In that case, the assertions would live within this helper method `_check_retry_test`. I'm still leaning towards having the assertions in `test_conformance_retry_strategy()`.

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,449 @@
...
+                    )
+                    continue
+
+                # Run retry tests on library methods.
+                try:
+                    _run_retry_test(
+                        host, id, function, precondition_provided, **resources
+                    )
+                except Exception as e:
+                    # Should we be catching specific exceptions
```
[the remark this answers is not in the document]
cojenco: Right, I haven't been able to associate an exception with the particular instruction. 

I have a try/exception block here in order to catch the exception and furthermore assert with `expectedSuccess`. For scenarios 3 and 4 where `expectedSuccess` is false, catching the exception prevents the test erroring out.

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,449 @@
...
+        blob.compose(sources)
+
+
+# Method invocation mapping. Methods to retry. This is a map whose keys are a string describing a standard
+# API call (e.g. storage.objects.get) and values are a list of functions which
+# wrap library methods that implement these calls. There may be multiple values
+# because multiple library methods may use the same call (e.g. get could be a
+# read or just a metadata get).
+method_mapping = {
+    "storage.buckets.delete": [delete_bucket],  # S1 start
```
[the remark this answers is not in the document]
cojenco: Good idea! Are you referring to the library methods listed?

tests/conformance/test_conformance.py
cojenco: `if_generation_match=0` works when there are no live versions of the destination object. The destination blob `new_blob` is just instantiated at this point, so the operation will succeed.

tests/conformance/test_conformance.py
cojenco: Currently trying to use pytest fixtures and create individual test cases, WIP

tests/conformance/retry_strategy_test_data.json
```diff
@@ -0,0 +1,299 @@
...
+              "OBJECT"
+            ]
+          },
+          {
+            "name": "storage.hmacKey.delete",
+            "resources": []
+          },
+          {
+            "name": "storage.hmacKey.get",
+            "resources": []
```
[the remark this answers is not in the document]
cojenco: Per discussed, the project-scoped endpoints are not ready in the emulator Retry API yet. Tracking [here](https://github.com/googleapis/google-cloud-cpp/issues/6636).

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,449 @@
...
+    return r.json()
+
+
+def _check_retry_test(host, id):
+    status_get_uri = "{base}{retry}/{id}".format(base=host, retry="/retry_test", id=id)
+    r = requests.get(status_get_uri)
+    return r.json()
+
+
+def _run_retry_test(host, id, func, _preconditions, **resources):
```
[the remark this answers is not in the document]
cojenco: Added docstrings, thanks!

tests/conformance/test_conformance.py
cojenco: @tritone I've made a version using pytest fixtures to populate resources. Given that test cases may modify or delete resources, we're unable to fully utilize the teardown/cleanup benefits that come with pytest.fixtures. Another minus is that given our nested structure, I need to populate ALL resources for each test case, despite that the test case may only need a bucket resouce and nothing else. Therefore, it may not make sense to use pytest fixtures in this case.

tests/conformance/test_conformance.py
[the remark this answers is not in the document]
tritone: Discussed offline-- this refers to the wrappers on library methods that you wrote.

tests/conformance/test_conformance.py
```diff
@@ -321,6 +506,11 @@ def _populate_resources(client, json_resource):
 
 
 def _create_retry_test(host, method_name, instructions):
+    """
+    Initialize a Retry Test resource with a list of instructions and an API method.
+    This offers a mechanism to send multiple retry instructions while sending a single, constant header through all the HTTP requests in a test.
+    See also: https://github.com/googleapis/google-cloud-cpp/tree/main/google/cloud/storage/emulator
```
tritone, line 512: This link should point to the new repo

tests/conformance/test_conformance.py
```diff
@@ -356,94 +549,95 @@ def _delete_retry_test(host, id):
...
+        s = "s{}".format(scenario_id)
+        if s in metafunc.fixturenames:
+            metafunc.parametrize(s, [scenario_id])
+        if m in metafunc.fixturenames:
+            metafunc.parametrize(m, test_data["methods"])
+        if c in metafunc.fixturenames:
+            metafunc.parametrize(c, test_data["cases"])
+
+
+def test_retry_s1_always_idempotent(s1, s1method, s1case):
```
tritone, line 566: This feels like a leak of the schema into the logic of this test file. What's your rationale for including this?

tests/conformance/test_conformance.py
```diff
@@ -356,94 +549,95 @@ def _delete_retry_test(host, id):
...
+    if method_name not in method_mapping:
+        pytest.skip("No tests for operation {}".format(method_name),)
+
+    for function in method_mapping[method_name]:
+        # Create the retry test in the emulator to handle instructions.
+        try:
+            r = _create_retry_test(host, method_name, instructions)
+            id = r["id"]
+        except Exception as e:
+            warnings.warn(
```
tritone, line 599: Again, this and following should be hard failures (for the specific case) rather than warnings. We can discuss the path from here given that there may still be some compliance gaps.

tests/conformance/test_conformance.py
```diff
@@ -79,6 +135,45 @@ def upload_from_string(client, _preconditions, bucket):
...
+        blob_writer = BlobWriter(blob, chunk_size=chunk_size, if_metageneration_match=0)
+        blob_writer.write(bytearray(os.urandom(262144)))
+    else:
+        blob_writer = BlobWriter(blob, chunk_size=chunk_size)
+        blob_writer.write(bytearray(os.urandom(262144)))
+
+
+def blob_create_resumable_upload_session(client, _preconditions, bucket):
+    blob = client.bucket(bucket.name).blob(uuid.uuid4().hex)
+    blob.create_resumable_upload_session()
```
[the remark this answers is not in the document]
cojenco: Currently, this library method doesn't support preconditions and retry. #480 adds preconditions and retry config, will add preconditions to the test if the PR lands.

tests/conformance/test_conformance.py
```diff
@@ -143,6 +252,27 @@ def delete_blob(client, _preconditions, bucket, object):
         bucket.delete_blob(object.name)
 
 
+def bucket_delete_blobs(client, _preconditions, bucket, object):
+    bucket = client.bucket(bucket.name)
+    blob_2 = bucket.blob(uuid.uuid4().hex)
+    blob_2.upload_from_string("foo")
```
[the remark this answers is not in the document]
cojenco: We don't need multiple objects here 😂

tests/conformance/test_conformance.py
cojenco: Done, thanks!

noxfile.py
cojenco: @tseaver Will the `noxfile.py` changes vanish with the existing [exclusion](https://github.com/googleapis/python-storage/blob/6486010c5835d25a9284c6b17730a3c31177e55a/owlbot.py#L40-L42)? 
```
s.move(
    templated_files, excludes=["docs/multiprocessing.rst", "noxfile.py", "CONTRIBUTING.rst"],
)
```
Is there any Owlbot/Synthtool documentation towards how to add `s.replace` entries with noxfile changes? So far I've only come to a [pydoc](https://htmlpreview.github.io/?https://github.com/googleapis/synthtool/blob/master/synthtool/pydoc.html). Would it be something like so?
```
s.replace(
    "google/cloud/python-storage/noxfile.py",
    "UNIT_TEST_PYTHON_VERSIONS = ["2.7", "3.6", "3.7", "3.8", "3.9"]",
    "UNIT_TEST_PYTHON_VERSIONS = ["2.7", "3.6", "3.7", "3.8", "3.9"]\n"
    "CONFORMANCE_TEST_PYTHON_VERSIONS = ["3.8"]"
)
```

tseaver: A couple of scars I have from using `s.replace`:

- The second argument is a regex, so you need to escape a bunch of special characters with `\`, which means the string should be "raw" (see https://docs.python.org/3/library/re.html).
- I generally prefer to use triple-quoted strings for any multi-line replacements:  it is easier to read.

So, for your example, it would be:

```python
s.replace(
    "google/cloud/python-storage/noxfile.py",
    r"UNIT_TEST_PYTHON_VERSIONS = \["2\.7", "3\.6", "3\.7", "3\.8", "3\.9"\]",
    """\
UNIT_TEST_PYTHON_VERSIONS = ["2.7", "3.6", "3.7", "3.8", "3.9"]
CONFORMANCE_TEST_PYTHON_VERSIONS = ["3.8"]"""
)
```

cojenco: Got it, @tseaver. I was following the pydoc and thought I could pass in `before` as a `str` [```replace(sources: Iterable[Union[str, pathlib.Path]], before: str, after: str, flags: int = <RegexFlag.MULTILINE: 8>) -> int```](https://htmlpreview.github.io/?https://github.com/googleapis/synthtool/blob/master/synthtool/pydoc.html)

I'm still kind of confused on the s.replace entries. I've added [a huge chunk of session conformance test in `noxfile.py`](https://github.com/googleapis/python-storage/blob/53b9f7d20dc5feefbcddbd8c8f52e6507acd6ddf/noxfile.py#L154-L185). How would I preserve that change in owlbot.py?

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,699 @@
...
+            host,
+            id,
+            lib_func,
+            precondition_provided,
+            bucket,
+            object,
+            notification,
+            hmac_key,
+        )
+    except Exception as e:
```
andrewsg, line 645: Why catch and log this instead of just letting the exception bubble up and have pytest deal with it?

andrewsg (changes requested):
Great work so far!

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,699 @@
...
+########################################################################################################################################
+### Library methods for mapping ########################################################################################################
+########################################################################################################################################
+
+
+def bucket_get_blob(client, _preconditions, **resources):
+    bucket = resources.get("bucket")
+    object = resources.get("object")
+    bucket = client.bucket(bucket.name)
+    bucket.get_blob(object.name)
```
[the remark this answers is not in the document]
cojenco: For most test cases, we're not using any of the return values. On [line 606](https://github.com/googleapis/python-storage/blob/994a2f6f9e32ab317d0a39ecdce948ee3e2c002f/tests/conformance/test_conformance.py#L606), I am just simply making a func call. Would it be more pythonic to add return values to all of these methods?

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,699 @@
...
+
+def blobwriter_write(client, _preconditions, **resources):
+    import os
+    from google.cloud.storage.fileio import BlobWriter
+
+    chunk_size = 256 * 1024
+    bucket = resources.get("bucket")
+    blob = client.bucket(bucket.name).blob(uuid.uuid4().hex)
+    if _preconditions:
+        blob_writer = BlobWriter(blob, chunk_size=chunk_size, if_generation_match=0)
```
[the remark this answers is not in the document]
cojenco: Good point! Updated, thanks!

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,699 @@
...
+
+
+@pytest.fixture
+def bucket(client):
+    bucket = client.bucket(uuid.uuid4().hex)
+    client.create_bucket(bucket)
+    yield bucket
+    try:
+        bucket.delete(force=True)
+    except Exception:  # in cases where resources are deleted within the test
```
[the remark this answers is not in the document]
cojenco: The exception is a 404 Not Found. This happens when the fixture/resource (bucket, blob, notification) is deleted within the test, and the fixture no longer exists during cleanup.

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,699 @@
...
+
+
+########################################################################################################################################
+### Helper Methods for Emulator Retry API ##############################################################################################
+########################################################################################################################################
+
+
+def _create_retry_test(host, method_name, instructions):
+    """
+    Initialize a Retry Test resource with a list of instructions and an API method.
```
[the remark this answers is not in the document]
cojenco: I've updated the docstrings. Hopefully they're now more informative and clear. 

As for the method name, I agree `_create_retry_test` isn't as self-explanatory as it could be. Though it's sort of symmetrical to the other helper method names😂  I could change it to `_send_retry_test_instructions` WDYT?

tests/conformance/test_conformance.py
cojenco: Catching the exception here helps verify whether or not the test ran successfully, with a variable `success_results`. `success_results` is compared against `expect_success` from the schema. 

For test scenarios 3 and 4, the tests are expected to fail, hence why we're catching the exception.

noxfile.py
```diff
@@ -148,6 +151,40 @@ def system(session):
...
+    conformance_test_folder_path = os.path.join("tests", "conformance")
+
+    # Environment check: Only run tests if the STORAGE_EMULATOR_HOST is set.
+    if (
+        os.environ.get("STORAGE_EMULATOR_HOST", _DEFAULT_STORAGE_HOST)
+        == _DEFAULT_STORAGE_HOST
+    ):
+        session.skip("Set STORAGE_EMULATOR_HOST to run, skipping")
+    # Environment check: Only run tests if the environment variable is set.
+    if not os.environ.get("GOOGLE_APPLICATION_CREDENTIALS", ""):
```
tritone, line 167: Out of curiosity, why is this required?

noxfile.py
```diff
@@ -148,6 +151,40 @@ def system(session):
         session.run("py.test", "--quiet", system_test_folder_path, *session.posargs)
 
 
+@nox.session(python=CONFORMANCE_TEST_PYTHON_VERSIONS)
+def conformance(session):
```
tritone, line 155: Maybe note that this is just retry conformance and/or rename (to avoid confusion with signed URL conformance tests).

tests/conformance/test_conformance.py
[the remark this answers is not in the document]
tritone: I don't think it should be necessary to have return values given that Python uses exceptions for error handling.

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,712 @@
...
+_BYTE_CONTENT = b"12345678"
+
+
+########################################################################################################################################
+### Library methods for mapping ########################################################################################################
+########################################################################################################################################
+
+
+def bucket_get_blob(client, _preconditions, **resources):
+    bucket = resources.get("bucket")
```
tritone, line 50: How do these fail if a resource is missing for some reason?

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,712 @@
...
+    if _preconditions:
+        blob.upload_from_string(_STRING_CONTENT, if_generation_match=0)
+    else:
+        blob.upload_from_string(_STRING_CONTENT)
+
+
+def blob_upload_from_file(client, _preconditions, **resources):
+    from io import BytesIO
+
+    file_obj = BytesIO()
```
tritone, line 383: What is actually uploaded here? I'm surprised you don't need a tempfile.

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,712 @@
...
+    bucket = resources.get("bucket")
+    blob = client.bucket(bucket.name).blob(uuid.uuid4().hex)
+    if _preconditions:
+        blob.upload_from_string(_STRING_CONTENT, if_generation_match=0)
+    else:
+        blob.upload_from_string(_STRING_CONTENT)
+
+
+def blob_upload_from_file(client, _preconditions, **resources):
+    from io import BytesIO
```
tritone, line 381: Any reason not to import this at the toplevel?

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,712 @@
...
+    return r.json()
+
+
+def _run_retry_test(
+    host, id, lib_func, _preconditions, bucket, object, notification, hmac_key
+):
+    """
+    To execute tests against the list of instrucions sent to the Retry API,
+    create a client to send the retry test ID using the x-retry-test-id header
+    in each request. For incoming requests which match the given API method,
```
tritone, line 600: Which match the test ID and API method

tests/conformance/test_conformance.py
tritone: I see-- would the test logs be filled with all these exceptions?

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,712 @@
+# Copyright 2020 Google LLC
```
tritone, line 1: 2021?

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,712 @@
...
+
+import os
+import requests
+import tempfile
+import uuid
+import logging
+import functools
+import pytest
+
+from google.cloud import storage
```
tritone, line 23: Maybe add a brief top-level comment for this file explaining what is going on.

tritone (commented):
Couple nits and questions, otherwise LGTM! This is looking so much cleaner, thank you for all your work!

tests/conformance/test_conformance.py
cojenco: Yep the exceptions here are caught and logged.

noxfile.py
cojenco: This is required in order to instantiate a Client. Checking to see if we have anything similar in python to the `NoCredentials` used in Java.

tseaver: You could use `Client.create_anonymous_client()` if you aren't going to be making real, over-the-wire requests.

tseaver: Or pass in an instance of `google.auth.credentials.AnonymousCredentials` to the client constructor.

cojenco: Awesome, thanks! I'll change to use one of those.

tests/conformance/test_conformance.py
cojenco: The resources are populated through pytest fixtures. If for any reason, a resource is missing, we expect to either see a Fixture Error or an AttributeError. 

If a fixture has a problem, pytest will stop executing fixtures for that test and mark the test as having an error. Another case would be raising an AttributeError when attempting to retrieve an attribute of the resource (most cases retrieving `bucket.name`)

tritone (approved):

tests/conformance/test_conformance.py
```diff
@@ -0,0 +1,729 @@
...
+
+
+@pytest.fixture
+def bucket(client):
+    bucket = client.bucket(uuid.uuid4().hex)
+    client.create_bucket(bucket)
+    yield bucket
+    try:
+        bucket.delete(force=True)
+    except Exception:
```
andrewsg, line 517: As you've identified, these would ideally be narrower so we can find out if there is a problem with clean-up, but I agree it's okay to have this as-is until the emulator is improved.

andrewsg (approved):
Looks great. The conformance test idea has come a long way thanks in part to this implementation and review. Good stuff.

PR #454, opened 2021-06-05
fix: revise blob.compose query parameters `if_generation_match`

### LINKED ISSUE #453: Change `blob.compose` to fit API behavior
Currently `blob.compose()` is misusing `if_generation_match` as a list of [`sourceObjects[].objectPreconditions.ifGenerationMatch`](https://cloud.google.com/storage/docs/json_api/v1/objects/compose#request-body) sent with the API request body. Based on the [doc](https://cloud.google.com/storage/docs/json_api/v1/objects/compose#parameters), `if_generation_match` is an optional query parameter that makes the operation conditional only when the composed object's current generation matches the given value.

cojenco:
This changes `if_generation_match`, `if_metageneration_match`, `if_source_generation_match` for  `blob.compose()` to fit API behavior

- Add pending deprecation warning in cases where `if_generation_match` and `if_metageneration_match` is passed in as a list of long. Add conditional handling for backwards compatibility.
-  Update `if_generation_match` to be correctly included within the API request query parameters if it is type long.  This makes the operation conditional on whether the **composed object's** current generation matches the given value. 
- Add `if_source_generation_match` as an optional parameter that serves as sourceObjects preconditions. This is sent with the API request body. When provided, the composition only performs if the generation of the **source object** that would be used matches this value.
- Revise docstring, example and tests

Fixes #453🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -2986,8 +2986,8 @@ class Blob(_PropertyMixin):
             warnings.warn(
                 "if_generation_match: type list is deprecated and supported for backwards-compatability reasons only."
                 "Use if_source_generation_match instead to match source objects generations.",
-                PendingDeprecationWarning,
-                stacklevel=1,
+                DeprecationWarning,
+                stacklevel=2,
             )
 
             if if_source_generation_match is not None:
@@ -3002,10 +3002,11 @@ class Blob(_PropertyMixin):
 
         if isinstance(if_metageneration_match, list):
             warnings.warn(
-                "if_metageneration_match matches the given value to the destination object's metageneration."
+                "if_metageneration_match: type list is deprecated and supported for backwards-compatability reasons only."
+                "Note that the metageneration to be matched is that of the destination blob."
                 "Please pass in a single value (type long).",
-                PendingDeprecationWarning,
-                stacklevel=1,
+                DeprecationWarning,
+                stacklevel=2,
             )
 
             if_metageneration_match = None
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -3009,6 +3009,7 @@ class Blob(_PropertyMixin):
         timeout=_DEFAULT_TIMEOUT,
         if_generation_match=None,
         if_metageneration_match=None,
+        if_source_generation_match=None,
         retry=DEFAULT_RETRY_IF_GENERATION_SPECIFIED,
     ):
         """Concatenate source blobs into this one.
@@ -3031,18 +3032,23 @@ class Blob(_PropertyMixin):
             Can also be passed as a tuple (connect_timeout, read_timeout).
             See :meth:`requests.Session.request` documentation for details.
 
-        :type if_generation_match: list of long
+        :type if_generation_match: long
         :param if_generation_match:
             (Optional) Make the operation conditional on whether the blob's
             current generation matches the given value.  Setting to 0 makes the
             operation succeed only if there are no live versions of the blob.
-            The list must match ``sources`` item-to-item.
 
         :type if_metageneration_match: list of long
         :param if_metageneration_match:
             (Optional) Make the operation conditional on whether the blob's
             current metageneration matches the given value. The list must match
             ``sources`` item-to-item.
+
+        :type if_source_generation_match: list of long
+        :param if_source_generation_match:
+            (Optional) Make the operation conditional on whether the source blob's
+            current generation matches the given value.
+            The list must match ``sources`` item-to-item.
 
         :type retry: google.api_core.retry.Retry or google.cloud.storage.retry.ConditionalRetryPolicy
         :param retry: (Optional) How to retry the RPC. A None value will disable retries.
@@ -3053,37 +3059,32 @@ class Blob(_PropertyMixin):
             activates it only if certain conditions are met. This class exists to provide safe defaults
             for RPC calls that are not technically safe to retry normally (due to potential data
             duplication or other side-effects) but become safe to retry if a condition such as
-            if_metageneration_match is set.
+            if_generation_match is set.
 
             See the retry.py source code and docstrings in this package (google.cloud.storage.retry) for
             information on retry types and how to configure them.
 
         Example:
-            Compose blobs using generation match preconditions.
+            Compose blobs using source generation match preconditions.
 
             >>> from google.cloud import storage
             >>> client = storage.Client()
             >>> bucket = client.bucket("bucket-name")
 
             >>> blobs = [bucket.blob("blob-name-1"), bucket.blob("blob-name-2")]
-            >>> if_generation_match = [None] * len(blobs)
-            >>> if_generation_match[0] = "123"  # precondition for "blob-name-1"
+            >>> if_source_generation_match = [None] * len(blobs)
+            >>> if_source_generation_match[0] = "123"  # precondition for "blob-name-1"
 
             >>> composed_blob = bucket.blob("composed-name")
-            >>> composed_blob.compose(blobs, if_generation_match)
+            >>> composed_blob.compose(blobs, if_source_generation_match=if_source_generation_match)
         """
         sources_len = len(sources)
-        if if_generation_match is not None and len(if_generation_match) != sources_len:
-            raise ValueError(
-                "'if_generation_match' length must be the same as 'sources' length"
-            )
-
         if (
-            if_metageneration_match is not None
-            and len(if_metageneration_match) != sources_len
+            if_source_generation_match is not None
+            and len(if_source_generation_match) != sources_len
         ):
             raise ValueError(
-                "'if_metageneration_match' length must be the same as 'sources' length"
+                "'if_source_generation_match' length must be the same as 'sources' length"
             )
 
         client = self._require_client(client)
@@ -3092,22 +3093,23 @@ class Blob(_PropertyMixin):
         if self.user_project is not None:
             query_params["userProject"] = self.user_project
 
+        _add_generation_match_parameters(
+            query_params,
+            if_generation_match=if_generation_match,
+            if_metageneration_match=if_metageneration_match,
+        )
+
         source_objects = []
         for index, source in enumerate(sources):
             source_object = {"name": source.name}
+            source_object["generation"] = source.generation
 
             preconditions = {}
             if (
-                if_generation_match is not None
-                and if_generation_match[index] is not None
+                if_source_generation_match is not None
+                and if_source_generation_match[index] is not None
             ):
-                preconditions["ifGenerationMatch"] = if_generation_match[index]
-
-            if (
-                if_metageneration_match is not None
-                and if_metageneration_match[index] is not None
-            ):
-                preconditions["ifMetagenerationMatch"] = if_metageneration_match[index]
+                preconditions["ifGenerationMatch"] = if_source_generation_match[index]
 
             if preconditions:
                 source_object["objectPreconditions"] = preconditions
@@ -3228,7 +3230,7 @@ class Blob(_PropertyMixin):
             activates it only if certain conditions are met. This class exists to provide safe defaults
             for RPC calls that are not technically safe to retry normally (due to potential data
             duplication or other side-effects) but become safe to retry if a condition such as
-            if_metageneration_match is set.
+            if_generation_match is set.
 
             See the retry.py source code and docstrings in this package (google.cloud.storage.retry) for
             information on retry types and how to configure them.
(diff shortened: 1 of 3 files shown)


google/cloud/storage/blob.py
```diff
@@ -3031,19 +3032,24 @@ def compose(
...
 
         :type if_metageneration_match: list of long
         :param if_metageneration_match:
             (Optional) Make the operation conditional on whether the blob's
             current metageneration matches the given value. The list must match
             ``sources`` item-to-item.
 
+        :type if_source_generation_match: list of long
+        :param if_source_generation_match:
+            (Optional) Make the operation conditional on whether the source blob's
```
tseaver, line 3049: Given that `sources` is expected to be plural, this should likely expect that here, e.g.:

```(python)
            (Optional) Make the operation conditional on whether the current generation
            of each source blob matches the corresponding generation.
            The list must match ``sources`` item-to-item.
```

google/cloud/storage/blob.py
```diff
@@ -3053,37 +3059,32 @@ def compose(
...
-            >>> if_generation_match = [None] * len(blobs)
-            >>> if_generation_match[0] = "123"  # precondition for "blob-name-1"
+            >>> if_source_generation_match = [None] * len(blobs)
+            >>> if_source_generation_match[0] = "123"  # precondition for "blob-name-1"
 
             >>> composed_blob = bucket.blob("composed-name")
-            >>> composed_blob.compose(blobs, if_generation_match)
+            >>> composed_blob.compose(blobs, if_source_generation_match=if_source_generation_match)
         """
         sources_len = len(sources)
```
tseaver, line 3081: It would seem clearer to map the default value for `if_source_generation_match` as a sequence of `None` values, e.g.:

```python
        if if_source_generation_match is None:
            if_source_generation_match = [None] * sources_len
```
and then do the check:
```python
        if len(if_source_generation_match) != sources_len:
            raise ValueError(...)
```
I would also move both statements down to where they are being used below.

google/cloud/storage/blob.py
```diff
@@ -3092,22 +3093,23 @@ def compose(
         if self.user_project is not None:
             query_params["userProject"] = self.user_project
 
+        _add_generation_match_parameters(
+            query_params,
+            if_generation_match=if_generation_match,
+            if_metageneration_match=if_metageneration_match,
+        )
+
```
tseaver, line 3101: Given the change proposed above, and that `index` is used only to find the corresponding entry in `if_source_generation_match`, this would be clearer using the `zip` built-in, e.g.:

```python
        for source, source_generation in zip(sources, if_source_generation_match):
```

google/cloud/storage/blob.py
```diff
@@ -3092,22 +3093,23 @@ def compose(
...
         source_objects = []
         for index, source in enumerate(sources):
             source_object = {"name": source.name}
+            source_object["generation"] = source.generation
```
tseaver, line 3105: Clearer to just use the literal:

```python
            source_object = {"name": source.name, "generation": source.generation}
```

google/cloud/storage/blob.py
```diff
@@ -3092,22 +3093,23 @@ def compose(
...
+                and if_source_generation_match[index] is not None
             ):
-                preconditions["ifGenerationMatch"] = if_generation_match[index]
-
-            if (
-                if_metageneration_match is not None
-                and if_metageneration_match[index] is not None
-            ):
-                preconditions["ifMetagenerationMatch"] = if_metageneration_match[index]
+                preconditions["ifGenerationMatch"] = if_source_generation_match[index]
```
tseaver, line 3112: This would then all simplify down to:

```python
            if source_generation is not None:
                source_object["objectPreconditions"] = {"ifGenerationMatch": source_generation}
```

tseaver (changes requested):

google/cloud/storage/blob.py
cojenco: Done, thanks @tseaver! This is much clearer. I like it a lot.

google/cloud/storage/blob.py
cojenco: Thanks @tseaver! That makes sense. I've made the changes and grouped them closer together. Though wouldn't we want to do the check and raise the exception early in the method? WDYT?

tseaver: @cojenco 
> Wouldn't we want to do the check and raise the exception early in the method?

I like the clarity of doing all the manipulations which depend on `if_source_generation_match` in one place:  in the odd case where the user passes an invalid sequence, we will have done only a trivial amount of work populating `query_params` before raising.

tseaver (approved):

tseaver:
@cojenco I will merge / repair the tests broken by my PRs this week.

tseaver:
@googlebot I consent.

tseaver:
@cojenco I put back the `do not merge` label to allow us to consider whether this might be a breaking change.

cojenco:
Got it, thanks @tseaver!

cojenco:
This currently will introduce a breaking change. Revising PR for backwards compatibility.

cojenco:
Updated PR to handle backwards compatibility. Ready for review:)

google/cloud/storage/blob.py
```diff
@@ -2979,18 +2979,36 @@ def compose(
...
         client = self._require_client(client)
         query_params = {}
 
-        if self.user_project is not None:
-            query_params["userProject"] = self.user_project
+        if isinstance(if_generation_match, list):
+            warnings.warn(
+                "if_generation_match: type list is deprecated and supported for backwards-compatability reasons only."
+                "Use if_source_generation_match instead to match source objects generations.",
+                PendingDeprecationWarning,
```
tseaver, line 2989: This can be a regular `DeprecationWarning` -- we are already deprecating the usage, not "fixin' to commence" to deprecated it.  Also, we want to use `stacklevel=2` to highlight the call site.

google/cloud/storage/blob.py
```diff
@@ -2934,35 +2934,35 @@ def compose(
...
+            (Optional) Makes the operation conditional on whether the
+            destination object's current generation matches the given value.
+            Setting to 0 makes the operation succeed only if there are no live
+            versions of the object.
+
+            If a list of long is passed in, makes the operation conditional on whether the
+            current generation of each source blob matches the corresponding generation.
+            The list must match ``sources`` item-to-item.
+            (Deprecated: type(list of long) is supported for backwards-compatability reasons only.
+            Please use if_source_generation_match instead to match source objects generations.)
```
tseaver, line 2946: Ouch!  PR #464 is also mangling the docstrings for the conditional parameters.

google/cloud/storage/blob.py
```diff
@@ -2979,18 +2979,36 @@ def compose(
...
+
+            # if_generation_match: type list is deprecated. Instead use if_source_generation_match.
+            if_source_generation_match = if_generation_match
+            if_generation_match = None
+
+        if isinstance(if_metageneration_match, list):
+            warnings.warn(
+                "if_metageneration_match matches the given value to the destination object's metageneration."
+                "Please pass in a single value (type long).",
+                PendingDeprecationWarning,
```
tseaver, line 3007: Likewise here.

tests/system/test_system.py
```diff
@@ -1736,17 +1736,22 @@ def test_compose_with_source_generation_match(self):
         self.case_blobs_to_delete.append(to_append)
 
         with self.assertRaises(google.api_core.exceptions.PreconditionFailed):
-            original.compose([original, to_append], if_source_generation_match=[6, 7])
+            original.compose(
+                [original, to_append],
+                if_generation_match=[6, 7],
+                if_metageneration_match=[8, 9],
+            )
```
tseaver: Do we need a system test for the deprecated case?

tseaver (changes requested):

google/cloud/storage/blob.py
cojenco: Done. Also revised the warning message based on #464. Thanks for catching this.

tests/system/test_system.py
```diff
@@ -1751,6 +1751,49 @@ def test_compose_with_generation_match(self):
         composed = original.download_as_bytes()
         self.assertEqual(composed, BEFORE + TO_APPEND)
```
cojenco, line 1752: @tseaver Re: Do we need a system test for the deprecated case?
`test_compose_with_generation_match_list` tests both deprecated cases that would fail or succeed. Here we test operation to succeed with a deprecated case in which list of long is passed in and source objects generations match. Are there any other cases missing?

google/cloud/storage/blob.py
cojenco: Thanks for catching this. I think this is a special case where we will need to have a full explanation in the docstrings considering deprecation and backwards compatibility. Let me know if there is a better way to handle the docstrings here.

tests/system/test_system.py
tseaver: I was actually questioning the need for systests for the deprecated case(s), rather than arguing for more. :)

google/cloud/storage/blob.py
tseaver: I think I'll hold off merging #464 until this lands, and then look at how to merge this change.

google/cloud/storage/blob.py
```diff
@@ -2931,73 +2932,99 @@ def compose(
...
         :param if_generation_match:
-            (Optional) Make the operation conditional on whether the blob's
-            current generation matches the given value.  Setting to 0 makes the
-            operation succeed only if there are no live versions of the blob.
+            (Optional) Makes the operation conditional on whether the
+            destination object's current generation matches the given value.
+            Setting to 0 makes the operation succeed only if there are no live
+            versions of the object.
+
+            If a list of long is passed in, makes the operation conditional on whether the
```
andrewsg, line 2942: I would recommend phrasing this something like:

"Note: In a previous version, this argument worked identically to the if_source_generation_match argument. For backwards-compatibility reasons, if a list is passed in, this argument will behave like if_source_generation_match and also issue a DeprecationWarning."

google/cloud/storage/blob.py
```diff
@@ -2931,73 +2932,99 @@ def compose(
...
+        :type if_metageneration_match: long
         :param if_metageneration_match:
-            (Optional) Make the operation conditional on whether the blob's
-            current metageneration matches the given value. The list must match
-            ``sources`` item-to-item.
+            (Optional) Makes the operation conditional on whether the
+            destination object's current metageneration matches the given
+            value.
+
+            If a list of long is passed in, no match operation will be performed.
```
andrewsg, line 2954: This didn't behave like if_source_metageneration_match before? It just did nothing?

google/cloud/storage/blob.py
```diff
@@ -2931,73 +2932,99 @@ def compose(
...
         sources_len = len(sources)
-        if if_generation_match is not None and len(if_generation_match) != sources_len:
-            raise ValueError(
-                "'if_generation_match' length must be the same as 'sources' length"
+        client = self._require_client(client)
+        query_params = {}
+
+        if isinstance(if_generation_match, list):
+            warnings.warn(
+                "if_generation_match: type list is deprecated and supported for backwards-compatability reasons only."
```
andrewsg, line 2987: This warning text sounds good to me.

google/cloud/storage/blob.py
```diff
@@ -2931,73 +2932,99 @@ def compose(
...
 
-        if (
-            if_metageneration_match is not None
-            and len(if_metageneration_match) != sources_len
-        ):
-            raise ValueError(
-                "'if_metageneration_match' length must be the same as 'sources' length"
+            if if_source_generation_match is not None:
+                raise ValueError(
+                    "Use if_generation_match to match the generation of the destination object."
```
andrewsg, line 2995: I'd recommend the first line read "Use if_generation_match to match the generation of the destination object by passing in a generation number, instead of a list."

andrewsg (changes requested):

google/cloud/storage/blob.py
cojenco: I don't think source objects metagenerations checks are performed here, according to the [API doc](https://cloud.google.com/storage/docs/json_api/v1/objects/compose#request-body).

google/cloud/storage/blob.py
cojenco: Done. I like this phrasing a lot, thanks!

tseaver (approved):

andrewsg (approved):
Thanks!

PR #458, opened 2021-06-08
fix: apply idempotency policies for ACLs

tseaver:
Toward #38.

- Use `None` as default retry for `Client._patch_resource` and `Client._put_resource` helpers (they can only be idempotent conditionally, so the caller needs to configure the retry).
- Remove `retry` argument from `ACL.save`, `ACL.save_predefined`, and `ACL.clear` (ACLs cannot be saved idempotently).
- Likewise remove `retry` argument from `Bucket.make_public`, `Bucket.make_private`, `Blob.make_public`, and `Blob.make_private`, which only perform saves on ACLs.

### DIFF
--- a/google/cloud/storage/acl.py
+++ b/google/cloud/storage/acl.py
@@ -474,9 +474,7 @@ class ACL(object):
         for entry in found.get("items", ()):
             self.add_entity(self.entity_from_dict(entry))
 
-    def _save(
-        self, acl, predefined, client, timeout=_DEFAULT_TIMEOUT, retry=DEFAULT_RETRY,
-    ):
+    def _save(self, acl, predefined, client, timeout=_DEFAULT_TIMEOUT):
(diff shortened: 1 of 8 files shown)


google/cloud/storage/acl.py
```diff
@@ -534,9 +532,7 @@ def _save(
 
         self.loaded = True
 
-    def save(
```
andrewsg, line 537: Was this change already included in a published release? If so, we *might* prefer to change retry to default to None rather than removing it, in the case that a customer is already using the new signature.

Likewise for all other public methods.

If it hasn't been included in a release then we're in the clear.

andrewsg (changes requested):

google/cloud/storage/acl.py
tseaver: No, it wasn't published (merged yesterday, in fact).

andrewsg (approved):
Thanks!

PR #459, opened 2021-06-08
tests: avoid requiring real credentials in unit tests

### LINKED ISSUE #416: Unit tests must not depend on environment variables
E.g.,:

```bash
$ env | grep GOOGLE || echo NO
NO
$ nox -re unit-3.9 -- -x
_______________ Test_Blob.test_download_as_byte_w_custom_timeout _______________

self = <tests.unit.test_blob.Test_Blob testMethod=test_download_as_byte_w_custom_timeout>

    def test_download_as_byte_w_custom_timeout(self):
>       self._download_as_bytes_helper(raw_download=False, timeout=9.58)

tests/unit/test_blob.py:1714: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/unit/test_blob.py:1557: in _download_as_bytes_helper
    client = self._make_client()
tests/unit/test_blob.py:59: in _make_client
    return Client(*args, **kw)
google/cloud/storage/client.py:122: in __init__
    super(Client, self).__init__(
.nox/unit-3-9/lib/python3.9/site-packages/google/cloud/client.py:277: in __init__
    _ClientProjectMixin.__init__(self, project=project, credentials=credentials)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <google.cloud.storage.client.Client object at 0x7fb9f31415b0>
project = None, credentials = None

    def __init__(self, project=None, credentials=None):
        # This test duplicates the one from `google.auth.default`, but earlier,
        # for backward compatibility:  we want the environment variable to
        # override any project set on the credentials.  See:
        # https://github.com/googleapis/python-cloud-core/issues/27
        if project is None:
            project = os.getenv(
                environment_vars.PROJECT,
                os.getenv(environment_vars.LEGACY_PROJECT),
            )
    
        # Project set on explicit credentials overrides discovery from
        # SDK / GAE / GCE.
        if project is None and credentials is not None:
            project = getattr(credentials, "project_id", None)
    
        if project is None:
            project = self._determine_default(project)
    
        if project is None:
>           raise EnvironmentError(
                "Project was not passed and could not be "
                "determined from the environment."
            )
E           OSError: Project was not passed and could not be determined from the environment.

.nox/unit-3-9/lib/python3.9/site-packages/google/cloud/client.py:228: OSError
```

tseaver:
Fixes #416

### DIFF
--- a/tests/unit/test_blob.py
+++ b/tests/unit/test_blob.py
@@ -57,7 +57,7 @@ class Test_Blob(unittest.TestCase):
     def _make_client(*args, **kw):
         from google.cloud.storage.client import Client
 
-        return Client(*args, **kw)
+        return mock.create_autospec(Client, instance=True, **kw)
 
     def test_ctor_wo_encryption_key(self):
         BLOB_NAME = "blob-name"
@@ -422,10 +422,10 @@ class Test_Blob(unittest.TestCase):
     def test_generate_signed_url_w_invalid_version(self):
         BLOB_NAME = "blob-name"
         EXPIRATION = "2014-10-16T20:34:37.000Z"
-        connection = _Connection()
-        client = _Client(connection)
+        client = self._make_client()
         bucket = _Bucket(client)
         blob = self._make_one(BLOB_NAME, bucket=bucket)
+
         with self.assertRaises(ValueError):
             blob.generate_signed_url(EXPIRATION, version="nonesuch")
 
@@ -464,8 +464,13 @@ class Test_Blob(unittest.TestCase):
         if expiration is None:
             expiration = datetime.datetime.utcnow().replace(tzinfo=UTC) + delta
 
-        connection = _Connection()
-        client = _Client(connection)
+        if credentials is None:
+            expected_creds = _make_credentials()
+            client = self._make_client(_credentials=expected_creds)
+        else:
+            expected_creds = credentials
+            client = self._make_client(_credentials=object())
+
         bucket = _Bucket(client)
         blob = self._make_one(blob_name, bucket=bucket, encryption_key=encryption_key)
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_bucket.py
```diff
@@ -3351,96 +3348,14 @@ def test_make_private_recursive_too_many(self):
 
         client.list_blobs.assert_called_once()
 
-    def test_page_empty_response(self):
```
chrisrossi, line 3354: Why are these tests being removed?

tseaver: They are non-unit tests (they really only test the already-well-tested `api_core.page_iterator.Iterator`).  The local bits are replaced by the (new) explicit tests of `_item_to_blob` and `_blobs_page_start` (at the top of the file).

tests/unit/test_blob.py
```diff
@@ -1024,6 +1021,99 @@ def _mock_requests_response(status_code, headers, content=b""):
         response.request = requests.Request("POST", "http://example.com").prepare()
         return response
 
+    def test__extract_headers_from_download_gzipped(self):
```
chrisrossi, line 1024: Adding coverage, too?

chrisrossi (approved):
LGTM!

tests/unit/test_blob.py
tseaver: The tests were misnamed, and out of place:  I just fixed that, and split a couple of over-large ones.

Issue #460, opened 2021-06-08
Move boilerplate docs for 'if_generation_match' et al. out of method docstrings


tseaver:
See also #455.

There are a ton of API-invoking methods which take `if_generation_match`, `if_generation_not_match`, etc. and have docstring entries for each of them which are both too verbose and too terse: they don't include / reference examples, and they are wordy. y preference, they would point instead to a narrative page (built from e.g. `docs/conditional_arguments.rst`) which would include comprehensive examples for them.

PR #461, opened 2021-06-08
docs: streamline 'timeout' / 'retry' docs in docstrings

### LINKED ISSUE #455: Move boilerplate docs for 'timeout' and 'retry' out of method docstrings
The repeted long-form docs for the `timeout` and `retry` arguments to all API-invoking methods are both too verbose and too terse:  they don't include / reference examples, and they are wordy.  By preference, they would point instead to a narrative page (built from e.g. `docs/timeout_retry.rst`) which would include comprehensive examples for them.

tseaver:
* Add 'requests' intersphinx refs.
* Add narrative docs for timeouts and retries.
* Replace boilerplate docs for 'timeout' and 'retry' in API-invoking methods with links to narrative docs.

Closes #455

### DIFF
--- a/docs/retry_timeout.rst
+++ b/docs/retry_timeout.rst
@@ -65,7 +65,7 @@ for each method, base on its semantics:
   retries any API request which returns a "transient" error.
 
 - For API requests which are idempotent only if the bucket or blob has
-  the same "metageneration", the library uses its
+  the same "generation", the library uses its
   :data:`~google.cloud.storage.retry.DEFAULT_RETRY_IF_GENERATION_SPECIFIED`
   policy, which retries API requests which returns a "transient" error,
   but only if the original request includes an ``ifGenerationMatch`` header.
@@ -81,6 +81,9 @@ for each method, base on its semantics:
   :data:`~google.cloud.storage.retry.DEFAULT_RETRY_IF_ETAG_IN_JSON`
   policy, which retries API requests which returns a "transient" error,
   but only if the original request includes an ``ETAG`` in its payload.
+
+- For those API requests which are never idempotent, the library passes
+  ``retry=None`` by default, suppressing any retries.
 
 Rather than using one of the default policies, you may choose to configure an
 explicit policy in your code.
--- a/docs/retry_timeout.rst
+++ b/docs/retry_timeout.rst
@@ -55,6 +55,11 @@ See also:
 Configuring Retries
 --------------------
 
+.. note::
+
+   For more background on retries, see also the
+   `GCS Retry Strategies Document <https://cloud.google.com/storage/docs/retry-strategy#python>`_ 
+
 Methods which invoke API methods may fail for a number of reasons, some of
 which represent "transient" conditions, and thus can be retried
 automatically.  The library tries to provide a sensible default retry policy
--- a/docs/retry_timeout.rst
+++ b/docs/retry_timeout.rst
@@ -1,0 +1,114 @@
+Configuring Timeouts and Retries
+================================
+
+When using object methods which invoke Google Cloud Storage API methods,
+you have several options for how the library handles timeouts and
+how it retries transient errors.
+
+
+.. _configuring_timeouts:
+
+Configuring Timeouts
+--------------------
+
+For a number of reasons, methods which invoke API methods may take
+longer than expected or desired.  By default, such methods all time out
+after a default interval, 60.0 seconds.  Rather than blocking your application
+code for that interval, you may choose to configure explicit timeouts
+in your code, using one of three forms:
+
+- You can pass a single integer or float which functions as the timeout for the
+  entire request. E.g.:
+
+.. code-block:: python
+
+   bucket = client.get_bucket(BUCKET_NAME, timeout=300.0)  # five minutes
+
+- You can also be passed as a two-tuple, ``(connect_timeout, read_timeout)``,
+  where the ``connect_timeout`` sets the maximum time required to establish
+  the connection to the server, and the ``read_timeout`` sets the maximum
+  time to wait for a completed response.  E.g.:
+
+.. code-block:: python
+
+   bucket = client.get_bucket(BUCKET_NAME, timeout=(3, 10))
+
+
+- You can also pass ``None`` as the timeout value:  in this case, the library
+  will block indefinitely for a response.  E.g.:
+
+.. code-block:: python
+
+   bucket = client.get_bucket(BUCKET_NAME, timeout=None)
+
+.. note::
+   Depending on the retry strategy, a request may be
+   repeated several times using the same timeout each time.
+
+See also:
+
+  :ref:`Timeouts in requests <requests:timeouts>`
+
+
+.. _configuring_retries:
+
+Configuring Retries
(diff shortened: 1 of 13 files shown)


owlbot.py
```diff
@@ -38,4 +38,17 @@
...
+intersphinx_mapping = {
+    "python": ("https://python.readthedocs.org/en/latest/", None),
+""",
+    """\
+intersphinx_mapping = {
+    "python": ("https://python.readthedocs.org/en/latest/", None),
+    "requests": ("https://docs.python-requests.org/en/master/", None),
+""",
+)
+
```
tseaver, line 53: Hmm, owlbot seems not to be applying this change:  036499972586c97ca79fc9796fbc56b1bbd510a3 removes the entry for `"requests"`.  @parthea, @busunkim96 can you help?

busunkim96: It looks like you can pass dict to `py_library` with additional intersphinx mappings:

https://github.com/googleapis/synthtool/blob/2430f8d90ed8a508e8422a3a7191e656d5a6bf53/synthtool/gcp/templates/python_library/docs/conf.py.j2#L370-L374

```py
common.py_library(
    ...,
    intersphinx_dependencies={
        "requests": "https://docs.python-requests.org/en/master/"
    }
)
```

Not sure why the current replacement is getting undone though.

tseaver: I needed to escape the parens in the text-to-be-replaced. :)

docs/retry_timeout.rst
```diff
@@ -0,0 +1,114 @@
...
+
+.. _configuring_retries:
+
+Configuring Retries
+--------------------
+
+Methods which invoke API methods may fail for a number of reasons, some of
+which represent "transient" conditions, and thus can be retried
+automatically.  The library tries to provide a sensible default retry policy
+for each method, base on its semantics.
```
tritone, line 61: I would add a little more color on the default retry policy at the top here. Generally, the strategy is to retry any idempotent operation (some calls are always idempotent while others require a precondition to qualify) on any transient error.

docs/retry_timeout.rst
```diff
@@ -0,0 +1,114 @@
...
+- You can pass an instance of :class:`google.api_core.retry.Retry` to enable
+  retries;  the passed object will define retriable response codes and errors,
+  as well as configuring backoff and retry interval options.  E.g.:
+
+.. code-block:: python
+
+   from google.api_core.retry import Retry
+
+   def is_retryable(exc):
+       ... # your retriable exception types here
```
tritone, line 81: Maybe add an example condition here for clarity.

tritone (commented):
Thanks for adding this! A couple small suggestions.

tritone:
@andrewsg @cojenco PTAL

docs/retry_timeout.rst
tseaver: 4278187

docs/retry_timeout.rst
tseaver: 4278187

docs/retry_timeout.rst
```diff
@@ -58,9 +58,31 @@ Configuring Retries
...
-
-Rather than using the default policy, you may choose to configure an
+for each method, base on its semantics:
+
+- For API requests which are always idempotent, the library uses its
+  :data:`~google.cloud.storage.retry.DEFAULT_RETRY` policy, which
+  retries any API request which returns a "transient" error.
+
+- For API requests which are idempotent only if the bucket or blob has
+  the same "metageneration", the library uses its
```
tritone, line 68: Should be "generation" here

docs/retry_timeout.rst
```diff
@@ -58,9 +58,31 @@ Configuring Retries
...
+  :data:`~google.cloud.storage.retry.DEFAULT_RETRY_IF_METAGENERATION_SPECIFIED`
+  policy, which retries API requests which returns a "transient" error,
+  but only if the original request includes an ``ifMetagenerationMatch`` header.
+
+- For API requests which are idempotent only if the bucket or blob has
+  the same "etag", the library uses its
+  :data:`~google.cloud.storage.retry.DEFAULT_RETRY_IF_ETAG_IN_JSON`
+  policy, which retries API requests which returns a "transient" error,
+  but only if the original request includes an ``ETAG`` in its payload.
+
```
tritone, line 84: There are also a few APIs which are always non-idempotent; these are never retried.

tritone (commented):
Looks great!

docs/retry_timeout.rst
tseaver: 2b071b0

docs/retry_timeout.rst
tseaver: 2b071b0

docs/retry_timeout.rst
```diff
@@ -0,0 +1,147 @@
...
+Methods which invoke API methods may fail for a number of reasons, some of
+which represent "transient" conditions, and thus can be retried
+automatically.  The library tries to provide a sensible default retry policy
+for each method, base on its semantics:
+
+- For API requests which are always idempotent, the library uses its
+  :data:`~google.cloud.storage.retry.DEFAULT_RETRY` policy, which
+  retries any API request which returns a "transient" error.
+
+- For API requests which are idempotent only if the bucket or blob has
```
cojenco, line 67: We can remove `bucket` here. Since buckets have no payload data, [buckets have no generation numbers](https://cloud.google.com/storage/docs/generations-preconditions).  `if_generation_match` should only apply to blobs.

docs/retry_timeout.rst
```diff
@@ -0,0 +1,147 @@
...
+  the same "generation", the library uses its
```
cojenco, line 68: Since we are streamlining the retry strategy reference to one place, should we include here a descriptive use case: `Setting to 0 makes the operation succeed only if there are no live versions of the blob.`

tseaver: That language would actually be documentation for the `if_generation_match` argument, rather than the retry policy:  I've proposed doing a similar streamline for those arguments in #460.

docs/retry_timeout.rst
tseaver: 4bb0001

docs/retry_timeout.rst
```diff
@@ -0,0 +1,147 @@
...
+
+   bucket = client.get_bucket(BUCKET_NAME, timeout=None)
+
+.. note::
+   Depending on the retry strategy, a request may be
+   repeated several times using the same timeout each time.
+
+See also:
+
+  :ref:`Timeouts in requests <requests:timeouts>`
```
cojenco, line 50: Looks great, thanks! @andrewsg and our TW, Karen, also wrote [documentation on Retry Strategies specifically for the Python client library](https://cloud.google.com/storage/docs/retry-strategy#python). Maybe we could link it as a reference too.

cojenco (approved):
LGTM, thanks @tseaver for streamlining this!

docs/retry_timeout.rst
tseaver: 1a56e89

PR #467, opened 2021-06-14
tests: pre-scrub old HMAC keys before testing creation

### LINKED ISSUE #334: 'test_hmac_key_crud` test flakes
See [this build failure](https://source.cloud.google.com/results/invocations/c7fde57b-8643-49c3-bca8-37f5016e3f2d/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-storage%2Fpresubmit%2Fpresubmit/log).

```python
=================================== FAILURES ===================================
________________________ TestClient.test_hmac_key_crud _________________________

self = <test_system.TestClient testMethod=test_hmac_key_crud>

    def test_hmac_key_crud(self):
        from google.cloud.storage.hmac_key import HMACKeyMetadata

        credentials = Config.CLIENT._credentials
        email = credentials.service_account_email

        before_keys = set(Config.CLIENT.list_hmac_keys())

>       metadata, secret = Config.CLIENT.create_hmac_key(email)

tests/system/test_system.py:136:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
google/cloud/storage/client.py:840: in create_hmac_key
    retry=None,
google/cloud/storage/_http.py:63: in api_request
    return call()
...
        response = self._make_request(
            method=method,
            url=url,
            data=data,
            content_type=content_type,
            headers=headers,
            target_object=_target_object,
            timeout=timeout,
        )

        if not 200 <= response.status_code < 300:
>           raise exceptions.from_http_response(response)
E           BadRequest: 400 POST https://storage.googleapis.com/storage/v1/projects/precise-truck-742/hmacKeys?serviceAccountEmail=kokoro%40precise-truck-742.iam.gserviceaccount.com&prettyPrint=false: Service account HMAC key limit reached
```

This seems likely triggered by simultaneous CI runs:  I can't remember how to check if we have leaked HMAC keys in the console.  @frankyn, can you check if there is a quota we could raise?

tseaver:
Avoids hitting 5-key-per-service-account quota.

Closes #334.

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -27,6 +27,7 @@ import unittest
 import mock
 
 import requests
+import pytest
 import six
 
 from google.cloud import exceptions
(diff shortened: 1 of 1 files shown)


tseaver:
Note to reviewers:  I created a key manually and, inside the testcase, hacked its `time_created` to be older than a day, just to ensure that the pre-scrub worked as expected.

tests/system/test_system.py
```diff
@@ -149,6 +150,18 @@ def test_hmac_key_crud(self):
 
         before_keys = set(Config.CLIENT.list_hmac_keys())
 
+        now = datetime.datetime.utcnow().replace(tzinfo=_helpers.UTC)
```
chrisrossi, line 153: With `pytz` in the standard library now, which includes a `utc`, is there any need to still use our own handrolled?

chrisrossi (approved):
LGTM. This feels like a think that should be a fixture, to me, but if it really only happens once, I guess it can go either way.

tests/system/test_system.py
tseaver:
```bash
$ python3.6
Python 3.6.10 (default, Apr 20 2021, 13:17:17) 
[GCC 9.3.0] on linux
>>> import pytz
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'pytz'
>>>
```
See https://github.com/googleapis/python-api-core/commit/61490f5e9c6f3ba4729e35cd1fc8435c82a43041 (part of https://github.com/googleapis/python-api-core/pull/212), which uses `datetime.timezone.utc`.

GCS is still stuck supporting Python 2.7, however.

tseaver:
@chrisrossi c07995d moves the pre-scrub logic to a helper method (these systests aren't pytest-fixtures-ified), so that any future systest which might create HMAC keys can reuse it.

PR #480, opened 2021-06-25
fix: replace default retry for upload operations

### LINKED ISSUE #477: Uploads and if_metageneration_match and retries
Sorry for the ticket here, I've also put in a support case but I'm guessing you all have more domain expertise than them and can quickly yes/no this one.

When doing an upload of a new blob that shouldn't exist already, I would like retries to actually work. But the code makes it pretty clear that I should set an if_metageneration_match precondition if I want retries on uploads. So I have this code:

```
client = storage.Client()
bucket = client.bucket("whatever")
blob = client.blob("a_file_that_doesnt_yet_exist.txt")
blob.upload_from_string("Hello world", if_generation_match=0, if_metageneration_match=0)
```

But that fails all requests. I _think_ the right answer is just that the default retry policy is incorrect for fresh uploads and that if_generation_match=0 actually provides all the idemopotence safety I need? So the fix would be to use 1.39 and the new `retry=storage.retry.DEFAULT_RETRY` (or `storage.retry.DEFAULT_RETRY_IF_GENERATION_SPECIFIED` but that's functionally the same if I'm hardwiring in if_generation_match myself in the same call)? Maybe the default retry policy on uploads should be tweaked to "either if_metageneration_match is set at all or if_generation_match==0"? Thanks in advance!

cojenco:
This PR replaces upload operations default retries with `storage.retry.DEFAULT_RETRY_IF_GENERATION_SPECIFIED` and aligns with the [retry strategy conditional idempotency](https://cloud.google.com/storage/docs/retry-strategy#idempotency-operations). 

With default retries  set to `storage.retry.DEFAULT_RETRY_IF_GENERATION_SPECIFIED`, passing in `ifGenerationMatch` makes the upload operation (1) conditional and (2) only retried by default on whether the object's current generation matches the given value. Setting `if_generation_match=0` makes the upload operation succeed only if there are no live versions of the object (fresh uploads) and supports retries. 

- revises upload operations both in entry methods and private methods
- updates tests
- revises docstrings
- (update: separating this to #484 ) adds preconditions and retry configuration to an upload entry method `blob.create_resumable_upload_session()`

Fixes #477 🦕


google/cloud/storage/blob.py
```diff
@@ -1750,7 +1749,7 @@ def _do_multipart_upload(
 
             This private method does not accept ConditionalRetryPolicy values
             because the information necessary to evaluate the policy is instead
-            evaluated in client.download_blob_to_file().
+            evaluated in blob._do_upload().
```
andrewsg, line 1752: Oh, it looks like we're evaluating it BOTH in blob._do_upload() and in client.download_blob_to_file(). This is harmless, which is why the tests didn't catch it, but we only need one of those.

I don't have a strong opinion on whether we do this in _do_upload or download_blob_to_file. Let's keep this docstring change and additionally we can remove the conditional retry policy code in client.download_blob_to_file, either in another PR, or in this one if you prefer.

andrewsg (approved):
Good work here. Please amend the PR title to include that create_resumable_upload_session is being amended in the same PR (or split it off into a separate PR) so that the change is included in the release notes.

google/cloud/storage/blob.py
```diff
@@ -1877,10 +1876,10 @@ def _initiate_resumable_upload(
         :type num_retries: int
         :param num_retries:
             Number of upload retries. By default, only uploads with
```
tritone, line 1878: This docstring is kind of misleading because the default behavior for this method is no retry, but I guess it's not really worth messing with here.

google/cloud/storage/blob.py
```diff
@@ -2783,6 +2782,11 @@ def create_resumable_upload_session(
         client=None,
         timeout=_DEFAULT_TIMEOUT,
         checksum=None,
+        if_generation_match=None,
```
tritone, line 2785: Adding these parameters here seems kind of separate from the rest of these changes. Is this method called under the hood by other retry operations, or just directly by external callers?

tests/unit/test_blob.py
```diff
@@ -3251,8 +3248,18 @@ def test_upload_from_string_w_text_w_num_retries(self):
...
+        origin=None,
+        side_effect=None,
+        timeout=None,
+        if_generation_match=None,
+        if_generation_not_match=None,
+        if_metageneration_match=None,
+        if_metageneration_not_match=None,
+        retry=None,
     ):
+        from six.moves.urllib.parse import urlencode
```
tritone, line 3261: Why does this have to be imported inside the test?

tritone (commented):
Nice catch on this! A couple questions

google/cloud/storage/blob.py
cojenco: IIUC, I believe we'll need to remain both ConditionalRetryPolicy evaluations as one is for **upload** operations and the other is for **downloads**. I revised the docstrings here to match particularly where we're doing the evaluations for upload operations, `blob._do_upload()` .

google/cloud/storage/blob.py
cojenco: Yep that is more for backwards-compatibility. This is a private method that is supposedly to be called by such entry methods as `upload_from_file()` , and the `retry` config and `num_retries` compatible logic is currently handled in `_do_upload()` before calling `_initiate_resumable_upload()`

google/cloud/storage/blob.py
cojenco: `create_resumable_upload_session()` is called directly by external callers. This is to add preconditions and retry configuration. I think it may make more sense to separate this out as an independent PR. Thanks @tritone and @andrewsg for the input!

cojenco: Done, thanks! Split into PR #484

tritone (approved):

PR #443, opened 2021-05-11
refactor: add / use 'Client._post_resource' method

tseaver:
Toward #38.

~~Note: this PR is based on 6244df0 of PR #442.  I will rebase to master when that PR lands.~~ Done.

PR #441, opened 2021-05-10
refactor:  add / use 'Client._put_resource' method

tseaver:
Toward #38.

~~Note: this PR is based on efd5a14 from PR #436. I will rebase to master when that PR lands.~~ Done.

PR #436, opened 2021-05-07
refactor:  add / use 'Client._patch_resource' method

tseaver:
Also, add retry support to `ACL.save`, `ACL.save_predefined`, and `ACL.clear`.

Toward #38.

~~Note: this PR is based on 44eeeb4 from PR #431.  I will rebase to master when that PR lands.~~ Done.

PR #483, opened 2021-06-28
fix: revise and rename is_etag_in_json(data)

### LINKED ISSUE #481: Handle data dict type in retry.is_etag_in_json(data)
`retry.is_etag_in_json(data)` is a helper method for conditional retries. It returns a boolean whether or not an `etag` is contained in the JSON body. 

There seem to be use cases in which a dictionary type of data is passed into [`retry.is_etag_in_json(data)`](https://github.com/googleapis/python-storage/blob/cb2a89b0b696defd56c4c1eee5a1cad67ae4cb43/google/cloud/storage/retry.py#L128). Currently the method is unable to handle dict type data correctly: a `JSONDecodeError` is caught before examining if an etag is contained within.

cojenco:
This PR revises and renames `retry.is_etag_in_json(data)` to ensure that eTags are correctly captured and validated from the request body. 

- Rename conditional retry predicate `retry.is_etag_in_json(data)`  to `retry.is_etag_in_data(data)`. It returns a boolean whether or not an `etag` is contained in the request body. 
- Revise `retry.is_etag_in_data(data)` to reflect library use cases in which `data` is `dict` or `None` 
- Update tests and docs

Fixes #481 🦕


google/cloud/storage/retry.py
```diff
@@ -125,6 +125,8 @@ def is_etag_in_json(data):
 
     Indended for use on calls with relatively short JSON payloads."""
     try:
+        if isinstance(data, dict) and data.get("etag"):
```
tritone, line 128: It's a bit disturbing that this can be called with both dict and string data. Is there a way to fix the call sites so that we are passing in a consistent kind of argument to this function, rather than accepting both?

tseaver: If we *do* have this method take both a JSON-encoded string and a `dict`, then we should factor it such that the JSON-encoded string is decoded to a `dict` *first*, leaving the test the same. Also, we should update the docstring to indicate that both types are allowed:

```python
def is_etag_in_json(data):
    """Return True if an etag is contained in the JSON body.

    :type data: str or dict
    :param data: A string containing a JSON-encoded dict, or the dict itsefl.

    Intended for use on calls with relatively short JSON payloads."""
    if isinstance(data, six.text_type):
        try:
            data = json.loads(data)
        # Though this method should only be called when a JSON body is expected,
        # the retry policy should be robust to unexpected payloads.
        # In Python 3 a JSONDecodeError is possible, but it is a subclass of ValueError.
        except (ValueError, TypeError):
            return False
    return "etag" in data
```

tseaver: The change in 73a9802 LGTM, assuming @tritone's concern above is addressed.

cojenco: Thanks @tseaver @tritone for the reviews. I understand it can be quite confusing. Given that retries are configurable, there are cases where `data` is type string or type dict. In our discussion, @andrewsg pointed out many cases `data` is a string containing a JSON-encoded dict. Meanwhile, the callers of this function in our library pass `data` as a dict, for example:
- [`client._put_resource`](https://github.com/googleapis/python-storage/blob/master/google/cloud/storage/client.py#L490) where `data` is type dict (called by `bucket.set_iam_policy` , `hmac_key.update`, `bucket.update`, `blob.update`)
- [`client._patch_resource`](https://github.com/googleapis/python-storage/blob/1ae766ed3e1313439c5a8698562c02d1a0344ff3/google/cloud/storage/client.py#L419) where `data` is type dict (called by `bucket.patch`, `blob.patch`)

Another option would be modifying all the `data` types in the above-mentioned library methods?

tseaver: @cojenco 

>  In our discussion, @andrewsg pointed out many cases data is a string containing a JSON-encoded dict. 

I'm not sure where `data` is going to be a string in cases where `if_etag_in_json` is called.

A quick survey of usage:

- Outside of tests, `is_etag_in_json` is used only as the predicate for the `DEFAULT_RETRY_IF_ETAG_IN_JSON` retry policy:

```bash
$ git grep -B 1 is_etag_in_json google/
google/cloud/storage/retry.py-
google/cloud/storage/retry.py:def is_etag_in_json(data):
--
google/cloud/storage/retry.py-DEFAULT_RETRY_IF_ETAG_IN_JSON = ConditionalRetryPolicy(
google/cloud/storage/retry.py:    DEFAULT_RETRY, is_etag_in_json, ["data"]
```
The `DEFAULT_RETRY_IF_ETAG_IN_JSON` policy is used in exactly three places:

```bash
$ git grep DEFAULT_RETRY_IF_ETAG_IN_JSON google/ | grep -v import
google/cloud/storage/blob.py:        retry=DEFAULT_RETRY_IF_ETAG_IN_JSON,
google/cloud/storage/bucket.py:        retry=DEFAULT_RETRY_IF_ETAG_IN_JSON,
google/cloud/storage/hmac_key.py:    def update(self, timeout=_DEFAULT_TIMEOUT, retry=DEFAULT_RETRY_IF_ETAG_IN_JSON):
google/cloud/storage/retry.py:DEFAULT_RETRY_IF_ETAG_IN_JSON = ConditionalRetryPolicy(
```

which turn out to be:

- `Blob.set_iam_policy` (passes a `dict` to `client._put_resource`)
- `Bucket.set_iam_policy` (passes a `dict` to `client._put_resource`)
- `HMACKeyMetadata.update` (passes a `dict` to `client._put_resource`)

Although `Bucket.patch`, `Blob.patch`, `Bucket.update`, and `Blob.update` do not use the `DEFAULT_RETRY_IF_ETAG_IN_JSON` policy, they all pass `dict` to `client._patch_resource`.

We do expose `DEFAULT_RETRY_IF_ETAG_IN_JSON` as a public symbol in the public `retry` module, so I guess there could be cases out in the wild which use it for some other API calls, but I'm doubtful any such hypothetical users are actually getting real retries done.

tseaver: FWIW, I just tweaked the implementation of `is_etag_in_json` as follows:

```python
def is_etag_in_json(data):
    """Return True if an etag is contained in the JSON body.

    Indended for use on calls with relatively short JSON payloads."""
    return data.get("etag") is not None
```

and all systests pass.

tseaver: Digging further:  since PRs #436, #441, and #443,  `Client._patch_resource`, `Client._put_resource`, and `Client._post_resource` are the only three places where the `Connection.api_request` method is called with a `data` argument.  Therefore their callers are the only ones which could potentially use the `DEFAULT_RETRY_IF_ETAG_IN_JSON` policy.

There are a few cases where `Client._post_resource` is called with `data` passed as `None`:  otherwise, `data` is always a `dict` instance.  So, I think we could simplify `is_etag_in_json` to just:

```python
    return data is not None and "etag" in data
```

The various upload mechanisms might still end up passing in `data` as bytes at a lower level, but retrying based on an `ETag` is never sensible for them:  they all create new versions of the blob, and `ETag` is not a writable property.

cojenco: Thanks for digging in and clarifying the issue. That makes lots of sense. Knowing that the library callers are passing in data as `dict` or `None`, I'm not worried about revising the predicate. My (no longer) biggest concern was `DEFAULT_RETRY_IF_ETAG_IN_JSON` exposed and configurable in the public retry module. I like the idea of simplifying the predicate. Updating the changes, thanks!

google/cloud/storage/retry.py
```diff
@@ -123,17 +121,11 @@ def is_metageneration_specified(query_params):
...
-        if content.get("etag"):
-            return True
-    # Though this method should only be called when a JSON body is expected,
-    # the retry policy should be robust to unexpected payloads.
-    # In Python 3 a JSONDecodeError is possible, but it is a subclass of ValueError.
-    except (ValueError, TypeError):
-        pass
-    return False
+    :type data: dict or None
+    :param data: A dict containing the JSON body. If not passed, returns False.
```
tritone, line 125: Nit: I think we could be clearer that the dict _represents_ the JSON body rather than containing it. Also, this is specifically intended for calls where metadata is sent in the request body.

It seems like it'd be better to actually just rename this function to `is_etag_in_data` but I guess technically this is part of the public surface of the library (even if we don't expect external callers to use it directly)?

tritone (approved):
Looks much cleaner, thank you!

google/cloud/storage/retry.py
```diff
@@ -122,7 +122,7 @@ def is_etag_in_json(data):
     """Return True if an etag is contained in the JSON body.
 
     :type data: dict or None
-    :param data: A dict containing the JSON body. If not passed, returns False.
+    :param data: A dict representing the JSON body. If not passed, returns False.
 
     Intended for use on calls with relatively short JSON payloads."""
```
tseaver: While we're nitpicking, the last line is now irrelevant:  we won't be loading a JSON string, because the caller already has it as a dict.

google/cloud/storage/retry.py
tseaver: We could rename it and leave `is_etag_in_json` around as a backward-compatibility alias.

cojenco: Would it be something like this:

```
def is_etag_in_data(data):
    """Return True if an etag is contained in the request body.

    :type data: dict or None
    :param data: A dict representing the request JSON body. If not passed, returns False.
    """
    return data is not None and "etag" in data


def is_etag_in_json(data):
    """
    ``is_etag_in_json`` is supported for backwards-compatibility reasons only;
    please use ``is_etag_in_data`` instead.
    """
    return is_etag_in_data(data)
```

tseaver: LGTM

tritone (approved):

PR #484, opened 2021-06-29
feat: add preconditions and retry configuration to blob.create_resumable_upload_session

cojenco:
This adds preconditions and retry configuration to the upload entry method `blob.create_resumable_upload_session()`, in the continuation of #447 and #480.

This exposes the method to preconditions and therefore can make the operation conditional. It can also be configured with Retry objects and ConditionalRetryPolicy objects.

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -2783,6 +2783,11 @@ class Blob(_PropertyMixin):
         client=None,
         timeout=_DEFAULT_TIMEOUT,
         checksum=None,
+        if_generation_match=None,
+        if_generation_not_match=None,
+        if_metageneration_match=None,
+        if_metageneration_not_match=None,
+        retry=DEFAULT_RETRY_IF_GENERATION_SPECIFIED,
     ):
         """Create a resumable upload session.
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_blob.py
```diff
@@ -3321,6 +3345,26 @@ def test_create_resumable_upload_session_with_custom_timeout(self):
...
+
+    def test_create_resumable_upload_session_with_conditional_retry_success(self):
+        self._create_resumable_upload_session_helper(
+            retry=DEFAULT_RETRY_IF_GENERATION_SPECIFIED, if_generation_match=123456
+        )
+
+    def test_create_resumable_upload_session_with_conditional_retry_failure(self):
+        self._create_resumable_upload_session_helper(
+            retry=DEFAULT_RETRY_IF_GENERATION_SPECIFIED
+        )
```
tritone, line 3366: How does this test verify that retries don't happen?

google/cloud/storage/blob.py
```diff
@@ -2858,6 +2863,41 @@ def create_resumable_upload_session(
...
+            A google.cloud.storage.retry.ConditionalRetryPolicy value wraps a
+            Retry object and activates it only if certain conditions are met.
+            This class exists to provide safe defaults for RPC calls that are
+            not technically safe to retry normally (due to potential data
+            duplication or other side-effects) but become safe to retry if a
+            condition such as if_generation_match is set.
+            See the retry.py source code and docstrings in this package
+            (google.cloud.storage.retry) for information on retry types and how
+            to configure them.
+            Media operations (downloads and uploads) do not support non-default
```
tritone, line 2896: This is a bit confusingly worded, but I guess it is boilerplate for other media ops?

tritone (commented):
Generally looks good. Did you test against the emulator that things are working?

tests/unit/test_blob.py
cojenco: @tritone, I've tested the retry behaviors against the emulator that things are working as intended. The unit test asserts if the mock call was made with the corresponding url and query params.

tritone: Sounds good!

tritone (approved):

PR #488, opened 2021-07-06
fix: populate etag / generation / metageneration properties during download

### LINKED ISSUE #490: Downloads do not populate properties from all headers
See https://github.com/googleapis/python-storage/issues/451#issuecomment-871463430

The following headers should be used to populate the corresponding properties:

| HTTP Header  | Property |
| ------------- | ------------- |
| `ETag`  | `etag`  |
| `x-goog-generation`  | `generation`  |
| `x-goog-metageneration`  | `metageneration`  |

daniellehanks:
Populates additional properties (`etag`, `generation`, `metageneration`) from the corresponding headers (`ETag`, `x-goog-generation`, and `x-goog-metageneration`) during download functions via `Blob._extract_headers_from_download`.

Additionally, update documentation for running system tests to use Python 3.8 instead of Python 3.7 to align with nox.

Closes #490.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -133,13 +133,13 @@ Running System Tests
 
 - To run system tests, you can execute::
 
-   $ nox -s system-3.7
+   $ nox -s system-3.8
    $ nox -s system-2.7
 
   .. note::
 
       System tests are only configured to run under Python 2.7 and
-      Python 3.7. For expediency, we do not run them in older versions
+      Python 3.8. For expediency, we do not run them in older versions
       of Python 3.
 
   This alone will not run the tests. You'll need to change some local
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.rst
```diff
@@ -133,13 +133,13 @@ Running System Tests
 
 - To run system tests, you can execute::
 
-   $ nox -s system-3.7
+   $ nox -s system-3.8
    $ nox -s system-2.7
```
tseaver, line 137: Please back out changes to this file -- it is generated.

tseaver (changes requested):
Thanks very much for the patch!

CONTRIBUTING.rst
tseaver: Please disregard:  this file is *excluded* from generation, due to Python 2.7 support (see #412).

tseaver (approved):

PR #489, opened 2021-07-06
feat: add support for Etag headers on reads

### LINKED ISSUE #451: Support ETag headers
Split from [this issue](https://github.com/googleapis/python-storage/issues/30#issuecomment-800003630).

Requesting support for `If-Match`/`If-None-Match` headers for GCS via the Python client library.

My use case:
My API is a thin layer over GCS (for storing profile images), and I only want to actually read the image data and serve it to the client if the actual content is different from their cache. They will specify an `If-None-Match` header to my API (as they won't know anything about GCS or generation numbers, kind of the point), so I just need to essentially pass this along to GCS. I don't want to either do the extra round tripping of reading the current file to check the ETag and then use the generation number from that in a subsequent request, or not use the client library.

daniellehanks:
Support conditional requests based on ETag for read operations (`reload`, `exists`, `download_*`). My own testing seems to indicate that the JSON API does not support ETag If-Match/If-None-Match headers on modify requests (`patch`, `delete`, etc.), please correct me if I am mistaken.

This part two of #451. Part one in https://github.com/googleapis/python-storage/pull/488.
Fixes #451 🦕

### DIFF
--- a/docs/generation_metageneration.rst
+++ b/docs/generation_metageneration.rst
@@ -17,7 +17,10 @@ Concepts
 ETag
 ::::::::::::::
 
-An ETag is returned as part of the response header whenever a resource is returned, as well as included in the resource itself. Users should make no assumptions about the value used in an ETag except that it changes whenever the underlying data changes, per the
+An ETag is returned as part of the response header whenever a resource is
+returned, as well as included in the resource itself. Users should make no
+assumptions about the value used in an ETag except that it changes whenever the
+underlying data changes, per the
 `specification <https://tools.ietf.org/html/rfc7232#section-2.3>`_
 
 The ``ETag`` attribute is set by the GCS back-end, and is read-only in the
--- a/docs/generation_metageneration.rst
+++ b/docs/generation_metageneration.rst
@@ -1,16 +1,27 @@
-Conditional Requests Via Generation / Metageneration Preconditions
-==================================================================
+Conditional Requests Via ETag / Generation / Metageneration Preconditions
+=========================================================================
 
 Preconditions tell Cloud Storage to only perform a request if the
-:ref:`generation <concept-generation>` or
+:ref:`ETag <concept-etag>`, :ref:`generation <concept-generation>`, or
 :ref:`metageneration <concept-metageneration>` number of the affected object
-meets your precondition criteria. These checks of the generation and
+meets your precondition criteria. These checks of the ETag, generation, and
 metageneration numbers ensure that the object is in the expected state,
 allowing you to perform safe read-modify-write updates and conditional
 operations on objects
 
 Concepts
 --------
+
+.. _concept-etag:
+
+ETag
+::::::::::::::
+
+An ETag is returned as part of the response header whenever a resource is returned, as well as included in the resource itself. Users should make no assumptions about the value used in an ETag except that it changes whenever the underlying data changes, per the
+`specification <https://tools.ietf.org/html/rfc7232#section-2.3>`_
+
+The ``ETag`` attribute is set by the GCS back-end, and is read-only in the
+client library.
 
 .. _concept-metageneration:
 
@@ -58,6 +69,32 @@ See also
 
 Conditional Parameters
 ----------------------
+
+.. _using-if-etag-match:
+
+Using ``if_etag_match``
+:::::::::::::::::::::::::::::
+
+Passing the ``if_etag_match`` parameter to a method which retrieves a
+blob resource (e.g.,
+:meth:`Blob.reload <google.cloud.storage.blob.Blob.reload>`)
+makes the operation conditional on whether the blob's current ``ETag`` matches
+the given value. This parameter is not supported for modification (e.g.,
+:meth:`Blob.update <google.cloud.storage.blob.Blob.update>`).
+
+
+.. _using-if-etag-not-match:
+
+Using ``if_etag_not_match``
+:::::::::::::::::::::::::::::
+
+Passing the ``if_etag_not_match`` parameter to a method which retrieves a
+blob resource (e.g.,
+:meth:`Blob.reload <google.cloud.storage.blob.Blob.reload>`)
+makes the operation conditional on whether the blob's current ``ETag`` matches
+the given value. This parameter is not supported for modification (e.g.,
+:meth:`Blob.update <google.cloud.storage.blob.Blob.update>`).
+
 
 .. _using-if-generation-match:
 
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -34,6 +34,12 @@ STORAGE_EMULATOR_ENV_VAR = "STORAGE_EMULATOR_HOST"
 
 _DEFAULT_STORAGE_HOST = u"https://storage.googleapis.com"
 
+# etag match parameters in snake case and equivalent header
+_ETAG_MATCH_PARAMETERS = (
+    ("if_etag_match", "If-Match"),
+    ("if_etag_not_match", "If-None-Match"),
+)
+
 # generation match parameters in camel and snake cases
 _GENERATION_MATCH_PARAMETERS = (
     ("if_generation_match", "ifGenerationMatch"),
@@ -384,8 +390,7 @@ class _PropertyMixin(object):
 
 
 def _scalar_property(fieldname):
-    """Create a property descriptor around the :class:`_PropertyMixin` helpers.
-    """
+    """Create a property descriptor around the :class:`_PropertyMixin` helpers."""
 
     def _getter(self):
         """Scalar property getter."""
@@ -447,6 +452,29 @@ def _convert_to_timestamp(value):
     utc_naive = value.replace(tzinfo=None) - value.utcoffset()
     mtime = (utc_naive - datetime(1970, 1, 1)).total_seconds()
     return mtime
+
+
+def _add_etag_match_headers(headers, **match_parameters):
+    """Add generation match parameters into the given parameters list.
+
+    :type headers: dict
+    :param headers: Headers dict.
+
+    :type match_parameters: dict
+    :param match_parameters: if*etag*match parameters to add.
+
+    :raises: :exc:`ValueError` if ``parameters`` is not a ``dict()``.
+    """
+    if not isinstance(headers, dict):
+        raise ValueError("`headers` argument should be a dict() or a list().")
+
+    for snakecase_name, header_name in _ETAG_MATCH_PARAMETERS:
+        value = match_parameters.get(snakecase_name)
+
+        if value is not None:
+            if isinstance(value, str):
+                value = [value]
+            headers[header_name] = ", ".join(value)
 
 
 def _add_generation_match_parameters(parameters, **match_parameters):
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -153,6 +153,8 @@ class _PropertyMixin(object):
         self,
         client=None,
         projection="noAcl",
+        if_etag_match=None,
+        if_etag_not_match=None,
         if_generation_match=None,
         if_generation_not_match=None,
         if_metageneration_match=None,
@@ -173,6 +175,12 @@ class _PropertyMixin(object):
         :param projection: (Optional) If used, must be 'full' or 'noAcl'.
                            Defaults to ``'noAcl'``. Specifies the set of
                            properties to return.
+
+        :type if_etag_match: Union[str, Set[str]]
+        :param if_etag_match: (Optional) See :ref:`using-if-etag-match`
+
+        :type if_etag_not_match: Union[str, Set[str]])
+        :param if_etag_not_match: (Optional) See :ref:`using-if-etag-not-match`
 
         :type if_generation_match: long
         :param if_generation_match:
@@ -211,10 +219,14 @@ class _PropertyMixin(object):
             if_metageneration_match=if_metageneration_match,
             if_metageneration_not_match=if_metageneration_not_match,
         )
+        headers = self._encryption_headers()
+        _add_etag_match_headers(
+            headers, if_etag_match=if_etag_match, if_etag_not_match=if_etag_not_match
+        )
         api_response = client._get_resource(
             self.path,
             query_params=query_params,
-            headers=self._encryption_headers(),
+            headers=headers,
             timeout=timeout,
             retry=retry,
             _target_object=self,
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -474,12 +474,7 @@ def _add_etag_match_headers(headers, **match_parameters):
 
     :type match_parameters: dict
     :param match_parameters: if*etag*match parameters to add.
-
-    :raises: :exc:`ValueError` if ``parameters`` is not a ``dict()``.
     """
-    if not isinstance(headers, dict):
-        raise ValueError("`headers` argument should be a dict() or a list().")
-
     for snakecase_name, header_name in _ETAG_MATCH_PARAMETERS:
         value = match_parameters.get(snakecase_name)
 
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -22,6 +22,7 @@ from hashlib import md5
 from datetime import datetime
 import os
 
+from six import string_types
 from six.moves.urllib.parse import urlsplit
 from google import resumable_media
 from google.cloud.storage.constants import _DEFAULT_TIMEOUT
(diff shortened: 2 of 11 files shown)


google/cloud/storage/_helpers.py
```diff
@@ -147,6 +153,8 @@ def reload(
         self,
         client=None,
         projection="noAcl",
+        if_etag_match=None,
```
daniellehanks, line 156: Question: How strict is the semver? Do I need to add new parameters to the end so as not to break parameter ordering for users who do not specify keys but instead rely on position?

google/cloud/storage/_helpers.py
```diff
@@ -449,6 +466,29 @@ def _convert_to_timestamp(value):
...
+    :param headers: Headers dict.
+
+    :type match_parameters: dict
+    :param match_parameters: if*etag*match parameters to add.
+
+    :raises: :exc:`ValueError` if ``parameters`` is not a ``dict()``.
+    """
+    if not isinstance(headers, dict):
+        raise ValueError("`headers` argument should be a dict() or a list().")
+
```
tseaver, line 482: Since this is a private helper, we can avoid this check.
```suggestion
```

google/cloud/storage/blob.py
```diff
@@ -692,12 +703,21 @@ def exists(
...
             if_metageneration_not_match=if_metageneration_not_match,
         )
+
+        headers = {}
+        _add_etag_match_headers(
+            headers, if_etag_match=if_etag_match, if_etag_not_match=if_etag_not_match
+        )
+        if not headers:
+            headers = None
+
```
tseaver, line 713: No harm passing the empty dict here:
```suggestion
```

google/cloud/storage/bucket.py
```diff
@@ -798,12 +809,21 @@ def exists(
...
             if_metageneration_not_match=if_metageneration_not_match,
         )
+
+        headers = {}
+        _add_etag_match_headers(
+            headers, if_etag_match=if_etag_match, if_etag_not_match=if_etag_not_match
+        )
+        if not headers:
+            headers = None
+
```
tseaver, line 819: Likewise here:

```suggestion
```

tests/unit/test__helpers.py
```diff
@@ -521,6 +554,39 @@ def read(self, block_size):
...
+        headers = {"foo": "bar"}
+        self._call_fut(headers, if_etag_match=ETAGS)
+        self.assertEqual(headers, EXPECTED_HEADERS)
+
+    def test_add_generation_match_parameters_headers_not_dict(self):
+        ETAG = "kittens"
+        headers = 42
+        with self.assertRaises(ValueError):
+            self._call_fut(headers, if_etag_match=ETAG)
+
```
tseaver, line 588: Per above, drop this test.

```suggestion
```

tests/unit/test_blob.py
```diff
@@ -705,6 +705,7 @@ def test_exists_miss_w_defaults(self):
         client._get_resource.assert_called_once_with(
             blob.path,
             query_params=expected_query_params,
+            headers=None,
```
tseaver, line 708: Empty dict, here and below (define `expected_headers` above):

```suggestion
            headers=expected_headers,
```

tests/unit/test_client.py
```diff
@@ -1502,6 +1529,12 @@ def _download_blob_to_file_helper(
             expected_retry = None
 
         headers = {"accept-encoding": "gzip"}
+        _helpers._add_etag_match_headers(
+            headers,
+            if_etag_match=extra_kwargs.get("if_etag_match"),
+            if_etag_not_match=extra_kwargs.get("if_etag_not_match"),
+        )
+
```
tseaver, line 1537: This would be clearer if we just assembled the values directly, rather than using the helper:

```suggestion
        if_etag_match = extra_kwargs.get("if_etag_match")
        if if_etag_match is not None:
            headers["If-Match"] = if_etag_match
        if_etag_not_match = extra_kwargs.get("if_not_etag_match")
        if if_etag_not_match is not None:
            headers["If-None-Match"] = if_etag_not_match
```

tseaver (changes requested):
Thanks very much for the patch!

daniellehanks:
> Thanks very much for the patch!

Appreciate the preview review so I can address requested changes before writing system tests.

google/cloud/storage/_helpers.py
daniellehanks: Excellent.

google/cloud/storage/_helpers.py
daniellehanks: Then why does `_add_generation_match_parameters` have a check? I was just being consistent with that.

google/cloud/storage/blob.py
daniellehanks: Cool, I didn't know if downstream was tolerant to that. Will remove all of these.

tests/unit/test_client.py
daniellehanks: Lol, that's originally what I had before I needed to copy/paste the logic into a 3rd place and ended up refactoring out into the helper. Understand the desire for DAMP tests though, so will revert this instance.

google/cloud/storage/_helpers.py
tseaver: Very good question!  The answer lies back in the mists of time, and my memory might be faulty, even with the help of `git log` and `git blame`.  `_add_generation_match_parameters` is used to add values to query parameters for a URL (versus headers here).  IIRC, in one specific case (the download URL with no `mediaLink` set), the order of parameters apparently mattered (at least, it used to) and so the function had to take *either* a `dict` or a `list`.  I don't recall why we didn't use a `collections.OrderedDict` there:  maybe we were still supporting Python 2.6 back then, and it  wasn't available.

At any rate, I'd prefer to "duck type" where possible.

daniellehanks: Sure, I don't particularly care about why it had to take both a list and dict, memories fade over time :D. I guess it makes sense to have a check in `_add_generation_match_parameters` because otherwise the function would just silently fail if there were no else block. `_add_etag_match_headers` will naturally raise a TypeError when trying to set a value if it isn't a dict (or dict like enough). Is that what you're getting at?

docs/generation_metageneration.rst
```diff
@@ -1,17 +1,28 @@
...
 
 Concepts
 --------
 
+.. _concept-etag:
+
+ETag
+::::::::::::::
+
+An ETag is returned as part of the response header whenever a resource is returned, as well as included in the resource itself. Users should make no assumptions about the value used in an ETag except that it changes whenever the underlying data changes, per the
```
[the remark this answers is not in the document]
daniellehanks: Haha, I didn't even notice that since my editor wraps at 80. Looks like the rest of the file uses 80.

google/cloud/storage/_helpers.py
daniellehanks: Check removed. I also removed the raises from the docstring, even though it can still raise TypeError, because of [this from the style guide](https://google.github.io/styleguide/pyguide.html#doc-function-raises):
>You should not document exceptions that get raised if the API specified in the docstring is violated (because this would paradoxically make behavior under violation of the API part of the API).

tests/unit/test_client.py
daniellehanks: It is slightly more complex than the suggestion because of str and list.

daniellehanks:
PR feedback addressed. Still need to add system tests. Hoping to find some time tonight.

google/cloud/storage/_helpers.py
```diff
@@ -34,6 +34,12 @@
 
 _DEFAULT_STORAGE_HOST = u"https://storage.googleapis.com"
 
+# etag match parameters in snake case and equivalent header
+_ETAG_MATCH_PARAMETERS = (
+    ("if_etag_match", "If-Match"),
+    ("if_etag_not_match", "If-None-Match"),
```
daniellehanks, line 40: I went with "not_match" for the parameter name even though the header is "none-match" to align with the generation parameter names. Really all the HTTP stuff is an implementation detail to most users of this SDK.

google/cloud/storage/_helpers.py
```diff
@@ -449,6 +466,24 @@ def _convert_to_timestamp(value):
...
+    :type match_parameters: dict
+    :param match_parameters: if*etag*match parameters to add.
+    """
+    for snakecase_name, header_name in _ETAG_MATCH_PARAMETERS:
+        value = match_parameters.get(snakecase_name)
+
+        if value is not None:
+            if isinstance(value, str):
+                value = [value]
+            headers[header_name] = ", ".join(value)
```
daniellehanks, line 484: It seems to me (though I'm not a spec expert) that the GCS API isn't following spec putting quotes around the etag values. E.g. `If-None-Match: COKaz4vVzfECEAE=` works as intended but `If-None-Match: "COKaz4vVzfECEAE="` does not. That said, the response header is also not quoted as it apparently should be `etag: COKaz4vVzfECEAE=` vs expected `etag: "COKaz4vVzfECEAE="`. Referencing [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) and [this](https://stackoverflow.com/questions/6719214/syntax-for-etag).

daniellehanks: Hmm, maybe its just the display layers or something (using Chrome dev tools and curl). Wikipedia does the same thing.

daniellehanks: microsoft.com has it quoted 🤔

daniellehanks:
I added two system tests. The first for the client as requested. The second for blob crud modeled after generation crud (but without write operations). There wasn't an existing system test for bucket metageneration stuff, so I didn't add one there for etags. Please let me know if there are any other tests you would like to see added.

I'm removing draft status as this PR is now complete from my perspective.

google/cloud/storage/_helpers.py
tseaver: Yes, that is exactly what I meant.

google/cloud/storage/_helpers.py
tseaver: Following the generation / metageneration pattern seems right to me as well.

google/cloud/storage/_helpers.py
tseaver: RFC 7232 is the actual authoritative spec for conditional HTTP requests: [`If-Match`](https://datatracker.ietf.org/doc/html/rfc7232#section-3.1) and [`If-None-Match`](https://datatracker.ietf.org/doc/html/rfc7232#section-3.2).  The [Collected ABNF appendix](https://datatracker.ietf.org/doc/html/rfc7232#appendix-C) does indeed seem to require the double quote.

The [GCS docs for the ETag header](https://cloud.google.com/storage/docs/json_api/v1/parameters#etag) show the quotes as well.

tests/system/test_client.py
```diff
@@ -102,3 +102,41 @@ def test_download_blob_to_file_w_uri(
...
+                    file_obj,
+                    if_etag_not_match=blob.etag,
+                )
+
+            with pytest.raises(exceptions.PreconditionFailed):
+                storage_client.download_blob_to_file(
+                    "gs://" + shared_bucket.name + "/" + filename,
+                    file_obj,
+                    if_etag_match="kittens",
+                )
```
tseaver, line 130: Because they share the same `file_obj`, the two successfull cases actually end up concatenating the payload twice, which is odd.  I think it would be better to split the four cases apart, using separate buffers for each:  since we are passing a file object, and not a filename, we can create `io.BytesIO` instances.  For the two cases which succeed, we can then assert that `buffer.getvalue() matches the expected payload.

tests/system/test_blob.py
```diff
@@ -235,6 +235,52 @@ def test_blob_crud_w_user_project(
...
+
+    info = file_data["simple"]
+    with open(info["path"], mode="rb") as to_read:
+        payload = to_read.read()
+
+    blob.upload_from_filename(info["path"])
+    blobs_to_delete.append(blob)
+    etag = blob.etag
+
+    blob = shared_bucket.blob("SmallFile")
```
tseaver, line 253: Please use a different name for the blob used to populate the bucket and the one used for testing the APIs passing ETags.

tests/system/test_blob.py
```diff
@@ -235,6 +235,52 @@ def test_blob_crud_w_user_project(
...
+    assert blob.exists(if_etag_not_match=wrong_etag)
+
+    with pytest.raises(exceptions.PreconditionFailed):
+        blob.reload(if_etag_match=wrong_etag)
+
+    with pytest.raises(exceptions.NotModified):
+        blob.reload(if_etag_not_match=etag)
+
+    blob.reload(if_etag_match=etag)
+    blob.reload(if_etag_not_match=wrong_etag)
```
tseaver, line 272: For cases with no assertable side effect, please add a comment, `  # no raise`.

tseaver (changes requested):
Thanks for the thoroughness of the systests you added!

google/cloud/storage/_helpers.py
daniellehanks: Yep, all the examples in the spec are also quoted. Fun times. Definitely out of scope for this PR. Just wanted to call it out. I would think the etag itself should have the quotes (i.e. `blob.etag = '"Csdlei="'`), so then this code would still function correctly. Of course getting etags quoted would be a far larger change spanning the backend, possibly ESF, and probably a fair amount of Hyrum's law.

It just stood out to me because I ran into a similar problem implementing the UI for BigQuery, which also doesn't follow the etag spec. In the end it meant we couldn't cache tables in the UI. But we also couldn't change the API without risking breaking clients. I'll leave it up to you if you want to file an internal ticket for that.

tests/system/test_client.py
daniellehanks: Yes, I did realize that but didn't really care since I wasn't checking the contents (as there are already tests that cover that). Will change as you suggest though. I'm assuming "split the four cases apart" doesn't mean to actually break them into four separate tests, as there is overhead to create the blob and such. If you do want that just let me know.

tests/system/test_blob.py
daniellehanks: I'm not sure I understand what you're asking for. The blob on this line has to use the same name as the one at the top of the test. I just have to reset the blob to clear out the metadata. Well, I guess I don't _have_ to. And if I do I should probably do it before each exists/reload/download call. Or I could just remove this line altogether.

daniellehanks:
FYI I ran the system tests for 2.7 and caught a bug in the etag helper (instance check for str doesn't work for unicode type in 2.7) that the unit tests didn't catch (because they don't round trip the etag). Fixed by using `six.string_types`.

tests/system/test_client.py
tseaver: One testcase function is still fine here -- we can just create new buffers for each of the four invocations.

google/cloud/storage/_helpers.py
tseaver: Effectively we're stuck with the as-implemented back-end.  @frankyn, @andrewsg, @tswast I'll leave it to y'all to open a ticket on the back-end (if one isn't there already) for the out-of-spec behavior.

tests/system/test_blob.py
tseaver: Not a different blob name (`"SmallFile"`), but a different *variable* name, e.g. `blob` and `other_blob` or some such, to make it clear that we are starting from scratch.

daniellehanks: Ah, will do.

tseaver (approved):

tseaver:
@googlebot I consent.

tests/system/test_client.py
```diff
@@ -102,3 +103,45 @@ def test_download_blob_to_file_w_uri(
...
+        buffer,
+        if_etag_not_match="kittens",
+    )
+    assert buffer.getvalue() == payload
+
+    buffer = io.BytesIO()
+    storage_client.download_blob_to_file(
+        "gs://" + shared_bucket.name + "/" + filename,
+        buffer,
+        if_etag_match=blob.etag,
```
tseaver: Ugh, the de-dent shortened this part just enough that `black` decided to pull it all onto one line.  I pushed a commit fixing it (6b3b6ab).

tests/unit/test_client.py
```diff
@@ -28,6 +29,7 @@
 from google.oauth2.service_account import Credentials
 from . import _read_local_json
 
+from google.cloud.storage import _helpers
```
tseaver, line 32: No longer used:
```suggestion
```

tseaver (changes requested):

tseaver (approved):

tseaver:
Thanks, again, @daniellehanks for your effort!

PR #496, opened 2021-07-07
fix(deps): update minimum dep versions to pick up bugfixes

### LINKED ISSUE #494: Use 'testing/constraints-{python_version}.txt' to test lower bound dependencies
The files exist, but `noxfile.py` does not use them.

tseaver:
Also, actually test those minimum versions using constraints file.

Closes #494


tswast (approved):
Thanks for adding the constraints tests!

PR #500, opened 2021-07-09
chore: pin Sphinx version in noxfile

### LINKED ISSUE #469: Noxfile is not updated with Sphinx version pin
`noxfile.py` is not updated with Sphinx version pin, please help take a look!

tseaver:
Closes #469.


dandhlee (approved):
Thanks for manually updating this, however let's keep the issue open if we can eventually template it.

PR #511, opened 2021-07-19
test: skip test__signing tests for mtls

arithmetic1728:
Add `no_mtls` decorator to skip all the tests in test__signing.py, if it is running in mtls testing environment.
The reason is mtls uses user credentials instead of service account credentials.

### DIFF
--- a/tests/system/test__signing.py
+++ b/tests/system/test__signing.py
@@ -48,13 +48,15 @@ def _create_signed_list_blobs_url_helper(
     assert response.status_code == 200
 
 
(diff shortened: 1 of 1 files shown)


busunkim96 (approved):

busunkim96:
I edited the title to remove `fix` so this doesn't end up in the Changelog. Otherwise LGTM!

arithmetic1728:
> I edited the title to remove `fix` so this doesn't end up in the Changelog. Otherwise LGTM!

Thanks!

PR #520, opened 2021-07-22
feat: add page_size to list_blobs, list_buckets

### LINKED ISSUE #425: `max_results` argument of `Client.list_blobs` does not behave correctly
According to [this](https://cloud.google.com/storage/docs/json_api/v1/objects/list)

`maxResults` should be:
> Maximum combined number of entries in items[] and prefixes[] to return in a **single page** of responses. Because duplicate entries in prefixes[] are omitted, fewer total results may be returned than requested. The service uses this parameter or 1,000 items, whichever is smaller. 

When I do this:
```
client = storage.Client()
bucket = storage.Bucket(client, "<bucket_name>")
blobs_iterator = client.list_blobs(bucket, max_results=10)
for page in blobs_iterator.pages:
    print(page.num_items)
```
Output:
```
10
```

Only one iteration of the loop is completed even though there are 100 blobs in the bucket. It seems the `max_results` argument is limiting the **total** results rather than **per page** results.

Related to #19

ryanyuan:
Adding page_size to list_blobs() and list_buckets() to have better support page_iterator.

Fixes #425 🦕

### DIFF
--- a/google/cloud/storage/client.py
+++ b/google/cloud/storage/client.py
@@ -384,6 +384,7 @@ class Client(ClientWithProject):
         max_results=None,
         extra_params=None,
         page_start=page_iterator._do_nothing_page_start,
+        page_size=None,
         timeout=_DEFAULT_TIMEOUT,
         retry=DEFAULT_RETRY,
     ):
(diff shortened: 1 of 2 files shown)


tseaver (approved):

tseaver:
@ryanyuan Thanks very much for the patch!

PR #526, opened 2021-07-23
fix: make 'requests.exceptions.ChunkedEncodingError retryable by default

### LINKED ISSUE #525: Make 'requests.exceptions.ChunkedEncodingError' retryable by default.
See https://github.com/googleapis/python-bigquery/issues/801 for context.

tseaver:
Closes #525.

### DIFF
--- a/google/cloud/storage/retry.py
+++ b/google/cloud/storage/retry.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import requests
+import requests.exceptions as requests_exceptions
 
 from google.api_core import exceptions as api_exceptions
 from google.api_core import retry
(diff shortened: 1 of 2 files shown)


andrewsg:
I think this is https://github.com/googleapis/google-resumable-media-python/issues/240 which can't be solved here in the client library - PR is in the works by @cojenco

tseaver:
@andrewsg I don't know whether the JSON API (versus the download API) can cause `requests` to perceive a `ChunkedEncodingError` -- I do know that the BigQuery API can emit them from JSON API calls (e.g, `Client.list_rows()`, see https://github.com/googleapis/python-bigquery/issues/801).

andrewsg (approved):
Thanks for pointing that out. Sounds like we may need it here too, then.

PR #539, opened 2021-08-17
chore: generate python samples templates in owlbot.py

busunkim96:
Generate python samples templates in owlbot.py

### DIFF
--- a/owlbot.py
+++ b/owlbot.py
@@ -18,6 +18,7 @@ import re
 
 import synthtool as s
 from synthtool import gcp
+from synthtool.languages import python
 
 common = gcp.CommonTemplates()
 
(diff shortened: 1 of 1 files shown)


owlbot.py
```diff
@@ -25,6 +25,7 @@
 # Add templated files
 # ----------------------------------------------------------------------------
 templated_files = common.py_library(
+python.py_samples(skip_readmes=True)
```
tseaver, line 28: Syntax error.

tseaver (changes requested):

parthea (approved):

parthea (approved):

owlbot.py
parthea: Good catch!

tseaver:
Flaky systest reported in #541.  Re-running.

tseaver (approved):

Issue #540, opened 2021-08-18
AttributeError: 'NoneType' object has no attribute '_http' when using Blob.upload_from_string


Ark-kun:
#### Environment details

  - OS type and version: Linux
  - Python version: 3.9.2
  - pip version: 21.2.4
  - `google-cloud-storage` version: 1.42.0

#### Code example

```python
from google.cloud import storage
storage.Blob.from_string(uri="gs://avolkov/mlflow/Model1/model_info.json").upload_from_string("foo")
```

#### Stack trace
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/google/home/avolkov/.local/lib/python3.9/site-packages/google/cloud/storage/blob.py", line 2853, in upload_from_string
    self.upload_from_file(
  File "/usr/local/google/home/avolkov/.local/lib/python3.9/site-packages/google/cloud/storage/blob.py", line 2580, in upload_from_file
    created_json = self._do_upload(
  File "/usr/local/google/home/avolkov/.local/lib/python3.9/site-packages/google/cloud/storage/blob.py", line 2382, in _do_upload
    response = self._do_multipart_upload(
  File "/usr/local/google/home/avolkov/.local/lib/python3.9/site-packages/google/cloud/storage/blob.py", line 1865, in _do_multipart_upload
    transport = self._get_transport(client)
  File "/usr/local/google/home/avolkov/.local/lib/python3.9/site-packages/google/cloud/storage/blob.py", line 809, in _get_transport
    return client._http
AttributeError: 'NoneType' object has no attribute '_http'
```

Ark-kun:
I'm encountering this error almost everywhere. Even in cases where I initially start with `storage.Client()`:
```python
        blobs = storage.Client().list_blobs(
            bucket_or_name=bucket,
            prefix=src_path,
        )
        for blob in blobs:
            blob.bucket.rename_blob(blob, blob.name + ".bak")
```

```
   blob.bucket.rename_blob(blob, new_path)
  File "/usr/local/google/home/avolkov/.local/lib/python3.9/site-packages/google/cloud/storage/bucket.py", line 1988, in rename_blob
    new_blob = self.copy_blob(
  File "/usr/local/google/home/avolkov/.local/lib/python3.9/site-packages/google/cloud/storage/bucket.py", line 1868, in copy_blob
    copy_result = client._post_resource(
AttributeError: 'NoneType' object has no attribute '_post_resource'
```

[...]

Issue #541, opened 2021-08-18
'test_bucket_update_labels' systest flakes with 400


tseaver:
From [this Kokoro failure](https://source.cloud.google.com/results/invocations/1a2dad90-74f8-4160-b0c0-e912f6a4fbe3/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-storage%2Fpresubmit%2Fsystem-3.8/log):

```python
__________________________ test_bucket_update_labels ___________________________

storage_client = <google.cloud.storage.client.Client object at 0x7fa25901f070>
buckets_to_delete = [<Bucket: update-labels-1629301504047>]

    def test_bucket_update_labels(storage_client, buckets_to_delete):
        bucket_name = _helpers.unique_name("update-labels")
        bucket = _helpers.retry_429_503(storage_client.create_bucket)(bucket_name)
        buckets_to_delete.append(bucket)
        assert bucket.exists()

        updated_labels = {"test-label": "label-value"}
        bucket.labels = updated_labels
        bucket.update()
        assert bucket.labels == updated_labels

        new_labels = {"another-label": "another-value"}
        bucket.labels = new_labels
        bucket.patch()
        assert bucket.labels == new_labels

        bucket.labels = {}
>       bucket.update()

tests/system/test_bucket.py:114:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
google/cloud/storage/bucket.py:949: in update
    super(Bucket, self).update(
google/cloud/storage/_helpers.py:394: in update
    api_response = client._put_resource(
google/cloud/storage/client.py:537: in _put_resource
    return self._connection.api_request(
google/cloud/storage/_http.py:78: in api_request
    return call()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <google.cloud.storage._http.Connection object at 0x7fa25901f310>
method = 'PUT', path = '/b/update-labels-1629301504047'
query_params = {'projection': 'full'}
[...]

PR #545, opened 2021-08-19
docs: pass explicit 'client' in '{Blob.Bucket}.from_string' examples

tseaver:
See #540.

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -405,7 +405,7 @@ class Blob(_PropertyMixin):
             >>> from google.cloud import storage
             >>> from google.cloud.storage.blob import Blob
             >>> client = storage.Client()
-            >>> blob = Blob.from_string("gs://bucket/object")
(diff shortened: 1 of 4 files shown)


google/cloud/storage/blob.py
```diff
@@ -405,7 +405,7 @@ def from_string(cls, uri, client=None):
             >>> from google.cloud import storage
             >>> from google.cloud.storage.blob import Blob
             >>> client = storage.Client()
-            >>> blob = Blob.from_string("gs://bucket/object")
+            >>> blob = Blob.from_string("gs://bucket/object", client=client)
```
cojenco, line 408: Thanks for the change! Could we also update both parameter descriptions with a note to strongly encourage specifying a client?

tseaver: 745e531

cojenco (approved):
Thanks so much for the clarification!

tseaver:
Flaky systest is #541.

PR #555, opened 2021-08-24
tests: update retry conf tests using the updated testbench emulator

cojenco:
The retry conf tests are ran against the storage/testbench emulator. This addresses new changes made to the testbench emulator:

- (change) a test bucket is automatically created only if the env var `GOOGLE_CLOUD_CPP_STORAGE_TEST_BUCKET_NAME` is set
- (fix) buckets in the testbench should always have a metageneration
- (fix) typo

### DIFF
--- a/tests/conformance/test_conformance.py
+++ b/tests/conformance/test_conformance.py
@@ -443,8 +443,8 @@ method_mapping = {
     "storage.buckets.getIamPolicy": [bucket_get_iam_policy],
     "storage.buckets.insert": [client_create_bucket, bucket_create],
     "storage.buckets.list": [client_list_buckets],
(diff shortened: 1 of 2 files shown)


tests/conformance/test_conformance.py
```diff
@@ -252,6 +251,7 @@ def bucket_set_iam_policy(client, _preconditions, **resources):
     if _preconditions:
         bucket.set_iam_policy(policy)
     else:
+        policy.etag = None
```
tseaver, line 254: Maybe a comment here to explain why this is different from e.g. `bucket_update`? 
```suggestion
        # IAM policies have no metageneration:  clear ETag to avoid checking that it matches.
        policy.etag = None
```

tritone (approved):

PR #575, opened 2021-09-02
chore: unbundle renovate reports

tseaver:
We need to be able to close those for the Python 2.7 requirements,
while allowing through the normal Python 3 reports.

Closes #578.


parthea (approved):

tseaver:
Failing systest due to changed(?) back-end behavior -- reported as #577.

cojenco:
Thanks @tseaver for the fix!

PR #570, opened 2021-09-01
fix: do not append duplicates to user agent string

ddelgrosso1:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-storage/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #565 🦕

Issue #565, opened 2021-08-31
Use of Batch adds duplicate, redundant gcloud-python/x.xx.x to User-Agent header in an unbounded fashion


michio-nikaido:
### Overview
We first became aware of this issue when we began seeing "Error 413 (Request Entity Too Large)" responses to a variety of GCS requests after we implemented blob.reload() calls in a Batch context.   We discovered after capturing the HTTP requests that the User-Agent headers for both the "outer" batch request and the contained requests were all filled with hundreds of 'gcloud-python/1.40.0' specifications.  For example, we've seen a single batch request containing 51 reload requests with each   User-Agent specification containing over 100 repetitions of 'gcloud-python/1.40.0'.  Overall, this request had 5,300 occurrences of the same 'gcloud-python/1.40.0' 

#### Analysis
When batch calls are made, a ' gcloud-python/x.xx.x ' is blindly appended to the underlying connection's _client_info.user_agent string:

https://github.com/googleapis/python-storage/blob/master/google/cloud/storage/_http.py#L59

As a result, successive batch calls result in the connection._client_info.user-agent string accumulating more and more, duplicate ' gcloud-python/x.xx.x ' specifications.  If a Client object continues making batch requests, the user-agent property continues to grow without limit.

#### Environment details
  - OS type and version: Ubuntu 20.04
  - Python version: 3.7.11
  - pip version: 21.2.4
  - `google-cloud-storage` version: 1.42.0

#### Repro Script
```
from google.api_core import exceptions as g_exc
from google.cloud.storage.blob import Blob
from google.cloud.storage.client import Client

project = '<GCP PROJCT NAME>'
bucket_name = '<GCS BUCKET NAME>'

# Client.__init__ initializes self._connection which initiallizes a Connection object as defined in
# google.cloud.storage._http.py which automatically appends a 'gcloud-python/x.xx.x' spec to the
# the user_agent property
c = Client(project=project)
print(c._connection._client_info.user_agent)

bucket = c.get_bucket(bucket_name)
blobs = list(c.list_blobs(bucket))[:1]

# ----- First batch call adds duplicate user_agent spec
print("Before batch: "  + c._connection._client_info.user_agent)

try:
    # Creation of a Batch object results in another call which again appends another (duplicate) 
    # 'gcloud-python/x.xx.x' to the user_agent property.
    with c.batch()as batch:
        print("Batch context: " + batch._client_info.user_agent)
        for blob in blobs:
            blob.reload()
[...]

PR #579, opened 2021-09-02
tests: revise test to run without setting credentials in env

cojenco:
Part of #570 and #565

### DIFF
--- a/tests/unit/test__http.py
+++ b/tests/unit/test__http.py
@@ -231,13 +231,18 @@ class TestConnection(unittest.TestCase):
 
     def test_duplicate_user_agent(self):
         # Regression test for issue #565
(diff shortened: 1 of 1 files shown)


busunkim96 (approved):

tseaver (commented):
@cojenco the need to use a "real" `Client` here has a bit of "code smell":  The `Connection` class shouldn't need to use `Client` for anything, even tests.  Would a better fix for #565 be to have the `Batch` class construct a copy of the original `ClientInfo`?

cojenco:
@tseaver Makes sense, thanks! I've revised the test to mock out `Client`. `ClientInfo` is propagated from the `Client` passed into the `Batch` class constuctor.

PR #580, opened 2021-09-03
tests: add retry conformance test cases

cojenco:
Update test schema json file and add retry conformance test cases:
- Project-based operations (hmacKey, service account)
- ACL operations
- Scenario 4 never idempotent operations
- Add instructions cases 
`["return-reset-connection", "return-reset-connection"]`
`["return-reset-connection", "return-503"]`
- Update test names

Filed issue #582 to track special cases: The PATCH API requests for ACL in python-storage utilize `storage.buckets.patch` or `storage.objects.patch`, instead of  ACL specific endpoints such as `storage.bucket_acl.patch`  `storage.object_acl.patch` `storage.default_object_acl.patch`

An individual test case is generated for each scenario ID, instruction set, method, and method_invocation_mapping_library_method. Test names are displayed as test-S{scenario_id}-{method}-{client-library-method-name}-{instructions index}, for example, `test-S1-storage.buckets.get-bucket_reload-1`

### DIFF
--- a/tests/conformance/retry_strategy_test_data.json
+++ b/tests/conformance/retry_strategy_test_data.json
@@ -85,13 +85,28 @@
       },
       {
         "id": 4,
-        "description": "non idempotent",
+        "description": "non_idempotent",
         "cases": [
           {
-            "instructions": []
+            "instructions": ["return-503"]
           }
         ],
-        "methods": [],
+        "methods": [
+          {"name": "storage.bucket_acl.delete",             "resources": ["BUCKET"]},
+          {"name": "storage.bucket_acl.insert",             "resources": ["BUCKET"]},
+          {"name": "storage.bucket_acl.patch",              "resources": ["BUCKET"]},
+          {"name": "storage.bucket_acl.update",             "resources": ["BUCKET"]},
+          {"name": "storage.default_object_acl.delete",     "resources": ["BUCKET"]},
+          {"name": "storage.default_object_acl.insert",     "resources": ["BUCKET"]},
+          {"name": "storage.default_object_acl.patch",      "resources": ["BUCKET"]},
+          {"name": "storage.default_object_acl.update",     "resources": ["BUCKET"]},
+          {"name": "storage.hmacKey.create",                "resources": []},
+          {"name": "storage.notifications.insert",          "resources": ["BUCKET"]},
+          {"name": "storage.object_acl.delete",             "resources": ["BUCKET", "OBJECT"]},
+          {"name": "storage.object_acl.insert",             "resources": ["BUCKET", "OBJECT"]},
+          {"name": "storage.object_acl.patch",              "resources": ["BUCKET", "OBJECT"]},
+          {"name": "storage.object_acl.update",             "resources": ["BUCKET", "OBJECT"]}
+        ],
         "preconditionProvided": false,
         "expectSuccess": false
       }
--- a/tests/conformance/retry_strategy_test_data.json
+++ b/tests/conformance/retry_strategy_test_data.json
@@ -6,6 +6,12 @@
         "cases": [
           {
             "instructions": ["return-503", "return-503"]
+          },
+          {
+            "instructions": ["return-reset-connection", "return-reset-connection"]
+          },
+          {
+            "instructions": ["return-reset-connection", "return-503"]
           }
         ],
         "methods": [
@@ -41,6 +47,12 @@
         "cases": [
           {
             "instructions": ["return-503", "return-503"]
+          },
+          {
+            "instructions": ["return-reset-connection", "return-reset-connection"]
+          },
+          {
+            "instructions": ["return-reset-connection", "return-503"]
           }
         ],
         "methods": [
@@ -65,6 +77,9 @@
         "cases": [
           {
             "instructions": ["return-503"]
+          },
+          {
+            "instructions": ["return-reset-connection"]
           }
         ],
         "methods": [
@@ -89,6 +104,9 @@
         "cases": [
           {
             "instructions": ["return-503"]
+          },
+          {
+            "instructions": ["return-reset-connection"]
           }
         ],
         "methods": [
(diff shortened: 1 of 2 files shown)


tests/conformance/retry_strategy_test_data.json
```diff
@@ -85,13 +85,28 @@
       },
       {
         "id": 4,
-        "description": "non idempotent",
+        "description": "non_idempotent",
         "cases": [
           {
-            "instructions": []
+            "instructions": ["return-503"]
```
tritone, line 91: We should make a separate PR at some point with additional instruction sets-- also, I see this doesn't exactly match what you have in your conformance-tests repo PR.

tests/conformance/test_conformance.py
```diff
@@ -217,8 +225,44 @@ def client_list_hmac_keys(client, _preconditions, **_):
...
+def hmac_key_exists(client, _preconditions, **resources):
+    access_id = resources.get("hmac_key").access_id
+    hmac_key = HMACKeyMetadata(client, access_id=access_id)
+    hmac_key.exists()
+
+
+def hmac_key_reload(client, _preconditions, **resources):
+    access_id = resources.get("hmac_key").access_id
+    hmac_key = HMACKeyMetadata(client, access_id=access_id)
+    hmac_key.reload()
```
tritone, line 242: What is the use case for this? Don't really understand why there are multiple get methods. (but that's not a problem for this PR!)

tests/conformance/test_conformance.py
```diff
@@ -462,14 +641,22 @@ def blob_create_resumable_upload_session(client, _preconditions, **resources):
...
-        bucket_delete,
-    ],  # S1 end
-    "storage.buckets.patch": [bucket_patch],  # S2/S3 start
+    "storage.objects.list": [client_list_blobs, bucket_list_blobs, bucket_delete],
+    "storage.serviceaccount.get": [client_get_service_account_email],  # S1 end
+    "storage.buckets.patch": [
+        bucket_patch,
+        bucket_make_public,
+        bucket_make_private,
+        bucket_acl_save,
```
tritone, line 650: Ah, this is tricky-- I see that in our doc we define this as `storage.bucket_acl.patch` rather than a variant on `storage.buckets.patch`. The endpoint looks like this: https://cloud.google.com/storage/docs/json_api/v1/bucketAccessControls/patch .

Do we not have this separated out in the emulator routing?

tests/conformance/test_conformance.py
```diff
@@ -485,9 +672,18 @@ def blob_create_resumable_upload_session(client, _preconditions, **resources):
         blobwriter_write,
         blob_create_resumable_upload_session,
     ],
-    "storage.objects.patch": [blob_patch],
+    "storage.objects.patch": [
+        blob_patch,
+        object_acl_save,
```
tritone, line 677: Same as for bucket ACLs above-- this should be an entirely separate route than storage.objects.patch I believe.

tests/conformance/test_conformance.py
```diff
@@ -423,6 +467,131 @@ def blob_create_resumable_upload_session(client, _preconditions, **resources):
...
+def bucket_make_private(client, _preconditions, **resources):
+    if _preconditions:
+        pytest.skip(_BUCKET_ACL_PATCH_MSG)
+    bucket = client.bucket(resources.get("bucket").name)
+    bucket.make_private()
+
+
+def bucket_make_public(client, _preconditions, **resources):
+    if _preconditions:
+        pytest.skip(_BUCKET_ACL_PATCH_MSG)
```
tritone, line 497: These should be never-idempotent so this message should never come up.

tritone (commented):
Few comments about the ACL methods, we should figure this out together. Otherwise looks good.

tests/conformance/test_conformance.py
cojenco: Thanks @tritone for the comments! We have this separated out in the emulator routing. It's the python client causing it to be tricky.

In python-storage, all of the ACL patch operations are done by calling `storage.buckets.patch` or `storage.objects.patch`, as the[ patch requests paths are either the bucket path or the object path](https://github.com/googleapis/python-storage/blob/3a489ad7e1cd2fe5adb443e781e72cf645a23331/google/cloud/storage/acl.py#L503). Therefore we're not calling any of the `storage.bucket_acl.patch` endpoints. Not sure if this is a bug in the library. I can check with other languages.

tests/conformance/retry_strategy_test_data.json
cojenco: Well noted! The difference now is the # of instructions cases.

tests/conformance/test_conformance.py
cojenco: There are 3 entry points in the library that call `storage.hmacKey.get`, and we want to make sure all of them are retried correctly. `hmac_key.reload()` is often used after instantiating a `HMACKeyMetadata` and before making updates.

tests/conformance/retry_strategy_test_data.json
cojenco: Added instruction cases. S1 ~ S4 now fully matches the conformance-tests repo json.

tests/conformance/test_conformance.py
tritone: Discussed offline, this is by design and not a problem.

tritone (approved):

tests/conformance/test_conformance.py
tseaver: @tritone, @cojenco FWIW, my, rusty seven-year-old memory says that `storage.bucket_acl.patch`didn't used to work as desired.  See [this commit](https://github.com/googleapis/google-cloud-python/commit/b7f1b7747bb493a5df1982b6460a26023d6ab3c3) from this dim mists of time.

tests/conformance/test_conformance.py
cojenco: Filed #582 for adding preconditions and retry config support.

PR #586, opened 2021-09-13
fix: add preconditions and retry config support to ACL patch operationss

### LINKED ISSUE #582: Expose preconditions to ACL methods
The PATCH API requests for ACL in python-storage utilize `storage.buckets.patch` or `storage.objects.patch`, instead of  ACL specific endpoints such as `storage.bucket_acl.patch`  `storage.object_acl.patch` `storage.default_object_acl.patch`

- Add preconditions support to the ACL library methods to make the patch operations conditionally idempotent
- Add retry configuration

Note:  IAM configurations `storage.buckets.getIamPolicy` `storage.buckets.setIamPolicy` are highly encouraged to control access instead

cojenco:
Add preconditions and retry config support to the ACL patch operations.  The PATCH API requests for ACL in python-storage utilize `storage.buckets.patch` or `storage.objects.patch`

- ACL patch methods in ACL (`_save` /`save` /`save_predefined` /`clear`)
- ACL patch methods in Bucket (`make_public` /`make_private`)
- ACL patch methods in Blob (`make_public` /`make_private`)
- Add and update tests
- Update doscstrings

Note: IAM configurations storage.buckets.getIamPolicy storage.buckets.setIamPolicy are highly encouraged to control access instead

Fixes #582

### DIFF
--- a/google/cloud/storage/acl.py
+++ b/google/cloud/storage/acl.py
@@ -84,8 +84,10 @@ This list of tuples can be used as the ``entity`` and ``role`` fields
 when sending metadata for ACLs to the API.
 """
 
+from google.cloud.storage._helpers import _add_generation_match_parameters
 from google.cloud.storage.constants import _DEFAULT_TIMEOUT
 from google.cloud.storage.retry import DEFAULT_RETRY
+from google.cloud.storage.retry import DEFAULT_RETRY_IF_METAGENERATION_SPECIFIED
 
 
 class _ACLEntity(object):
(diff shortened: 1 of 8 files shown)


tritone (approved):
Thanks for adding these!

tseaver (commented):
We patch ACLs using the [`objectAccessControls.pach`](https://cloud.google.com/storage/docs/json_api/v1/objectAccessControls/patch) and [`defaultObjectAccessControls.pach`](https://cloud.google.com/storage/docs/json_api/v1/defaultObjectAccessControls/patch) methods, which do not AFAICT take the metageneration / generation into account:  instead, they rely on the resource's `Etag` for concurrency.  Note the absence of the `if_*_match` parameters on those docs page.

tseaver:
At a minimum, if the back-end docs are incomplete, and the back-end actually supports these semantics, we need to add one or more system tests which exercise them.

cojenco:
@tseaver We patch ACLs here using `storage.buckets.patch` and `storage.objects.patch`, instead of  ACL specific endpoints (such as   `storage.object_acl.patch`). See [here](https://github.com/googleapis/python-storage/blob/3d43049cc94f65f71932b210e6edc5f195d408c4/google/cloud/storage/acl.py#L506) that we're using bucket and object paths for patching resources.

You are right about `storage.object_acl.patch` `storage.default_object_acl.patch` relying on the resource's `Etag` for concurrency. 

That is actually one of the considerations why `storage.buckets.patch` and `storage.objects.patch` are used here instead. If you recall our previous discussion,  the `If-Match` header that specifies an entity tag does not work for patch operations (internal tracking bug). 

Since we are using `storage.buckets.patch` and `storage.objects.patch` under the hood, it makes sense to expose the related `if_*_match` parameters.

tseaver:
My bad -- I had talked myself into "remembering" the need to back out some similar changes for ACLs in past PRs.  You are indeed correct that we do *not* use the ACL specific APIs, and so should be free to use the `if-*-match` headers.

tseaver:
I do think we should add a system test which exercises the new headers, though.

PR #589, opened 2021-09-16
fix: pin six as a required dependency

### LINKED ISSUE #588: Continuous test failing because of missing dependency `six`
Most continuous tests are failing with the following error message:
```
Traceback:
/usr/local/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/unit/test_retry.py:17: in <module>
    from google.cloud.storage import _helpers
google/cloud/storage/__init__.py:35: in <module>
    from google.cloud.storage.batch import Batch
google/cloud/storage/batch.py:27: in <module>
    import six
E   ModuleNotFoundError: No module named 'six'
```

dandhlee:
`six` used to be a transitive dependency through `protobuf`, but now it is not specified in any of the other dependencies thus the tests are failing, and any other installations to the current latest version would fail.

`protobuf` also dropped support for Python 2.7 as of 3.18.0. Pinning it for 2.7.

Fixes #588 🦕

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,7 @@ dependencies = [
     "google-resumable-media >= 1.3.0, < 3.0dev; python_version>='3.6'",
     "requests >= 2.18.0, < 3.0.0dev",
     "googleapis-common-protos < 1.53.0; python_version<'3.0'",
+    "six",
 ]
 extras = {}
 
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ dependencies = [
     "google-resumable-media >= 1.3.0, < 2.0dev; python_version<'3.0'",
     "google-resumable-media >= 1.3.0, < 3.0dev; python_version>='3.6'",
     "requests >= 2.18.0, < 3.0.0dev",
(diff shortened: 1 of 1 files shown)


busunkim96 (approved):

cojenco (approved):
Thanks for pinning these!

setup.py
```diff
@@ -37,7 +37,9 @@
     "google-resumable-media >= 1.3.0, < 2.0dev; python_version<'3.0'",
     "google-resumable-media >= 1.3.0, < 3.0dev; python_version>='3.6'",
     "requests >= 2.18.0, < 3.0.0dev",
+    "protobuf < 3.18.0; python_version<'3.0'",
```
tseaver, line 40: Should we be adding an un-pinned `protobuf` for Python 3.x (to avoid future transitive dep failures):
```suggestion
    "protobuf < 3.18.0; python_version<'3.0'",
    "protobuf; python_version>='3.6'",
```

PR #593, opened 2021-09-17
chore(deps): include six in direct dependencies

crwilcox:
six is imported directly in this package. As upstream packages are moving away from python 2, include this dependency directly here to ensure it is installed.


crwilcox:
Instance of direct usage at https://github.com/googleapis/python-storage/blob/ef65048656d7c67acc4f7600d2c4004d89fe8ba7/google/cloud/storage/_signing.py

Also used by a few tests.

crwilcox:
closing as I missed #589 . We may still want to look into version ranging six, though I wouldn't anticipate breaking changes with it moving forward anyway.

Issue #599, opened 2021-09-27
'test_new_bucket_created_w_unspecified_pap' systest flakes


tseaver:
From [this Kokoro failure](https://source.cloud.google.com/results/invocations/7ace8da7-f27c-4364-b826-21df53a9c3d3/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-storage%2Fpresubmit%2Fsystem-3.8/log):

```python
__________________ test_new_bucket_created_w_unspecified_pap ___________________

storage_client = <google.cloud.storage.client.Client object at 0x7fca9111bc10>
buckets_to_delete = [<Bucket: new-w-pap-unspecified-1632759084261>]
blobs_to_delete = []

    def test_new_bucket_created_w_unspecified_pap(
        storage_client, buckets_to_delete, blobs_to_delete,
    ):
        from google.cloud.storage import constants

        bucket_name = _helpers.unique_name("new-w-pap-unspecified")
        bucket = storage_client.bucket(bucket_name)
        bucket.iam_configuration.uniform_bucket_level_access_enabled = True
        bucket.create()
        buckets_to_delete.append(bucket)

>       assert (
            bucket.iam_configuration.public_access_prevention
            == constants.PUBLIC_ACCESS_PREVENTION_UNSPECIFIED
        )
E       AssertionError: assert 'inherited' == 'unspecified'
E         - unspecified
E         + inherited

tests/system/test_bucket.py:820: AssertionError
```

Breathtender:
Hm, seems perhaps this is linked to https://github.com/googleapis/python-storage/issues/597
We are currently changing unspecified to inherited.

cojenco:
^If needed, we can submit a PR to temporarily skip this test.

PR #600, opened 2021-09-27
fix: skip tests that use unspecified pap until we get the change in

unforced:
Temporary fix for #599, until we get change for #597 fully merged in

### DIFF
--- a/tests/unit/test_bucket.py
+++ b/tests/unit/test_bucket.py
@@ -350,7 +350,6 @@ class Test_IAMConfiguration(unittest.TestCase):
 
         return mock.create_autospec(Bucket, instance=True)
 
-    @pytest.mark.skip(reason="Unspecified PAP is changing to inherited")
     def test_ctor_defaults(self):
         bucket = self._make_bucket()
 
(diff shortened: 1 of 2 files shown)


shaffeeullah (approved):

tests/unit/test_bucket.py
```diff
@@ -350,6 +350,7 @@ def _make_bucket():
 
         return mock.create_autospec(Bucket, instance=True)
 
+    @pytest.mark.skip(reason="Unspecified PAP is changing to inherited")
```
tseaver, line 353: We shouldn't need to disable the unit test just now -- it still passes, and skipping it causes a coverage failure.

tseaver (changes requested):

tests/unit/test_bucket.py
Breathtender: Ah ok, that makes sense, thanks!

tseaver (approved):

cojenco (approved):
Thanks AG!

PR #602, opened 2021-09-28
Pap change

### LINKED ISSUE #597: Change Public Access Prevention "unspecified" to "inherited"
GCS is changing the `unspecified` public access prevention value to `inherited`.

This will require the following changes:

1. Modify the values in the library. Constants/enum values for unspecified: these should be marked as deprecated. They can stay in the library if deletion would constitute a breaking change
2. Update the system tests.
3. Create a new sample with region tag public_access_prevention_inherited. So we don't break any documentation links, please do not delete the unspecified sample or sample test from your repo at this point in time.

Please see nodejs example here: https://github.com/googleapis/nodejs-storage/pull/1614/files

Please wait for approval from myself or @tritone before merging or releasing this change so we can coordinate around release dates. New inherited value should be rolled out to allowlisted testing projects this coming week; should be at 100% by 9/30

unforced:
fix #597


Breathtender:
Disregard for a sec, this got dirty, cleaning it up

PR #644, opened 2021-11-02
feat: add ignore_flush parameter to BlobWriter

### LINKED ISSUE #631: Cannot stream-write a zipfile
I'm attempting to stream-write an uncompressed zip file (200MB+ -> 1GB+ eventually, mostly ~3MB images) to avoid writing to disk.  Unfortunately when the zip file is closed it attempts to at least partially flush the stream, and the storage client seems to assume that a flush will only occur to close the streamed file, which isn't the case here (and seems wrong in the general sense, since buffers are often flushed when they reach saturation).

#### Environment details

  - OS type and version: Ubuntu 20.04 (custom devcontainer docker image)
  - Python version: 3.8.10
  - pip version: 21.3
  - `google-cloud-storage` version: 1.42.3

#### Code example

```python
import zipfile

from google.cloud.storage import Blob, Client

client = Client("some-account")
blob = Blob.from_string("gs://some-bucket/some-folder/something.zip", client)

with blob.open(mode="wb") as blob_file, \
    zipfile.ZipFile(blob_file, mode="w") as zip_file:
    # Empty/not empty doesn't matter, the same error is generated
    pass
```

#### Stack trace
```
  File "s.py", line 14, in <module>
    pass
  File "/usr/lib/python3.8/zipfile.py", line 1312, in __exit__
    self.close()
  File "/usr/lib/python3.8/zipfile.py", line 1839, in close
    self._write_end_record()
  File "/usr/lib/python3.8/zipfile.py", line 1947, in _write_end_record
    self.fp.flush()
  File "/workspaces/someproject/.pyenv/lib/python3.8/site-packages/google/cloud/storage/fileio.py", line 401, in flush
    raise io.UnsupportedOperation(
io.UnsupportedOperation: Cannot flush without finalizing upload. Use close() instead.
```

</ br>

#### Workaround
Insert an `io.BufferedWriter`:

```python
import io
import zipfile

from google.cloud.storage import Blob, Client

client = Client("some-account")
blob = Blob.from_string("gs://some-bucket/some-folder/something.zip", client)

with blob.open(mode="wb") as blob_file, \
    io.BufferedWriter(blob_file) as binary_file, \
    zipfile.ZipFile(binary_file, mode="w") as zip_file:
    # Empty/not empty doesn't matter, the same error is generated
    pass
```

This results in the file being written to cloud storage (and at least for a simple case, including correctly written contents), but it prints a new error:
```
  File "s.py", line 24, in <module>
    pass
  File "/workspaces/someproject/.pyenv/lib/python3.8/site-packages/google/cloud/storage/fileio.py", line 406, in close
    self._checkClosed()  # Raises ValueError if closed.
  File "/workspaces/someproject/.pyenv/lib/python3.8/site-packages/google/cloud/storage/fileio.py", line 413, in _checkClosed
    raise ValueError("I/O operation on closed file.")
ValueError: I/O operation on closed file.
```

andrewsg:
Fixes #631  🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -3760,6 +3760,7 @@ class Blob(_PropertyMixin):
         self,
         mode="r",
         chunk_size=None,
+        ignore_flush=None,
         encoding=None,
         errors=None,
         newline=None,
@@ -3800,6 +3801,19 @@ class Blob(_PropertyMixin):
             is sent. Writes are implemented as a "resumable upload", so
             chunk_size for writes must be exactly a multiple of 256KiB as with
             other resumable uploads. The default is 40 MiB.
+
+        :type ignore_flush: bool
+        :param ignore_flush:
+            (Optional) For non text-mode writes, makes flush() do nothing
+            instead of raising an error. flush() without closing is not
+            supported by the remote service and therefore calling it normally
+            results in io.UnsupportedOperation. However, that behavior is
+            incompatible with some consumers and wrappers of file objects in
+            Python, such as zipfile.ZipFile or io.TextIOWrapper. Setting
+            ignore_flush will cause flush() to successfully do nothing, for
+            compatibility with those contexts. The correct way to actually flush
+            data to the remote server is to close() (using a context manager,
+            such as in the example, will cause this to happen automatically).
 
         :type encoding: str
         :param encoding:
@@ -3873,14 +3887,24 @@ class Blob(_PropertyMixin):
                 raise ValueError(
                     "encoding, errors and newline arguments are for text mode only"
                 )
+            if ignore_flush:
+                raise ValueError(
+                    "ignore_flush argument is for non-text write mode only"
+                )
             return BlobReader(self, chunk_size=chunk_size, **kwargs)
         elif mode == "wb":
             if encoding or errors or newline:
                 raise ValueError(
                     "encoding, errors and newline arguments are for text mode only"
                 )
-            return BlobWriter(self, chunk_size=chunk_size, **kwargs)
+            return BlobWriter(
+                self, chunk_size=chunk_size, ignore_flush=ignore_flush, **kwargs
+            )
         elif mode in ("r", "rt"):
+            if ignore_flush:
+                raise ValueError(
+                    "ignore_flush argument is for non-text write mode only"
+                )
             return TextIOWrapper(
                 BlobReader(self, chunk_size=chunk_size, **kwargs),
                 encoding=encoding,
@@ -3888,8 +3912,13 @@ class Blob(_PropertyMixin):
                 newline=newline,
             )
         elif mode in ("w", "wt"):
+            if ignore_flush is False:
+                raise ValueError(
+                    "ignore_flush is required for text mode writing and "
+                    "cannot be set to False"
+                )
             return TextIOWrapper(
-                BlobWriter(self, chunk_size=chunk_size, text_mode=True, **kwargs),
+                BlobWriter(self, chunk_size=chunk_size, ignore_flush=True, **kwargs),
                 encoding=encoding,
                 errors=errors,
                 newline=newline,
--- a/google/cloud/storage/fileio.py
+++ b/google/cloud/storage/fileio.py
@@ -229,11 +229,23 @@ class BlobWriter(io.BufferedIOBase):
         writes must be exactly a multiple of 256KiB as with other resumable
         uploads. The default is the chunk_size of the blob, or 40 MiB.
 
-    :type text_mode: boolean
+    :type text_mode: bool
     :param text_mode:
-        Whether this class is wrapped in 'io.TextIOWrapper'. Toggling this
-        changes the behavior of flush() to conform to TextIOWrapper's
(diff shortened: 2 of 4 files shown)


google/cloud/storage/blob.py
```diff
@@ -3873,23 +3887,38 @@ def open(
                 raise ValueError(
                     "encoding, errors and newline arguments are for text mode only"
                 )
+            if ignore_flush:
+                raise ValueError(
+                    "ignore_flush argument is for non-text write mode only"
+                )
```
tseaver, line 3893: Can we move this check out to a single point, e.g.:

```python

        if ignore_flush and mode in ("r", "rt", "rb", "w", "wt"):
            raise ValueError(
                "ignore_flush argument is for non-text write mode only"
            )
```

google/cloud/storage/fileio.py
```diff
@@ -229,11 +229,23 @@ class BlobWriter(io.BufferedIOBase):
         writes must be exactly a multiple of 256KiB as with other resumable
         uploads. The default is the chunk_size of the blob, or 40 MiB.
 
-    :type text_mode: boolean
+    :type text_mode: bool
     :param text_mode:
-        Whether this class is wrapped in 'io.TextIOWrapper'. Toggling this
-        changes the behavior of flush() to conform to TextIOWrapper's
-        expectations.
+        (Deprecated) A synynom for pass_on_flush. For backwards-compatibility,
```
tseaver, line 234: Should this be?
```suggestion
        (Deprecated) A synynom for `ignore_flush`. For backwards-compatibility,
```

google/cloud/storage/fileio.py
```diff
@@ -394,12 +406,14 @@ def tell(self):
         return self._buffer.tell() + len(self._buffer)
 
     def flush(self):
-        if self._text_mode:
+        if self._ignore_flush:
```
tseaver, line 409: Can we refactor to avoid the early return?  E.g.:

```python
    def flush(self):
        if not self._ignore_flush:
            raise io.UnsupportedOperation(
                "Cannot flush without finalizing upload. Use close() instead, or "
                "set ignore_flush=True when constructing this class (see "
                "docstring)."
            )
```

tseaver (changes requested):

google/cloud/storage/fileio.py
andrewsg: Yes, thank you for catching that.

google/cloud/storage/blob.py
andrewsg: This would only deduplicate two of the three ValueErrors, because the third has a different behavior (errors on false) and a different error message. To keep consistent organization I'd prefer to keep the current pattern, even if it's more verbose.

google/cloud/storage/fileio.py
andrewsg: done

andrewsg:
PTAL

google/cloud/storage/blob.py
tseaver: OK -- I'd missed that the error case was reversed for `w` / `wt`.

One might argue that pulling the validation out from the "construct the IO helper" would make that clearer, but I will defer here.

tseaver (approved):

google/cloud/storage/blob.py
```diff
@@ -3873,23 +3887,38 @@ def open(
                 raise ValueError(
                     "encoding, errors and newline arguments are for text mode only"
                 )
+            if ignore_flush:
```
cojenco, line 3890: @andrewsg Should these instead be `if ignore_flush is not None:`

if `ignore_flush=False` is accidentally set for reads, it would cause an AttributeError or such.

google/cloud/storage/fileio.py
```diff
@@ -229,11 +229,23 @@ class BlobWriter(io.BufferedIOBase):
...
-        Whether this class is wrapped in 'io.TextIOWrapper'. Toggling this
-        changes the behavior of flush() to conform to TextIOWrapper's
-        expectations.
+        (Deprecated) A synonym for ignore_flush. For backwards-compatibility,
+        if True, sets ignore_flush to True. Use ignore_flush instead. This
+        parameter will be removed in a future release.
+
+    :type ignore_flush: bool
+    :param ignore_flush:
+        Makes flush() do nothing instead of raise an error. flush() without
```
cojenco, line 240: +1 to a handwritten doc for FileIO now that I'm looking at it. We could have more thorough explanations written on an individual page, and streamline the docstrings by linking to the page.

google/cloud/storage/blob.py
andrewsg: Good point to focus on. I thought about that but decided that since False was the actual desired behavior, we didn't need to guard against it here. I don't think it would cause an AttributeError - is there a place that looks suspicious?

cojenco: I see what you mean. That's right; it wouldn't cause an error since `ignore_flush` is not passed in to the `BlobRead` constructor after all. I misread that.

cojenco (approved):

PR #645, opened 2021-11-02
samples: Add GCS fileio samples

unforced:


Breathtender:
This still likely needs a sample that covers some more complex things with bytes and zip and flush. @andrewsg

cojenco (commented):
Thanks for adding these! Could you also update the [samples readme](https://github.com/googleapis/python-storage/blob/main/samples/README.md) to include the new samples?

Breathtender:
> Thanks for adding these! Could you also update the [samples readme](https://github.com/googleapis/python-storage/blob/main/samples/README.md) to include the new samples?

Thanks for the reminder, added that it.

By the way, I notice that in the README it says how to run the samples but never demonstrates any args, and most of the samples will fail out if they're not passed certain arguments. Do you think that's a concern?

cojenco:
> > Thanks for adding these! Could you also update the [samples readme](https://github.com/googleapis/python-storage/blob/main/samples/README.md) to include the new samples?
> 
> Thanks for the reminder, added that it.
> 
> By the way, I notice that in the README it says how to run the samples but never demonstrates any args, and most of the samples will fail out if they're not passed certain arguments. Do you think that's a concern?

Indeed, that was something I considered. Adding the args in the commands is definitely more clear; but I also noticed that it was more prone to error and easily out-of date with future changes when I referenced other libraries/samples. How about instead I add a note about using Cloud Shell and running the commands with the required arguments. WDYT?

Breathtender:
> > > Thanks for adding these! Could you also update the [samples readme](https://github.com/googleapis/python-storage/blob/main/samples/README.md) to include the new samples?
> > 
> > 
> > Thanks for the reminder, added that it.
> > By the way, I notice that in the README it says how to run the samples but never demonstrates any args, and most of the samples will fail out if they're not passed certain arguments. Do you think that's a concern?
> 
> Indeed, that was something I considered. Adding the args in the commands is definitely more clear; but I also noticed that it was more prone to error and easily out-of date with future changes when I referenced other libraries/samples. How about instead I add a note about using Cloud Shell and running the commands with the required arguments. WDYT?

Ya that sounds like a nice way to do it; just some documentation somewhere so people don't get confused when the errors pop up.

samples/README.md
```diff
@@ -486,6 +488,22 @@ View the [source code](https://github.com/googleapis/python-storage/blob/main/sa
...
 `python storage_enable_versioning.py`
 
+-----
+### FileIO Write-Read
+[![Open in Cloud Shell][shell_img]](https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googleapis/python-storage&page=editor&open_in_editor=samples/snippets/storage_fileio_write_read.py,samples/README.md)
+
+View the [source code](https://github.com/googleapis/python-storage/blob/main/samples/snippets/storage_fileio_write_read.py). To run this sample:
+
+
+`python storage_fileio_write_read.py`
```
andrewsg, line 498: If the sample needs command-line args to work then I think we should mention that here as well.

samples/snippets/storage_fileio_pandas.py
```diff
@@ -0,0 +1,55 @@
...
+    with blob.open("r") as f:
+        df = pd.read_csv(f)
+        print(df.values.flatten())
+
+
+# [END storage_fileio_pandas]
+
+if __name__ == "__main__":
+    pandas_write_read(
+        bucket_name=sys.argv[1],
```
andrewsg, line 53: I'm not clear if best practice in python-docs-samples is to use sys.argv or a helper like argparse (I asked the same about the concurrent pubsub sample work). Can you please verify this is the standard method for Python samples these days?

samples/snippets/storage_fileio_pandas.py
```diff
@@ -0,0 +1,55 @@
...
+    # The ID of your new GCS object
+    # blob_name = "storage-object-name"
+
+    import pandas as pd
+
+    storage_client = storage.Client()
+    bucket = storage_client.bucket(bucket_name)
+    blob = bucket.blob(blob_name)
+
+    with blob.open("w") as f:
```
andrewsg, line 40: I think this is a good minimalist sample concept.

samples/snippets/storage_fileio_pandas.py
```diff
@@ -0,0 +1,55 @@
...
+import sys
+
+"""Sample that creates and consumes a GCS blob using pandas with file-like IO
+"""
+
+# [START storage_fileio_pandas]
+from google.cloud import storage
+
+
+def pandas_write_read(bucket_name, blob_name):
```
andrewsg, line 26: We could consider breaking this out into a write and a read sample, and then running them one after the other in `__main__`. That way users copy/pasting can easily copy/paste the writes or reads as they need them, but the command-line sample still works the same way.

andrewsg (changes requested):

samples/README.md
Breathtender: Yes, done.

samples/snippets/storage_fileio_pandas.py
Breathtender: Hmm, a little research has shown me more places tending towards argparse, so perhaps that's the standard?

samples/snippets/storage_fileio_pandas.py
Breathtender: Do you mean by that we may not need the other hello world write-read sample at all?

samples/snippets/storage_fileio_pandas.py
Breathtender: Good idea, on it.

samples/README.md
cojenco: Per our discussion, we can submit a separate PR to update all command-line args in the readme, if we decide to keep them. 

For now, could you help update the command-line args to `<BUCKET_NAME>` `<BLOB_NAME>` for consistency? Thanks!

i.e. `python storage_fileio_write_read.py <BUCKET_NAME> <BLOB_NAME>`

Breathtender: Yes, done.

samples/snippets/storage_fileio_pandas.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
+
+# Copyright 2019 Google Inc. All Rights Reserved.
```
cojenco, line 3: Sorry missed this earlier. Copyright year should be 2021?

Breathtender: Woops! Good catch :)

cojenco (approved):
LGTM, thanks!

samples/snippets/storage_fileio_pandas.py
andrewsg: Okay, well, given our conversation about Cloud-wide standards for samples, I'm ok with approving this as-is and we'll sort that out afterwards.

andrewsg (approved):
Thanks!

PR #650, opened 2021-11-11
tests: [WIP] add download data integrity conformance test cases

cojenco:
[WIP] 
- Add data integrity checks for downloads in the retry conf tests
- Add instruction case "return-broken-stream" that simulates interrupted media mid-stream


cojenco:
Replaced with #702

PR #655, opened 2021-11-19
tests: speed up retry conformance tests

cojenco:
- Speed up retry conf tests to run in parallel with pytest plugin
- Update readme file

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -167,7 +167,12 @@ def conftest_retry(session):
 
     # Run #CPU processes in parallel if no test session arguments are passed in.
     if session.posargs:
-        test_cmd = ["py.test", "--quiet", conformance_test_folder_path, *session.posargs]
+        test_cmd = [
+            "py.test",
+            "--quiet",
+            conformance_test_folder_path,
+            *session.posargs,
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -162,15 +162,24 @@ def conftest_retry(session):
...
+    # Then install this package in-place.
+    session.install("pytest", "pytest-xdist")
     session.install("-e", ".")
 
     # Run py.test against the conformance tests.
     if conformance_test_exists:
-        session.run("py.test", "--quiet", conformance_test_path, *session.posargs)
+        session.run(
+            "py.test", "-n", "auto", "--quiet", conformance_test_path, *session.posargs
+        )
```
tseaver, line 174: I think it might be a better DX to inject the `"-n", "auto"` into the command line only if `session.posargs` is empty:  that way, the user could chose to e.g. restrict the number of cores available, or even distribute the tests across multiple hosts.

cojenco: @tseaver good idea! I've updated the noxfile. PTAL, thanks!

tseaver (approved):

PR #665, opened 2021-12-10
samples: add async upload sample

unforced:
FYI: I'll be OOO for winter holidays starting Monday, so any requested changes after business hours on Friday, please feel free to amend and merge without me.


samples/snippets/storage_async_upload.py
```diff
@@ -0,0 +1,60 @@
...
+import asyncio
+import sys
+
+
+"""Sample that asynchronously uploads a file to GCS
+"""
+
+
+# [START storage_async_upload]
+async def async_upload_blob(bucket_name):
```
andrewsg, line 26: Can you add a comment that the sample can be run, in practice, with asyncio.run() as it is in the tests? So that people unfamiliar with async functions know they don't have to change their whole program.

andrewsg (changes requested):
Thanks!

andrewsg (approved):
Looks great, thanks!

PR #668, opened 2021-12-13
feat: add raw_download kwarg to BlobReader

### LINKED ISSUE #666: Allow `raw_download` kwarg for BlobReader
The current value for `BlobReader`'s [`VALID_DOWNLOAD_KWARGS`](https://github.com/googleapis/python-storage/blob/ee41b3f883632026e5d0460847122bd1c37b9b2d/google/cloud/storage/fileio.py#L31) constant is:

```python
VALID_DOWNLOAD_KWARGS = {
    "if_generation_match",
    "if_generation_not_match",
    "if_metageneration_match",
    "if_metageneration_not_match",
    "timeout",
    "retry",
}
```

Is there a reason why `raw_download` isn't in the list? I'd like to add it in order to read bytes of a compressed file *without* Cloud Storage doing the decompression.

Because of the current limitation, I have to write my own implementation using the `download_as_bytes(raw_download=True, ...)`. It'd be great if this parameter could be added to the allow-list so I can use the `BlobReader` API instead.

cojenco:
- Allow `raw_download` kwarg for `BlobReader`
- Add unit test
- Update docstrings for `blob.open()`

Fixes #666 🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -3850,6 +3850,10 @@ class Blob(_PropertyMixin):
             - ``timeout``
             - ``retry``
 
+            For downloads only, the following additional arguments are supported:
+
+            - ``raw_download``
+
             For uploads only, the following additional arguments are supported:
 
             - ``content_type``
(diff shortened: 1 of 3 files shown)


andrewsg (approved):
LGTM, thanks!

PR #676, opened 2021-12-21
docs: use writeable streamin example for 'download_blob_to_file'

micahjsmith:
The docstring shows an example of opening a file and passing the file handle to `download_blob_to_file`. However, this example must be untested, as it fails. The default mode for `open` is read-only. Instead, the file needs to be opened in some writeable model. This PR modifies the docstring example accordingly.


micahjsmith:
signed CLA

tseaver (approved):

tseaver:
@andrewsg, @cojenco Can one of y'all investigate the CLA failure?

PR #678, opened 2021-12-29
fix: project can be inferred from the environment and credentials upon client instantiation

### LINKED ISSUE #134: project ID error when project explicitly passed.
#### Environment details

  - OS type and version: Mac OS X
  - Python version: 3.7.6
  - pip version: 20.0.2
  - `google-cloud-storage` version: 1.28.1

pipenv list from a test venv:

```
(test) lana~/test% pip list
Package                  Version   
------------------------ ----------
cachetools               4.1.0     
certifi                  2020.4.5.1
chardet                  3.0.4     
google-api-core          1.17.0    
google-auth              1.14.1    
google-cloud-core        1.3.0     
google-cloud-storage     1.28.1    
google-resumable-media   0.5.0     
googleapis-common-protos 1.51.0    
idna                     2.9       
pip                      19.2.3    
protobuf                 3.11.3    
pyasn1                   0.4.8     
pyasn1-modules           0.2.8     
pytz                     2020.1    
requests                 2.23.0    
rsa                      4.0       
setuptools               41.2.0    
six                      1.14.0    
urllib3                  1.25.9  
```
#### Steps to reproduce

Normally, I can instantiate a `storage.Client()` object and access a bucket, even without setting a project explicitly, and even when my current default project is different from the project where the bucket lives. However, recently, I've seen some errors doing this, even though I'm sure the credentials are right (as if I set the project explicitly, I can still access the bucket).

  1. Create a clean Python3 environment with pipenv or venv. Either don't have a default project setup via `gcloud`, or be configured with a different project than the bucket lives in (ideally, since bucket names are unique, it would be great if this could just magically work, even if the currently configured project is not where the bucket lives). install latest google-cloud-storage
  2. Call some code, even with the bucket project defined.
`WARNING: No project ID could be determined. Consider running `gcloud config set project` or setting the GOOGLE_CLOUD_PROJECT environment variable`
If project is not set at all, also get:
`OSError: Project was not passed and could not be determined from the environment.`

With my regular setup (even checked all the versions I had installed), I don't get the same error, and can access the bucket just fine, even if I don't explicitly set the project. I have the same versions of google-auth, google-storage, etc. there (also via pip). Happy to provide more details, but would love any suggestions about what might be causing this under certain conditions.

#### Code example

```python

storage_client = storage.Client(project='buckets-actual-project')
bucket = storage_client.get_bucket('my-bucket')
blobs = bucket.list_blobs()
for blob in blobs:
    ## do some other stuff
```

### LINKED ISSUE #653: Could not connect to the cloud storage when running on the Google cloud compute engine

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/python-storage/issues
  - Search StackOverflow: https://stackoverflow.com/questions/tagged/google-cloud-platform+python

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS type and version:  Centos
  - Python version: 3.8
  - pip version: `pip --version`
  - `google-cloud-storage` version: `pip show google-cloud-storage`

#### Steps to reproduce

  1. I set up some Compute engines in Google cloud, and use k8s to create a pod
  2. the pod crash since it got an error 
  
image
  3. but when I just use the python shell of Compute engines directly. it succeeds
image
  4. I also run it in a docker container, it also got an error
  
image

#### Code example

```python
gcs.client.Client()
```

#### Stack trace
```
![image](https://user-images.githubusercontent.com/32006870/142226538-39fd0539-c009-42b6-82fd-a163f38f6949.png)
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

cojenco:
The storage `client` is expected to fall back to the default project inferred from the environment or credentials when a `project` is not explicitly passed in upon instantiation. 

This is the same behavior cited in the [documentation](https://github.com/googleapis/python-storage/blob/5e49e54adbf1fd1457d01df1ae2df63025ac7ae4/google/cloud/storage/client.py#L67-L69) as well as other client libraries (i.e. [BigQuery client](https://github.com/googleapis/python-bigquery/blob/881c081feb248e6a2b0d7ad72667924b959f8120/google/cloud/bigquery/client.py#L223)). The changes will also align with the [guidances](https://github.com/googleapis/python-cloud-core/issues/27) in the google-cloud-client [superclasses](https://github.com/googleapis/python-cloud-core/blob/main/google/cloud/client/__init__.py#L233-L280).

- revise storage.Client constructor
- remove extra client side validations
- update tests

Fixes #653 
Fixes #134 🦕

### DIFF
--- a/google/cloud/storage/client.py
+++ b/google/cloud/storage/client.py
@@ -57,9 +57,6 @@ from google.cloud.storage.retry import DEFAULT_RETRY
 from google.cloud.storage.retry import ConditionalRetryPolicy
 
 
-_marker = object()
-
-
 class Client(ClientWithProject):
     """Client to bundle configuration needed for API requests.
 
@@ -104,22 +101,13 @@ class Client(ClientWithProject):
 
     def __init__(
         self,
-        project=_marker,
+        project=None,
         credentials=None,
         _http=None,
         client_info=None,
         client_options=None,
     ):
         self._base_connection = None
-
-        if project is None:
-            no_project = True
-            project = "<none>"
-        else:
-            no_project = False
-
-        if project is _marker:
-            project = None
 
         super(Client, self).__init__(
             project=project,
@@ -147,9 +135,6 @@ class Client(ClientWithProject):
             if client_options.api_endpoint:
                 api_endpoint = client_options.api_endpoint
                 kw_args["api_endpoint"] = api_endpoint
-
-        if no_project:
-            self.project = None
 
         self._connection = Connection(self, **kw_args)
         self._batch_stack = _LocalStack()
--- a/google/cloud/storage/client.py
+++ b/google/cloud/storage/client.py
@@ -906,9 +906,6 @@ class Client(ClientWithProject):
         if project is None:
             project = self.project
 
-        if project is None:
-            raise ValueError("Client project not set:  pass an explicit project.")
-
         if requester_pays is not None:
             warnings.warn(
                 "requester_pays arg is deprecated. Use Bucket().requester_pays instead.",
@@ -917,7 +914,10 @@ class Client(ClientWithProject):
             )
             bucket.requester_pays = requester_pays
 
-        query_params = {"project": project}
+        query_params = {}
+
+        if project is not None:
+            query_params = {"project": project}
 
         if predefined_acl is not None:
             predefined_acl = BucketACL.validate_predefined(predefined_acl)
@@ -1348,10 +1348,10 @@ class Client(ClientWithProject):
         if project is None:
             project = self.project
 
-        if project is None:
-            raise ValueError("Client project not set:  pass an explicit project.")
+        extra_params = {}
 
-        extra_params = {"project": project}
+        if project is not None:
+            extra_params = {"project": project}
 
         if prefix is not None:
             extra_params["prefix"] = prefix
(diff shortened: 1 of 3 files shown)


tseaver (commented):
@cojenco I'm not sure how this PR addresses #653 or #134.  The current behavior is that if the user does **not** pass an explicit project, the constructor falls back to inferring the project from the credentials / environment, but if the user passes an **explicit** `None` for the project, the constructor remembers that fact, preventing its accidental use when calling `Client.create_bucket` or `Client.list_buckets`.  For the rationale, see https://github.com/googleapis/google-cloud-python/issues/4239.

The two referenced bugs are both about failures to find the project ID in the credentials inferred from the environment, and thus won't be affected by this change at all.

Removing the client-side validation from `Client.create_bucket` / `Client.list_buckets` shouldn't hurt anything:  the [`buckets.insert`](https://cloud.google.com/storage/docs/json_api/v1/buckets/insert) and [`buckets.list`](https://cloud.google.com/storage/docs/json_api/v1/buckets/list) APIs will both raise exceptions if `project` is not passed.

cojenco:
Thanks for the context @tseaver  I understand the use of the client constructor `_marker` now. I was expecting the `project` to be inferred from the environment when `project=None` upon instantiating a client, hence the changes. 

I'll look more into the issue why the project ID is not found in the credentials inferred from the environment. With the current constructor logic, It seems beyond the surface of the storage client. Closing this PR, and will move the client-side validation revisions to a separate one.

PR #679, opened 2021-12-30
feat: avoid authentication with storage emulator

### LINKED ISSUE #324: Avoid Authentication When Connecting to Storage Emulator
**Problem**
When connecting to a 3rd party storage emulator, like `gcloud-storage-emulator`, authentication is required. This requires injecting credentials into containers and adding conditional logic in code to authenticate when running against emulators.

```python
import os
from firebase_admin import initialize_app, storage

os.environ['STORAGE_EMULATOR_HOST'] = 'http://localhost:9090'
assert 'GOOGLE_APPLICATION_CREDENTIALS' not in os.environ

initialize_app()
storage.bucket() # google.auth.exceptions.DefaultCredentialsError
```

**Solution**
When `STORAGE_EMULATOR_HOST` is set, use an insecure connection, as is done for `firestore`.

cojenco:
Fixes #324  🦕


unforced (approved):
Oh very good, thanks for doing this.

PR #702, opened 2022-01-24
tests: add retry conf s8 test cases and data integrity checks

cojenco:
This adds scenario 8 and data integrity checks to the retry conformance tests
- add s8 to test schema - tests downloads handle complex retries
- instruction cases `return-broken-stream` and `return-broken-stream-after-256K` simulate interrupted media mid-stream
- move fixtures to newly added `conftest.py` file
- create a 16MB payload/object fixture `file_data` to test large file complex downloads (and later on resumable uploads)
- add data integrity checks for downloads in the retry conf tests


unforced (approved):
LGTM

PR #703, opened 2022-01-25
feat: allow no project in client methods using storage emulator

### LINKED ISSUE #699: `create_bucket` expects a project even with an emulator "ValueError: Client project not set:  pass an explicit project."
Thanks for fixing https://github.com/googleapis/python-storage/issues/324. I just noticed that `create_bucket` still expects a project. Would it make sense that it'd also detect `STORAGE_EMULATOR_HOST` and use a "fake" project automatically like [here](https://github.com/cojenco/python-storage/blob/820f3ead72236db05521fd36213b2acaab048e8b/google/cloud/storage/client.py#L157)?

#### Environment details

  - OS type and version: MacOS 12.1
  - Python version: `python --version` Python 3.9.9
  - pip version: `pip --version` pip 21.3.1
  - `google-cloud-storage` version: `pip show google-cloud-storage` 2.1.0

#### Steps to reproduce

  1. See the example below.

#### Code example

```python
import os

HOST = "localhost"
PORT = 9023
BUCKET = "test-bucket"

os.environ["STORAGE_EMULATOR_HOST"] = f"http://{HOST}:{PORT}"
client = storage.Client()

bucket = client.create_bucket(BUCKET)
```

#### Stack trace
```
  File "/Users/user/workspace/temp/test.py", line 12, in <module>
    bucket = client.create_bucket(BUCKET)
  File "/Users/user/workspace/temp/.venv/lib/python3.9/site-packages/google/cloud/storage/client.py", line 940, in create_bucket
    raise ValueError("Client project not set:  pass an explicit project.")
ValueError: Client project not set:  pass an explicit project.
```

cojenco:
When using a storage emulator, allow no project/fake project marker in `client.create_bucket()` and `client.list_buckets()`

Alongside, this removes client-side validation and falls back to API validations. In this case, if a project is not set, instead of raising a `ValueError` in the client, the GCS server will return a `BadRequest` with message `Required parameter: project`

Fixes #699

### DIFF
--- a/google/cloud/storage/client.py
+++ b/google/cloud/storage/client.py
@@ -932,12 +932,22 @@ class Client(ClientWithProject):
 
         """
         bucket = self._bucket_arg_to_bucket(bucket_or_name)
+        query_params = {}
 
         if project is None:
             project = self.project
 
-        if project is None:
-            raise ValueError("Client project not set:  pass an explicit project.")
+        # Use no project if STORAGE_EMULATOR_HOST is set
+        if _get_storage_host().find("storage.googleapis.com") < 0:
+            if project is None:
+                project = _get_environ_project()
+            if project is None:
+                project = "<none>"
+
+        # Only include the project parameter if a project is set.
+        # If a project is not set, falls back to API validation (BadRequest). Removes client-side validation.
+        if project is not None:
+            query_params = {"project": project}
 
         if requester_pays is not None:
             warnings.warn(
@@ -946,8 +956,6 @@ class Client(ClientWithProject):
                 stacklevel=1,
             )
             bucket.requester_pays = requester_pays
-
-        query_params = {"project": project}
 
         if predefined_acl is not None:
             predefined_acl = BucketACL.validate_predefined(predefined_acl)
@@ -1375,13 +1383,22 @@ class Client(ClientWithProject):
         :returns: Iterator of all :class:`~google.cloud.storage.bucket.Bucket`
                   belonging to this project.
         """
+        extra_params = {}
+
         if project is None:
             project = self.project
 
-        if project is None:
-            raise ValueError("Client project not set:  pass an explicit project.")
+        # Use no project if STORAGE_EMULATOR_HOST is set
+        if _get_storage_host().find("storage.googleapis.com") < 0:
+            if project is None:
+                project = _get_environ_project()
+            if project is None:
+                project = "<none>"
 
-        extra_params = {"project": project}
+        # Only include the project parameter if a project is set.
+        # If a project is not set, falls back to API validation (BadRequest). Removes client-side validation.
+        if project is not None:
+            extra_params = {"project": project}
 
         if prefix is not None:
             extra_params["prefix"] = prefix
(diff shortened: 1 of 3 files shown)


google/cloud/storage/client.py
```diff
@@ -932,12 +932,22 @@ def create_bucket(
...
-            raise ValueError("Client project not set:  pass an explicit project.")
+        # Use no project if STORAGE_EMULATOR_HOST is set
+        if _get_storage_host().find("storage.googleapis.com") < 0:
+            if project is None:
+                project = _get_environ_project()
+            if project is None:
+                project = "<none>"
+
+        # Only include the project parameter if a project is set.
+        # If a project is not set, falls back to API validation (BadRequest). Removes client-side validation.
```
andrewsg, line 948: "Removes client-side validation" may be a confusing comment here when read in the future, without the context of this PR.

andrewsg (changes requested):

google/cloud/storage/client.py
```diff
@@ -932,12 +932,22 @@ def create_bucket(
...
         bucket = self._bucket_arg_to_bucket(bucket_or_name)
+        query_params = {}
 
         if project is None:
             project = self.project
 
-        if project is None:
-            raise ValueError("Client project not set:  pass an explicit project.")
+        # Use no project if STORAGE_EMULATOR_HOST is set
+        if _get_storage_host().find("storage.googleapis.com") < 0:
```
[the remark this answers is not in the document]
cojenco: Thanks for the review! Updated according to your comments.

The [user ask](https://github.com/googleapis/python-storage/issues/324#issuecomment-978125907) is to be able to run the same code against a storage emulator as well as the live server (without writing extra conditions or wrappers), often seen in testing. 

`client.create_bucket()` and `client.list_buckets()` are the only 2 library methods that explicitly expect a `project`. I think it makes sense to allowing these 2 methods to detect the emulator and use a fake project marker.

google/cloud/storage/client.py
cojenco: Makes sense, revised.

google/cloud/storage/client.py
```diff
@@ -1375,13 +1384,22 @@ def list_buckets(
...
+        if _BASE_STORAGE_URI not in _get_storage_host():
+            if project is None:
+                project = _get_environ_project()
+            if project is None:
+                project = "<none>"
 
-        extra_params = {"project": project}
+        # Only include the project parameter if a project is set.
+        # If a project is not set, falls back to API validation (BadRequest).
+        if project is not None:
```
Breathtender, line 1401: Same as above; wondering the necessity of this given the previous lines.

unforced (approved):
LGTM; aside from the comments but I might just be missing something there.

google/cloud/storage/client.py
```diff
@@ -932,12 +933,22 @@ def create_bucket(
...
+        if _BASE_STORAGE_URI not in _get_storage_host():
+            if project is None:
+                project = _get_environ_project()
+            if project is None:
+                project = "<none>"
+
+        # Only include the project parameter if a project is set.
+        # If a project is not set, falls back to API validation (BadRequest).
+        if project is not None:
+            query_params = {"project": project}
```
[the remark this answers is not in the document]
Breathtender: Ohhh nevermind, it could fail the first prereq about _base_storage_uri and therefore it wouldn't hit the other ones that set project.

andrewsg (approved):

PR #710, opened 2022-02-03
samples: add endpoint sample

Breathtender:
Adds sample for changing endpoint on client.

One note: I accidentally merged an earlier version of this sample into my last README fix, which is why this includes a deletion as well. Oops.

### DIFF
--- a/samples/snippets/storage_set_client_endpoint.py
+++ b/samples/snippets/storage_set_client_endpoint.py
@@ -19,27 +19,21 @@ import sys
 """Sample that creates a new bucket in a specified region
 """
 
-# [START storage_create_bucket_location]
+# [START storage_set_client_endpoint]
 
 from google.cloud import storage
 
 
-def create_bucket_location(bucket_name, bucket_location):
-    """Creates bucket in specified region."""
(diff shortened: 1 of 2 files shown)


samples/snippets/storage_set_client_endpoint.py
```diff
@@ -19,27 +19,23 @@
...
+    """Initiates client with specified endpoint."""
+    # api_endpoint = 'https://storage.googleapis.com'
 
-    storage_client = storage.Client()
-    bucket = storage_client.bucket(bucket_name)
-    bucket.create(location=bucket_location)
+    storage_client = storage.Client(client_options={'api_endpoint': api_endpoint})
 
-    print(f"{bucket.name} created in {bucket.location}.")
+    print(f"client initiated with endpoint: {api_endpoint}")
```
frankyn, line 33: Could you rely on storatge_client value instead of the parameter `api_endpoint` to show devs how to check which endpoint is being used?

frankyn (changes requested):

frankyn (approved):

PR #715, opened 2022-02-14
fix: add user agent in python-storage when calling resumable media

unforced:
BEGIN_COMMIT_OVERRIDE
fix: add user agent in python-storage when calling resumable media
END_COMMIT_OVERRIDE

### DIFF
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -581,3 +581,20 @@ def _api_core_retry_to_resumable_media_retry(retry, num_retries=None):
         return resumable_media.RetryStrategy(max_retries=num_retries)
     else:
         return resumable_media.RetryStrategy(max_retries=0)
+
+
+def _get_default_headers(user_agent):
+    """Get the headers for a request.
+
+    Args:
+        user_agent (str): The user-agent for requests.
+    Returns:
+        Dict: The headers to be used for the request.
+    """
+    return {
+        "Accept": "application/json",
+        "Accept-Encoding": "gzip, deflate",
+        "User-Agent": user_agent,
+        "X-Goog-Api-Client": user_agent,
+        "content-type": "application/json",
+    }
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -583,7 +583,10 @@ def _api_core_retry_to_resumable_media_retry(retry, num_retries=None):
         return resumable_media.RetryStrategy(max_retries=0)
 
 
-def _get_default_headers(user_agent, content_type="application/json; charset='UTF-8"):
+def _get_default_headers(
+    user_agent,
+    content_type="application/json; charset='UTF-8",
+    x_upload_content_type=None):
     """Get the headers for a request.
 
     Args:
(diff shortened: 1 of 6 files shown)


google/cloud/storage/_helpers.py
```diff
@@ -581,3 +581,20 @@ def _api_core_retry_to_resumable_media_retry(retry, num_retries=None):
...
+    Args:
+        user_agent (str): The user-agent for requests.
+    Returns:
+        Dict: The headers to be used for the request.
+    """
+    return {
+        "Accept": "application/json",
+        "Accept-Encoding": "gzip, deflate",
+        "User-Agent": user_agent,
+        "X-Goog-Api-Client": user_agent,
```
frankyn, line 598: This is completely replacing x-goog-api-client; is it possible to keep the existing x-goog-api-client header value while adding `gcloud-python/`?

frankyn (changes requested):

google/cloud/storage/_helpers.py
Breathtender: Hmm, I'm not necessarily sure it is. This helper alone isn't really modifying any headers, as much as it's just setting the default. I think the thing to here though would be in the two places this is called, rearrange the order of the header dicts, so that the default one comes first which means anything specified in the other header will override these.

Breathtender: Ohhh although perhaps somewhere later this would override anything happening. I admittedly don't know much about X-Goog-Api-Client, I just know that @andrewsg suggested we set that as well in this change.

frankyn: I see you started adding tests, I think that may clarify if we need to care about information passed through x-goog-api-client vs. User-Agent.

Breathtender:
Just sharing an update on here since this is taking a good bit longer than expected. The change seems to be working and I am in the process of going through and updating tests. Most of them are able to be addressed with just adding the default headers in to the expected case. I've had to make some adjustments with content type, to ensure changing content type appropriately updates headers, and I'm currently working with some weirdness in terms of how content type and headers are updated with multipart uploads.

andrewsg (approved):
LGTM

google/cloud/storage/_helpers.py
```diff
@@ -581,3 +581,25 @@ def _api_core_retry_to_resumable_media_retry(retry, num_retries=None):
...
+    Args:
+        user_agent (str): The user-agent for requests.
+    Returns:
+        Dict: The headers to be used for the request.
+    """
+    return {
+        "Accept": "application/json",
+        "Accept-Encoding": "gzip, deflate",
+        "User-Agent": user_agent,
+        "x-goog-api-client": user_agent,
```
frankyn, line 602: @Breathtender does `user_agent` include language version as well?

frankyn (changes requested):
Overall LGTM, just have a nit question.

google/cloud/storage/_helpers.py
Breathtender: User agent is populated with client._connection.user_agent, which already has the nice well populated list of libraries and versions. Does that answer your question?

frankyn (approved):
LGTM, thanks @Breathtender!

PR #716, opened 2022-02-16
chore: add py.typed marker file for PEP 561

### LINKED ISSUE #621: Add a py.typed file for submodule
To be in accordance with [PEP 561](https://www.python.org/dev/peps/pep-0561/) we should add a py.typed file to google/cloud/storage/py.typed

The file can be empty but does need to be present for some type checkers.

cojenco:
- add py.typed marker file
- add mypy.ini file
- add mypy nox session
- ensure mypy passes

```
mypy google/cloud/storage
Success: no issues found in 16 source files
```
```
mypy -p google.cloud.storage --no-incremental
Success: no issues found in 16 source files
```
```
nox -s mypy
Success: no issues found in 16 source files
nox > Session mypy was successful.
```
Fixes #621 🦕

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,4 @@
+[mypy]
+python_version = 3.7
+namespace_packages = True
+ignore_missing_imports = True
--- a/google/__init__.py
+++ b/google/__init__.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 try:
-    import pkg_resources
+    import pkg_resources  # type: ignore
 
     pkg_resources.declare_namespace(__name__)
 except ImportError:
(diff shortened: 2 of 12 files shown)


mypy.ini
```diff
@@ -0,0 +1,4 @@
+[mypy]
+python_version = 3.7
```
Breathtender, line 2: I'm guessing this makes sense because 3.7 is our oldest supported version?

unforced (approved):

mypy.ini
cojenco: Yep this indicates the minimum python version supported

PR #741, opened 2022-03-25
feat: track invocation id for retry metrics

unforced:


google/cloud/storage/_helpers.py
```diff
@@ -599,7 +600,7 @@ def _get_default_headers(
         "Accept": "application/json",
         "Accept-Encoding": "gzip, deflate",
         "User-Agent": user_agent,
-        "x-goog-api-client": user_agent,
+        "x-goog-api-client": user_agent + " gccl-invocation-id/" + uuid4(),
```
frankyn, line 603: IIUC this uuid4() call will only occur once*;

Breathtender: From my understanding, that is true. Tests will confirm.

google/cloud/storage/_http.py
```diff
@@ -71,7 +71,7 @@ def api_request(self, *args, **kwargs):
         retry = kwargs.pop("retry", None)
         headers = {
             **kwargs.pop("headers", {}),
-            "x-goog-api-client": self._connection.user_agent + " gccl-invocation-id: " + uuid4(),
+            "x-goog-api-client": self._client_info.user_agent + " gccl-invocation-id: " + uuid4(),
```
frankyn: Two things:

1. it should be `gccl-invocation-id/UUID`
2. introduce an option to disable this invocation ID from being added

frankyn (changes requested):

google/cloud/storage/_http.py
Breathtender: Ah yep, 1 should be updated now.
And cool, will work on 2. Trying to figure out testing now, going to add my current quandary in another comment.

Breathtender:
Not sure the right way to test this at the moment.

As an example, this test fails
https://github.com/googleapis/python-storage/blob/main/tests/unit/test_client.py#L1725

With the failure occurring because they have different invocation ids in the headers

The same thing is causing failures in many different parts of the code

Especially in cases like this where it's using assert_called_once_with, I'm not even sure how I might extract the invocation id so as to insert it in the expected  headers

Breathtender:
Ok, old tests pass now. Still need to write new tests for the new feature add.

google/cloud/storage/_http.py
Breathtender: After discussion with @frankyn and @BenWhitehead, it was decided that for the moment, we will not be introducing an option to disable the invocation id.

tests/unit/test_client.py
```diff
@@ -1704,14 +1713,16 @@ def _download_blob_to_file_helper(
...
         blob._do_download = mock.Mock()
-
         file_obj = io.BytesIO()
-        if raw_download:
-            client.download_blob_to_file(
-                blob, file_obj, raw_download=True, **extra_kwargs
-            )
-        else:
-            client.download_blob_to_file(blob, file_obj, **extra_kwargs)
+        with patch.object(
```
frankyn, line 1717: @Breathtender can you test that the invocation id changes for each call of download_blob_to_file()?

Breathtender: Alright, check now; lines 1778-1795 in this file.

tests/unit/test_client.py
```diff
@@ -1775,6 +1775,24 @@ def test_download_blob_to_file_wo_chunks_w_raw(self):
     def test_download_blob_to_file_w_chunks_w_raw(self):
         self._download_blob_to_file_helper(use_chunks=True, raw_download=True)
 
+    def test_download_blob_to_file_multiple(self):
```
frankyn, line 1778: test_download_blob_have_different_uuid

frankyn (changes requested):

tests/unit/test_client.py
```diff
@@ -1760,6 +1775,24 @@ def test_download_blob_to_file_wo_chunks_w_raw(self):
...
+        blob = mock.create_autospec(Blob)
+        blob._encryption_key = None
+        blob._do_download = mock.Mock()
+        file_obj = io.BytesIO()
+        client.download_blob_to_file(blob, file_obj)
+        client.download_blob_to_file(blob, file_obj)
+
+        self.assertNotEqual(
+            blob._do_download.call_args_list[0][0][3]["X-Goog-API-Client"],
+            blob._do_download.call_args_list[1][0][3]["X-Goog-API-Client"],
```
frankyn, line 1793: just want to confirm the UUID is changing right?

Breathtender: Yes, that's what this is testing. 
```
(Pdb) blob._do_download.call_args_list[0][0][3]['X-Goog-API-Client']
'gcloud-python/2.2.1  gl-python/3.8.12 gax/2.7.1 gccl/2.2.1 gccl-invocation-id/90296518-9508-4a41-83f6-247eba7712a2'
(Pdb) blob._do_download.call_args_list[1][0][3]['X-Goog-API-Client']
'gcloud-python/2.2.1  gl-python/3.8.12 gax/2.7.1 gccl/2.2.1 gccl-invocation-id/b7fc4b3a-4cf4-48c1-9475-c95096336c83'
```

frankyn (approved):

tests/unit/test_client.py
frankyn: @Breathtender re-raising this; do you know if there's a separate UUID per resumable upload request? 

I didn't ask this in prior discussions and may have missed it.

PR #743, opened 2022-03-29
fix: Add setuptools as a dependency so pkg_resources is available

### LINKED ISSUE #740: add `setuptools` to dependencies
If you import this library in an isolated environment, it will fail with `ModuleNotFoundError: No module named 'pkg_resources'` due to the line referenced below. While `setuptools` is a commonly installed alongside Python, it's not a core module, and should be listed as a dependency so that package managers (e.g. pipenv, poetry, pdm) will pick up on it and install it.

https://github.com/googleapis/python-storage/blob/a8fcebb901abc297126cbb183eedbbd603ac5add/google/cloud/storage/_http.py#L19

andrewsg:
Fixes #740 🦕

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ dependencies = [
     "google-resumable-media >= 2.3.2",
     "requests >= 2.18.0, < 3.0.0dev",
     "protobuf",
+    "setuptools",
 ]
 extras = {}
 


unforced (approved):

parthea:
Superseded by #744

PR #744, opened 2022-03-30
fix(deps): drop pkg_resources

### LINKED ISSUE #740: add `setuptools` to dependencies

parthea:
Fixes https://github.com/googleapis/python-storage/issues/740 🦕

`google-cloud-core>=1.6.0` is already listed as a dependency in `setup.py`. We should remove code that uses `pkg_resources` to check `google-cloud-core>=1.6.0`.
https://github.com/googleapis/python-storage/blob/ad3911610a5178942c1ebdd5d8280ca70b890231/setup.py#L33

### DIFF
--- a/google/cloud/storage/_http.py
+++ b/google/cloud/storage/_http.py
@@ -15,18 +15,9 @@
 """Create / interact with Google Cloud Storage connections."""
 
 import functools
-import os
-import pkg_resources
 
 from google.cloud import _http
-
 from google.cloud.storage import __version__
-
-
-if os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE") == "true":  # pragma: NO COVER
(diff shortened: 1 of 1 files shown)


busunkim96 (approved):

cojenco (approved):
Thank you!

andrewsg:
Thanks a lot, very helpful!

PR #765, opened 2022-04-13
feat: add AbortIncompleteMultipartUpload lifecycle rule

### LINKED ISSUE #753: Support AbortIncompleteMultipartUpload LifecycleAction
This feature adds the following capabilities to the existing GCS Lifecycle APIs:
- A new type of LifecycleAction -  `AbortIncompleteMultipartUpload` will be added. 

**If there are code changes other than documentation to implement this feature, please add the following system tests:**

- Validate that “AbortIncompleteMultipartUpload” lifecycle action can be set via Bucket Create, Patch and Update operations.
- Bucket Get and List operations should return the correct state of the lifecycle configuration on the bucket which includes the rules with the new action if present.

**If no code changes are required, verify that there are tests proving strings are properly propagated.**

Internal: see [go/gcs-dpe-olm-mpu](http://go/gcs-dpe-olm-mpu) for more info.

Please wait for approval from myself before merging or releasing this change so we can coordinate around release dates.

andrewsg:
Fixes #753 🦕

### DIFF
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -377,6 +377,8 @@ class LifecycleRuleSetStorageClass(dict):
 class LifecycleRuleAbortIncompleteMultipartUpload(dict):
     """Map a rule aborting incomplete multipart uploads of matching items.
 
+    The "age" lifecycle condition is the only supported condition for this rule.
+
     :type kw: dict
     :params kw: arguments passed to :class:`LifecycleRuleConditions`.
     """
@@ -2343,6 +2345,9 @@ class Bucket(_PropertyMixin):
 
     def add_lifecycle_abort_incomplete_multipart_upload_rule(self, **kw):
         """Add a "abort incomplete multipart upload" rule to lifestyle rules.
+
+        Note that the "age" lifecycle condition is the only supported condition
+        for this rule.
 
         See https://cloud.google.com/storage/docs/lifecycle and
              https://cloud.google.com/storage/docs/json_api/v1/buckets
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -323,7 +323,7 @@ class LifecycleRuleDelete(dict):
     def __init__(self, **kw):
         conditions = LifecycleRuleConditions(**kw)
         rule = {"action": {"type": "Delete"}, "condition": dict(conditions)}
-        super(LifecycleRuleDelete, self).__init__(rule)
+        super().__init__(rule)
 
     @classmethod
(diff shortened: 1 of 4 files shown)


google/cloud/storage/bucket.py
```diff
@@ -374,6 +374,36 @@ def from_api_repr(cls, resource):
         return instance
 
 
+class LifecycleRuleAbortIncompleteMultipartUpload(dict):
+    """Map a rule aborting incomplete multipart uploads of matching items.
+
+    :type kw: dict
+    :params kw: arguments passed to :class:`LifecycleRuleConditions`.
```
cojenco, line 381: Could we add in the docstrings that the Age Lifecycle condition is the only allowed condition for this action?

tests/system/test_bucket.py
```diff
@@ -79,6 +83,9 @@ def test_bucket_lifecycle_rules(storage_client, buckets_to_delete):
             is_live=False,
             matches_storage_class=[constants.NEARLINE_STORAGE_CLASS],
         ),
+        LifecycleRuleAbortIncompleteMultipartUpload(
+            age=42,
+        ),
```
cojenco, line 88: Just noticed that the system tests never tested updating lifecycle rules (except for clearing them). I think we can add a section testing lifecycle actions being set via bucket patch/update.

cojenco (commented):
Looks good, thanks for working on this! I'll be adding the prefix/suffix support on top of this :)

tests/system/test_bucket.py
andrewsg: PTAL

tests/system/test_bucket.py
```diff
@@ -87,9 +94,21 @@ def test_bucket_lifecycle_rules(storage_client, buckets_to_delete):
...
     assert list(bucket.lifecycle_rules) == expected_rules
 
+    # Test modifying lifecycle rules
+    expected_rules[0] = LifecycleRuleDelete(age=30)
+    rules = list(bucket.lifecycle_rules)
+    rules[0]["condition"] = {"age": 30}
+    bucket.lifecycle_rules = rules
+    bucket.patch()
+
+    bucket.reload()
```
cojenco, line 104: Curious do we need to reload the bucket here, same on L111? The patch call should return the bucket metadata? Otherwise, LGTM, thanks!

cojenco (approved):

tests/system/test_bucket.py
andrewsg: No, you're right, we should be able to skip that. I'll remove them and confirm it still works.

PR #784, opened 2022-05-06
chore: publish docs for missing modules

cojenco:
As part of the docs revamp, add missing modules FileIO and Retry to Sphinx docs / TOC tree


dandhlee (approved):
Confirmed that the `retry_timeout` page is shorter, and new entry for `fileio` has been created with @cojenco. LGTM.

Could you file an issue / buganizer bug to track the followup for this work? Apologies if one already exists!

cojenco:
Thanks @dandhlee! Filed b/232410558 for follow-up work.

PR #787, opened 2022-05-10
tests: move range read test

cojenco:
Move range read test to system testing. Skip conf test as the testbench is now aligned with XML and returns complete object checksum.

Range reads headers are returned differently in JSON vs XML. The JSON response omits the x-goog-hash header whereas the XML response returns checksum that covers the complete object content for range reads.

### DIFF
--- a/tests/conformance/test_conformance.py
+++ b/tests/conformance/test_conformance.py
@@ -85,19 +85,6 @@ def blob_download_as_bytes(client, _preconditions, **resources):
     assert stored_contents == data.encode("utf-8")
 
 
-def blob_download_as_bytes_w_range(client, _preconditions, **resources):
-    # Range read response headers are returned differently in JSON vs XML. JSON response omits the x-goog-hash header;
-    # XML response returns checksum that covers the complete object content for range reads.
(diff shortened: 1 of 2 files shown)


tests/conformance/test_conformance.py
```diff
@@ -86,6 +86,9 @@ def blob_download_as_bytes(client, _preconditions, **resources):
 
 
 def blob_download_as_bytes_w_range(client, _preconditions, **resources):
+    # Range read response headers are returned differently in JSON vs XML. JSON response omits the x-goog-hash header;
+    # XML response returns checksum that covers the complete object content for range reads.
+    # Skip test as the testbench is now aligned with XML and returns complete object checksum.
```
andrewsg, line 91: Comment unclear as discussed in chat

andrewsg (changes requested):

tests/conformance/test_conformance.py
cojenco: Done, thanks!

andrewsg (approved):
Thanks!

PR #790, opened 2022-05-13
chore: update sphinx arrangement to remove duplicate entries on cloudsite

cojenco:
As part of the docs revamp, remove duplicate entries in [python/docs/reference/storage](https://cloud.google.com/python/docs/reference/storage/latest/)

For example, module `Bucket` appears both in https://cloud.google.com/python/docs/reference/storage/latest/buckets and https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.bucket

Collaborated with dandhlee@ on rearranging the sphinx arrangement. This PR
- removes this extra entries while keeping the rest of the documentation intact
- revises sphinx tags that are incorrectly used as region tags
- clean up ACL docstrings

Verified that the documentation generated looks correct. Follow-up work tracked in internal doc.

Fixes internal b/232410558


unforced (approved):

dandhlee (approved):
Cloud RAD likely does not have support for `.. literalinclude` (I think I remember this was an issue in other libraries too) but we can deal with it later. LGTM! 🚀

PR #807, opened 2022-06-02
fix: rewrite object in CMEK enabled bucket

### LINKED ISSUE #806: Rewrite object receives 400 error when CMEK is enabled for a bucket
Tracking issue for a customer.

Unable to rewrite on top of an existing GCS blob in CMEK enabled bucket. Receives a 400 Bad Request with the following repro code. This looks very similar to #155 

#### Environment details

  - python 3.9
  - google-cloud-storage 2.3.0

#### Steps to reproduce

  1. Create a bucket with Customer-managed encryption keys (via Cloud KMS)
  2. Write blob to bucket
  3. Overwrite blob

#### Code example

```python
bucket_name = 'your-bucket-name'
blob_name1 = 'your-object-name'
blob_name1 = 'your-different-name'
client = storage.Client()
bucket = client.bucket(bucket_name)
blob1 = bucket.get_blob(blob_name1)
blob2 = bucket.get_blob(blob_name2)
blob2.rewrite(blob1)
```

Expected behavior: Success, blob2 is overwritten
Actual behavior: google.api_core.exceptions.BadRequest: 400 POST <...> Bad Cloud KMS crypto key <...>
As long as blob1 and blob2 are not equal and both exist, this issue happens.

#### Workaround
Either delete blob2 first, then overwrite. OR call `blob2 = bucket.blob(blob_name2)` instead.

Potential Fix:

* `blob.rewrite()` should only use `kms_key_name`  if it doesn't end with a version.

cojenco:
Allow rewrite on top of an existing blob in CMEK enabled bucket by ignoring kmsKeyName version resource ID in a rewrite request.

The Cloud Storage API expects the **`kmsKeyName` resource ID without version information**. If a Blob instance with [kmsKeyName version resource ID](https://cloud.google.com/kms/docs/resource-hierarchy#retrieve_resource_id) set is used to rewrite the object, then it can't be used in the rewrite request and the client should ignore it.

Note: A metadata `GET` on a blob includes the kmsKeyName version resource ID

Fixes #806

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -3576,7 +3576,15 @@ class Blob(_PropertyMixin):
         if source.generation:
             query_params["sourceGeneration"] = source.generation
 
-        if self.kms_key_name is not None:
+        # When a Customer Managed Encryption Key is used to encrypt Cloud Storage object
+        # at rest, object resource metadata will store the version of the Key Management
+        # Service cryptographic material. If a Blob instance with KMS Key metadata set is
+        # used to rewrite the object, then the existing kmsKeyName version
+        # value can't be used in the rewrite request and the client instead ignores it.
+        if (
+            self.kms_key_name is not None
+            and "cryptoKeyVersions" not in self.kms_key_name
+        ):
             query_params["destinationKmsKeyName"] = self.kms_key_name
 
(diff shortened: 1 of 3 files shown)


andrewsg (approved):
Thanks!

PR #817, opened 2022-06-15
fix: propagate blob.generation in Bucket.delete_blobs

### LINKED ISSUE #814: `Bucket.delete_blobs` drops 'generation'
Whilst iterating the list, `generation` is dropped in the call to `Bucket.delete_blob()` (if `Blob` is passed in it should be a passthrough) resulting in much gnashing of teeth...

https://github.com/googleapis/python-storage/blob/9e5329c7dbab2a8023a288cf276c38f435a72a30/google/cloud/storage/bucket.py#L1787-L1806

cojenco:
Fixes #814  🦕

### DIFF
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -1578,6 +1578,9 @@ class Bucket(_PropertyMixin):
                 ) % (self._MAX_OBJECTS_FOR_ITERATION,)
                 raise ValueError(message)
 
+            # Convert to a list of blob names as we want to delete the latest versions of each blob.
+            # Blob generations propagated from the list op may no longer represent the latest versions.
+            blobs = [blob.name for blob in blobs]
             # Ignore 404 errors on delete.
             self.delete_blobs(
                 blobs,
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -1787,11 +1787,16 @@ class Bucket(_PropertyMixin):
         for blob in blobs:
             try:
                 blob_name = blob
+                generation = None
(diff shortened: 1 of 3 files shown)


frankyn (approved):

google/cloud/storage/bucket.py
```diff
@@ -1787,11 +1790,16 @@ def delete_blobs(
         for blob in blobs:
             try:
                 blob_name = blob
+                generation = None
                 if not isinstance(blob_name, str):
                     blob_name = blob.name
+                    if blob.generation:
+                        generation = blob.generation
```
andrewsg, line 1797: I'm still uncertain whether we should change behavior in this way, though I do agree the current behavior is bad. If we do make this change, can we skip the "take the name from the blob and recreate it" runaround and just run delete on the blob objects directly? This seems unnecessarily convoluted.

andrewsg (changes requested):

google/cloud/storage/bucket.py
cojenco: I understand your concerns. Unfortunately, with the current setup, it will be difficult to bypass "take the name from the blob...". Reasons being (1) params `blobs` can be a list of `str` blob_name OR blob instances, (2) `bucket.delete_blob()` only takes `str` blob_name whereas `blob.delete()` requires a blob instance.

cojenco: As discussed, I'm wondering if adding a flag to the method will better ensure backwards compatibility, but I'm also concerned that it will make this even more convoluted >"<

cojenco:
superseded by #840

PR #833, opened 2022-08-04
chore: delete benchwrapper obselete and unmaintained

frankyn:
Project never made it out of prototype phase and we are moving away from this testing method in general.

If we do come back to it; this code will still be too old / unmaintained.


cojenco (approved):
sgtm, thanks Frank!

frankyn:
@cojenco it looks like kokoro is failing due to unrelated test failures. Is someone addressing this?

cojenco:
> @cojenco it looks like kokoro is failing due to unrelated test failures. Is someone addressing this?

@frankyn Yes I'm looking into refactoring the flaky tests, tracking in #835

PR #840, opened 2022-08-10
feat: enable delete_blobs() to preserve generation

### LINKED ISSUE #814: `Bucket.delete_blobs` drops 'generation'

cojenco:
This adds a flag `preserve_generation` to the method `bucket.delete_blobs()`
- allows preserving and propagating blob generations when set to True (default False)
- better ensures backwards compatibility with both `delete_blobs()` and `bucket.delete(force=True)`

Fixes #814

### DIFF
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -1698,6 +1698,7 @@ class Bucket(_PropertyMixin):
         blobs,
         on_error=None,
         client=None,
+        preserve_generation=False,
         timeout=_DEFAULT_TIMEOUT,
         if_generation_match=None,
         if_generation_not_match=None,
@@ -1724,6 +1725,11 @@ class Bucket(_PropertyMixin):
         :type client: :class:`~google.cloud.storage.client.Client`
         :param client: (Optional) The client to use.  If not passed, falls back
                        to the ``client`` stored on the current bucket.
+
+        :type preserve_generation: bool
+        :param preserve_generation: (Optional) Preserves blob generation if set to True. A list of
+                                    :class:`~google.cloud.storage.blob.Blob`-s is required to preserve generation.
+                                    Default: False.
 
         :type if_generation_match: list of long
         :param if_generation_match:
@@ -1787,11 +1793,20 @@ class Bucket(_PropertyMixin):
         for blob in blobs:
             try:
                 blob_name = blob
+                generation = None
                 if not isinstance(blob_name, str):
                     blob_name = blob.name
+                    generation = blob.generation if preserve_generation else None
+                else:
+                    if preserve_generation:
+                        raise ValueError(
+                            "A list of blob instances need to be passed in to preserve blob generations."
+                        )
+
                 self.delete_blob(
                     blob_name,
                     client=client,
+                    generation=generation,
                     if_generation_match=next(if_generation_match, None),
                     if_generation_not_match=next(if_generation_not_match, None),
                     if_metageneration_match=next(if_metageneration_match, None),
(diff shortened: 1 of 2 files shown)


andrewsg:
Looks good, but can we additionally update the docstring to mention that the default behavior is to drop the generation? That way preserve_generation makes sense in context.

cojenco:
> Looks good, but can we additionally update the docstring to mention that the default behavior is to drop the generation? That way preserve_generation makes sense in context.

Great idea, I've noted the default behavior in the docstring

google/cloud/storage/bucket.py
```diff
@@ -1709,6 +1710,9 @@ def delete_blobs(
 
         Uses :meth:`delete_blob` to delete each individual blob.
 
+        By default, the latest versions of the blobs are deleted. Set `preserve_generation`
```
andrewsg, line 1713: Technically it's the live version, not the latest version. May I suggest the wording, "By default, any generation information in the list of blobs is ignored, and the live versions of all blobs are deleted. Set `preserve_generation` to True if blob generation should instead be propagated from the list of blobs."

google/cloud/storage/bucket.py
```diff
@@ -1725,6 +1729,11 @@ def delete_blobs(
         :param client: (Optional) The client to use.  If not passed, falls back
                        to the ``client`` stored on the current bucket.
 
+        :type preserve_generation: bool
+        :param preserve_generation: (Optional) Preserves blob generation if set to True. A list of
```
andrewsg, line 1733: Suggested wording: "Deletes only the generation specified on the blob object, instead of the live version, if set to True. Only :class:`~google.cloud.storage.blob.Blob` objects can have their generation set in this way."

google/cloud/storage/bucket.py
```diff
@@ -1787,11 +1796,20 @@ def delete_blobs(
         for blob in blobs:
             try:
                 blob_name = blob
+                generation = None
                 if not isinstance(blob_name, str):
                     blob_name = blob.name
+                    generation = blob.generation if preserve_generation else None
+                else:
+                    if preserve_generation:
```
andrewsg, line 1804: In my opinion, we can do without this ValueError, since it's valid to pass a mixed list of blobs and strings and this will trigger if even one of the list is a string. Let me know if you disagree and we can discuss.

andrewsg (changes requested):

google/cloud/storage/bucket.py
cojenco: Ooooh I've overlooked the possibility of passing a mixed list of blobs and strings. Considering this use case, it would make sense to remove the ValueError. Also like your proposed wording above, along with that description, I think that's clear!

google/cloud/storage/bucket.py
cojenco: SGTM👍👍

Agreed "live version" works better here especially in the suggested wording, updating the docstring. Argh I've been seeing both terms floating around in the API docs, and I tried to align wording with the [objects delete doc](https://cloud.google.com/storage/docs/json_api/v1/objects/delete)

andrewsg: Yeah, I think we should standardize on live version everywhere if we can. If "latest version", people would expect the behavior to be that you can run this command on a blob twice, once to delete the latest version, and then once more to delete the version before that. But that's not how it works - you run the command on a blob twice and the first time it will delete the live version, and then there is no more live version, so the second time it won't delete anything. You'd need to, indeed, use this new flag to delete the older version deliberately.

andrewsg (approved):

Issue #388, opened 2021-03-04
Implement sliced object download in the Python Client library.


yaseenlotfi:
**Is your feature request related to a problem? Please describe.**
My use case is to download a single, large blob (~16GBs) into memory in a Python application. This happens as part of a startup process that currently takes 5min. The command line utility, `gsutil`, has a way to enable sliced downloads and only takes 30s (same machine+network). I would like to take advantage of this optimization in a Pythonic way.

 **Describe the solution you'd like**
Enable sliced downloads in the Python client library such as:
`blob.download_to_filename(..., sliced_downloads=True, max_components=16)`

This would match `gsutil` which copies the blob to the local filesystem. It would be great, however, if the blob could be downloaded into memory like:
`blob.download_as_bytes(..., sliced_downloads=True, max_components=16)`

 **Describe alternatives you've considered**
Knowing that `gsutil` can run the download concurrently, I tried using the `subprocess` module to call it. This doesn't work bc it will not run more than one process unlike calling it from the command line. It's also not great to run a shell command from a Python process because it assumes Cloud SDK is setup.

I've tried using `ChunkedDownloads` in conjunction with `multiprocessing` but I have not been able to get it to download chunks in parallel. There is also the additional overhead of dealing with the byte stream buffer, transport authentication, checksum/data validation, etc making  it non-trivial.

 **Additional context**
Since `gsutil` is a Python executable itself, I would imagine this could be implemented in the client library (ultimately making the same HTTP Range Requests).

The `gsutil` command I used on a GCE instance with 16 vCPUs:
`gsutil -o ‘GSUtil:parallel_thread_count=1’ -o ‘GSUtil:sliced_object_download_max_components=16’ cp gs://bucket/key /path/to/destination`

Open to existing solution I'm not aware of, either, but documentation is sparse on this topic.

andrewsg:
Thanks for your detailed request. I'll look into this.

yaseenlotfi:
I'm not sure - ran it on a GCE instance (e2-highmem-16). How can I specifically test the network saturation?

andrewsg:
Thanks, that should be enough info; if it was on a GCE instance then we can use that for analysis when we're tackling this feature.

yaseenlotfi:
[...]

PR #844, opened 2022-08-19
feat: Add "transfer_manager" module for concurrent uploads and downloads as a preview feature

### LINKED ISSUE #36: Support parallel operations for copying objects
It would be great if there were a way to do the equivalent of the following command from the python API:

`gsutil -m cp -r myfolder gs://mybucket/`

You can currently upload each file individually, but this is slow for folders with many files.

I've also submitted a bug to gsutil: https://github.com/GoogleCloudPlatform/gsutil/issues/492

andrewsg:
This module adds support for concurrent uploads and downloads of multiple files, and concurrent chunked downloads of a single file.

Transfer Manager is currently a preview feature. The API and behavior of the module may change depending on user feedback. Please freely let us know about your experiences with the feature (be they problems, suggestions, or general feedback) by opening an issue on Github or commenting on an existing issue on the topic.

Fixes #36 and #388 🦕


andrewsg:
Docstrings pending initial review

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,188 @@
...
+):
+    if download_kwargs is None:
+        download_kwargs = {}
+    # We must know the size of the object, and the generation.
+    if not blob.size or not blob.generation:
+        blob.reload()
+
+    def download_range_via_tempfile(blob, file_obj, start, end, download_kwargs):
+        tmp = tempfile.TemporaryFile()
+        blob.download_to_file(tmp, start=start, end=end, **download_kwargs)
```
cojenco, line 104: Default retries are applied for downloads... it seems like retries will behave as intended. Just thinking how if any multiple worker threads might impact retries for this new feature

cojenco (commented):
Couple questions, generally looking really good!

This is probably in the planning, but it'd be very helpful to add method and parameter descriptions in the docstring for the new module.

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,188 @@
...
+    concurrent.futures.wait(
+        futures, timeout=deadline, return_when=concurrent.futures.ALL_COMPLETED
+    )
+    for future in futures:
+        if not raise_exception:
+            exp = future.exception()
+            if exp:
+                results.append(exp)
+                continue
+        results.append(future.result())
```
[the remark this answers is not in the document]
andrewsg: It should be impossible because the call `concurrent.futures.wait(futures, timeout=deadline, return_when=concurrent.futures.ALL_COMPLETED)` will ensure that every single future is complete before getting to that stage.

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,188 @@
...
+    download_kwargs=None,
+    deadline=None,
+):
+    if download_kwargs is None:
+        download_kwargs = {}
+    # We must know the size of the object, and the generation.
+    if not blob.size or not blob.generation:
+        blob.reload()
+
+    def download_range_via_tempfile(blob, file_obj, start, end, download_kwargs):
```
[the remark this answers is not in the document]
andrewsg: Yep, that's cruft from an earlier draft. Good catch!

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,188 @@
...
+    skip_if_exists=False,
+    upload_kwargs=None,
+    max_workers=None,
+    deadline=None,
+    raise_exception=False,
+):
+    if upload_kwargs is None:
+        upload_kwargs = {}
+    if skip_if_exists:
+        upload_kwargs["if_not_generation_match"] = 0
```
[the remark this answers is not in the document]
andrewsg: Hmm... this needs significantly more careful thought from me. Thank you, I'll dig in deep here and cover it with a system test in a future draft.

google/cloud/storage/transfer_manager.py
andrewsg: Retries are still configurable using upload/download_kwargs and I expect them to work normally.

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,188 @@
...
+        if not raise_exception:
+            exp = future.exception()
+            if exp:
+                results.append(exp)
+                continue
+        results.append(future.result())
+    return results
+
+
+def download_chunks_concurrently_to_file(
```
frankyn, line 88: What is the guidance on when a customer should rely on `download_chunks_conccurrently_to_file` directly vs. not?

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,188 @@
...
+    upload_kwargs=None,
+    max_workers=None,
+    deadline=None,
+    raise_exception=False,
+):
+    file_blob_pairs = []
+
+    for filename in filenames:
+        path = root + filename
+        blob_name = prefix + filename
```
frankyn, line 151: I think you will want to make sure filenames don't accidentally fall outside of root similar to https://github.com/advisories/GHSA-c28r-hw5m-5gv3;

You'll want to protect destination root so a filename doesn't attempt to write data outside of the expected location. Right now, this code IIRC fall to the same issue linked above.

@BenWhitehead had originally brought this up originally (thanks Ben).

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,188 @@
...
+# limitations under the License.
+
+import concurrent.futures
+
+import tempfile
+
+
+def upload_many(
+    file_blob_pairs,
+    skip_if_exists=False,
```
frankyn, line 22: race conditions can occur with preconditions set; we may need to include an "ignore 412" option as well otherwise customers will need to figure out how to handle the failure case.

frankyn (changes requested):
Few comments, thanks @andrewsg

google/cloud/storage/transfer_manager.py
andrewsg: That's in the design doc under "Directory traversal" - of course, still open for debate. Let's take this discussion back to that doc if needed so it's all in one place.

google/cloud/storage/transfer_manager.py
andrewsg: Good question! The benefit is that it adds speed for large files in cases where one thread can't saturate the network bandwidth. And the drawback is that it's much more complex, and adds overhead esp. if chunk size is low. How to distill that into customer guidance? I'll have to think about it.

google/cloud/storage/transfer_manager.py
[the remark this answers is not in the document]
andrewsg: This is resolved and tested via system tests. You were spot on with your feedback here. Thanks!

google/cloud/storage/transfer_manager.py
andrewsg: This should be resolved with the latest version, which (as documented) ignores 412s if skip_if_exists and raise_exceptions are both True.

andrewsg:
@frankyn @cojenco PTAL

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,533 @@
...
+def download_many(
+    blob_file_pairs,
+    download_kwargs=None,
+    max_workers=None,
+    deadline=None,
+    raise_exception=False,
+):
+    """Download many blobs concurrently via a worker pool.
+
+    This function is a FEATURE PREVIEW: the API may change in a future version.
```
frankyn, line 131: Nit: "FEATURE PREVIEW" -> "PREVIEW FEATURE"

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,533 @@
...
+        passed into the download methods and is not validated by this function.
+
+    :type max_workers: int
+    :param max_workers:
+        The number of workers (effectively, the number of threads) to use in
+        the worker pool. Refer to concurrent.futures.ThreadPoolExecutor
+        documentation for details.
+
+    :type deadline: int
+    :param deadline:
```
frankyn, line 153: Is there any guidance on what to set this value to?

frankyn (commented):
Two comments; overall LGTM,

tests/unit/test_transfer_manager.py
```diff
@@ -0,0 +1,315 @@
...
+        blob.download_to_file = fake_download_to_file
+
+        file_obj = io.BytesIO()
+
+        transfer_manager.download_chunks_concurrently_to_file(
+            blob, file_obj, chunk_size=4, download_kwargs=DOWNLOAD_KWARGS
+        )
+
+        # Generation wasn't set, so reload should have been called.
+        blob.reload.assert_called_with()
```
cojenco, line 210: Nice test here!! Maybe we want to mention in the docstrings that `blob.reload()` is called if the generation or size is not set on the blob

cojenco (commented):
A few questions, thanks for the thorough tests!

google/cloud/storage/transfer_manager.py
andrewsg: Under normal operation it should be left blank. I'll add the explanation of why to the docstrings. Thanks.

google/cloud/storage/transfer_manager.py
andrewsg: Okay

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,533 @@
...
+    :type blob_file_pairs: List(Tuple('google.cloud.storage.blob.Blob', IOBase or str))
+    :param blob_file_pairs:
+        A list of tuples of blob and a file or filename. Each blob will be
+        downloaded to the corresponding blob by using blob.download_to_file() or
+        blob.download_to_filename() as appropriate.
+
+    :type download_kwargs: dict
+    :param download_kwargs:
+        A dictionary of keyword arguments to pass to the download method. Refer
+        to the documentation for blob.download_to_file() or
```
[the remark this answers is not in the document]
andrewsg: It was deprecated due to a product direction that we're no longer taking, so I'm drawing up plans to un-deprecate it in the near future.

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,533 @@
...
+    download_kwargs=None,
+    max_workers=None,
+    deadline=None,
+    raise_exception=False,
+):
+    """Download many blobs concurrently via a worker pool.
+
+    This function is a FEATURE PREVIEW: the API may change in a future version.
+
+    :type blob_file_pairs: List(Tuple('google.cloud.storage.blob.Blob', IOBase or str))
```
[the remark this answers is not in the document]
andrewsg: That's a tough one. I like the consistency of FROM x TO y in that order for each of them and I can't think of any strong use cases for uploading and then downloading the exact same sets of files/blobs, or vice versa.

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,533 @@
...
+                results.append(exp)
+                continue
+        results.append(future.result())
+    return results
+
+
+def download_chunks_concurrently_to_file(
+    blob,
+    file_obj,
+    chunk_size=200 * 1024 * 1024,
```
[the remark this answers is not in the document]
andrewsg: Good idea.

tests/unit/test_transfer_manager.py
andrewsg: Good call.

google/cloud/storage/transfer_manager.py
```diff
@@ -0,0 +1,533 @@
...
+        tmp.seek(0)
+        file_obj.write(tmp.read())
+        tmp.close()
+
+
+def upload_many_from_filenames(
+    bucket,
+    filenames,
+    root,
+    prefix="",
```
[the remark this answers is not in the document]
andrewsg: Good idea. I'll try that and see how it looks.

bwalsh:
Great to see this,  starting a new project that needs exactly this functionality.  Thanks.

tests/system/test_transfer_manager.py
```diff
@@ -0,0 +1,100 @@
...
+    blob.upload_from_filename(file_data["big"]["path"])
+    blobs_to_delete.append(blob)
+
+    blob.reload()
+    fp = tempfile.TemporaryFile()
+    result = transfer_manager.download_chunks_concurrently_to_file(
+        blob, fp, chunk_size=1024 * 1024
+    )
+    assert result is None
+    assert fp.tell() != 0
```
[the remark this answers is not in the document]
andrewsg: Great idea. This is done for multi-chunk concurrent downloads.

andrewsg:
@cojenco @frankyn PTAL

cojenco (approved):
LGTM, thanks Andrew!

frankyn (approved):

PR #868, opened 2022-09-20
fix: remove client side validations

### LINKED ISSUE #879: Remove client side validation for storage class inputs
Remove client side validation on restricting storage class input. This would require an update if we introduce more storage classes in the future (which is possible).

Instead fall back to server side errors, e.g.
```
"error": {
    "code": 400,
    "message": "Invalid storage class",
}
```

### LINKED ISSUE #880: Remove client side validation in HMAC Key
Remove client side validation on restricting HMAC key states. This would require an update if we introduce more settable states in the future (which is possible). Part of go/gcs-dpes-no-extra-validation

Instead allow for nonspecific states and fall back to server side errors, e.g. 400 Bad Request

cojenco:
Remove client-side validations and fall back to API errors per [internal review doc](https://docs.google.com/spreadsheets/d/1swPw-7U7P43pUtpayuCqLN775GZHjRriFIC9tZpd97Q/edit?resourcekey=0-jakUDSZDsK4KoaCZUqyl6Q#gid=479962285)

- included: errors raised with unknown type/value
  - STORAGE_CLASSES
  - HMAC _SETTABLE_STATES
- excluded: client-side validations that are reasonable and can avoid wasteful processing with API actions

Fixes #879
Fixes #880


andrewsg (approved):
Thanks for adding tests as discussed

PR #870, opened 2022-09-23
tests: ensure config changes are propagated in sys tests

cojenco:
It seems like [bucket config changes may not be propagated in time](https://cloud.google.com/storage/docs/json_api/v1/buckets/patch), resulting in flaky system tests. 

- refactor system tests to use preconfigured fixtures
- rearrange tests with appropriate wait and test order
- adjust max retries

### DIFF
--- a/tests/system/_helpers.py
+++ b/tests/system/_helpers.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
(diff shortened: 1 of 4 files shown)


andrewsg (approved):
Looks great, thank you!

PR #873, opened 2022-09-26
docs: add clarifications to read timeout

cojenco:
Clarify timeout definition and usage, especially on read timeout
- read timeout is the number of seconds the client will wait for the server to send a response.
In most cases, this is the maximum wait time before the server sends the first byte.
- connect timeout is the number of seconds to establish a connection to the server

Fixes internal b/238779055


docs/storage/retry_timeout.rst
```diff
@@ -12,22 +12,28 @@ Configuring Timeouts
...
+longer than expected or desired. By default, such methods are applied a
+default timeout of 60.0 seconds.
 
-- You can pass a single integer or float which functions as the timeout for the
-  entire request. E.g.:
+The python-storage client uses the timeout mechanics of the underlying
+``requests`` HTTP library. The connect timeout is the number of seconds
+to establish a connection to the server. The read timeout is the number
+of seconds the client will wait for the server to send a response.
+In 99.9% of cases, this is the maximum wait time before the server sends
```
andrewsg, line 22: I recommend we drop the 99.9% bit (rather unsubstantiated on the part of the requests docs) and instead say: "In most cases this is the maximum wait time before the server sends the first byte; please refer to the ``requests`` documentation (link) for details."

docs/storage/retry_timeout.rst
```diff
@@ -12,22 +12,28 @@ Configuring Timeouts
...
+the first byte.
+
+Rather than blocking your application code for that interval,you may choose to
```
andrewsg, line 25: I think the "Rather than blocking your application code for that interval" phrase from our previous version is confusing and we should leave it out.

andrewsg (changes requested):
Is this information accurate for non-media API calls as well as media calls? I remember there being some nuance there even though both of them boil down to the requests library.

docs/storage/retry_timeout.rst
cojenco: good call; updated!

andrewsg (approved):

PR #883, opened 2022-10-07
feat: make tests run against environments other than prod

rsaksham:


google/cloud/storage/client.py
```diff
@@ -1761,7 +1762,10 @@ def generate_signed_post_policy_v4(
         elif bucket_bound_hostname:
             url = f"{_bucket_bound_hostname_url(bucket_bound_hostname, scheme)}/"
         else:
-            url = f"https://storage.googleapis.com/{bucket_name}/"
+            api_access_endpoint = os.getenv(
+                "API_ENDPOINT_OVERRIDE", "https://storage.googleapis.com"
```
andrewsg, line 1766: I know we were having trouble with some of the signing tests; does this change make any tests pass that weren't previously passing?

andrewsg (changes requested):

google/cloud/storage/client.py
andrewsg: Let's skip this as discussed because the if/elif is not covered by system tests yet.

tests/system/test_blob.py
```diff
@@ -97,6 +97,10 @@ def test_large_file_write_from_stream_w_failed_checksum(
     assert not blob.exists()
 
 
+@pytest.mark.skipif(
+    "API_ENDPOINT_OVERRIDE" in os.environ,
```
andrewsg, line 101: Replace with boolean constant at top of file as discussed, thank you for the idea.

andrewsg (changes requested):

noxfile.py
```diff
@@ -151,9 +151,17 @@ def system(session):
 
     # Run py.test against the system tests.
     if system_test_exists:
-        session.run("py.test", "--quiet", system_test_path, *session.posargs)
+        session.run(
+            "py.test", "--quiet", "--reruns=2", system_test_path, *session.posargs
```
andrewsg, line 155: As discussed, let's limit reruns=2 to prepod

andrewsg (approved):

PR #901, opened 2022-11-10
feat: add support for use_auth_w_custom_endpoint

### LINKED ISSUE #895: Enable authentication when a non-default api_endpoint is set (just like the googleapis/nodejs-storage)
**Issue we're facing:**
After version 2.1.0, authorization bearer token is stopped being sent to the custom endpoints (non-default api_endpoint), we forcefully needed to downgrade to the version 2.0.0 to make things work on our end. We had other option like explicitly passing credentials while instantiating the storage client, but this change was needed to be done in all places wherever the client is instantiated, so we just went to downgrade to the version 2.0.0.

**Feature request:**
Since this change [feat: avoid authentication with storage emulator](https://github.com/googleapis/python-storage/pull/679/files#top) was intentional, can you please add a flag to enable the auth for custom endpoints something like in the googleapis/nodejs-storage? 

A flag named **useAuthWithCustomEndpoint** was added in the nodejs storage client library to support the auth with custom endpoints. Just fyi, here are the code changes related to that flag.
 [Allow authentication with custom endpoint](https://github.com/googleapis/nodejs-common/pull/698/files#top)
 [Add support for useAuthWithCustomEndpoint option](https://github.com/googleapis/nodejs-storage/pull/1645/files#top)

cojenco:
This adds a boolean flag `use_auth_w_custom_endpoint`  -  whether authentication is required with custom endpoints
- only used when a custom endpoint in set in conjunction
- **defaults to True** and uses authentication
- if set to False, bypasses authentication

For backwards compatibility
- by default, auth is bypassed when `STORAGE_EMULATOR_HOST` is set

Fixes #895  and internal b/258502260


google/cloud/storage/client.py
```diff
@@ -144,19 +151,23 @@ def __init__(
...
-                project = _get_environ_project()
-            if project is None:
-                no_project = True
-                project = "<none>"
+        # If a custom endpoint is set, the client checks for credentials
+        # or finds the default credentials based on the current environment.
+        # Authentication may be bypassed under certain conditions:
+        # (1) STORAGE_EMULATOR_HOST is set (for backwards compatibility), OR
+        # (2) use_auth_w_custom_endpoint is set to False.
+        if kw_args["api_endpoint"] is not None:
```
frankyn, line 159: I would recommend if the `api_endpoint` has a hostname like `.googleapis.com` that auth be used; otherwise it does require always supplying `use_auth_w_custom_endpoint` everytime. IIRC this was done in Node.js Storage by @shaffeeullah

cojenco: I see what you mean. It makes sense that we'd want to enforce auth on `.googleapis.com` endpoints. Is that the same case for TPC, Regional Endpoints and PSC? In other words, is it reasonable to bypass auth under those use cases?

If so, would it make more sense to only bypass auth when `STORAGE_EMULATOR_HOST` is set?

frankyn: IIRC we discussed this in Node: to rely on `.googleapis.com` to tell us to enable auth by default; otherwise it's left to the user and defaults to no auth;

cojenco: Got it, thanks Frank! I've also commented in the internal bug asking for more details on the TPC/Regional Endpoints/PSC use cases. In particular, will the users have an option to bypass auth in those scenarios?

@shaffeeullah @ddelgrosso1 In Node, can a user set `useAuthWithCustomEndpoint=false`  if the api_endpoint has a hostname like `.googleapis.com`

cojenco: Synced with Node and Go folks offline. We're doing something quite similar here. Updated the PR description for clarity
- `use_auth_w_custom_endpoint` is only used when a custom endpoint is set in conjunction (L159 and L141-143)
- auth is enforced if it's a default endpoint "https://storage.googleapis.com"
- no extra validations on the custom endpoints

Special note
- `use_auth_w_custom_endpoint` defaults to True in Python
- for backwards compatibility, however, auth is automatically bypassed if `STORAGE_EMULATOR_HOST` is set

google/cloud/storage/client.py
```diff
@@ -897,7 +907,7 @@ def create_bucket(
             project = self.project
 
         # Use no project if STORAGE_EMULATOR_HOST is set
-        if _BASE_STORAGE_URI not in _get_storage_host():
+        if _get_storage_host() != _DEFAULT_STORAGE_HOST:
```
andrewsg, line 910: Is this a change that will impact backwards compatibility?

andrewsg (approved):
Had one question, otherwise LGTM. Thanks!

google/cloud/storage/client.py
cojenco: This is a special case only suited for when `STORAGE_EMULATOR_HOST` is set. This fix ensures the methods operate the same for custom endpoints and storage default endpoint.

google/cloud/storage/client.py
frankyn: Thanks Cathy, it's a bit tricky; Thanks for updating documentation.

frankyn (approved):

cojenco:
Update: pending discussion with the GCS EngProd team to streamline usage and validate testing against non-prod environments

cojenco:
Synced offline with @rsaksham - prepod testing should not be affected by this change.

PR is ready to merge and release once freeze ends

PR #907, opened 2022-11-15
fix: implement closed property on fileio.py classes

### LINKED ISSUE #903: BlobWriter does not implement closed property
#### Environment details

  - OS type and version: Mac OS X 12.6
  - Python version: `python --version`: 3.9.0
  - pip version: `pip --version`: 22.2.0
  - `google-cloud-storage` version: `pip show google-cloud-storage`: 2.6.0

#### Steps to reproduce

Run the code example below, substituting a real bucket for `some-bucket`.

#### Code example

```python

client = storage.Client()
bucket = storage.Bucket.from_string("gs://some-bucket", client=client)
blob = bucket.blob('deleteme')
stream = blob.open(mode='w', ignore_flush=True)
stream.write("foobarbaz")
stream.close()
print(stream.closed) # --> False.
```

`stream.closed` should return True after `stream.close()` is called. I think it's probably returning False because BlobWriter is not calling `super.close()` and does not override `IOBase.closed`.

BlobReader may have a similar problem, I haven't checked.

andrewsg:
Replace `_checkClosed()` function with `closed` property on fileio.py classes. `_checkClosed()` will still work, as the inherited implementation uses `self.closed`.

Fixes #903 🦕

### DIFF
--- a/google/cloud/storage/fileio.py
+++ b/google/cloud/storage/fileio.py
@@ -211,9 +211,9 @@ class BlobReader(io.BufferedIOBase):
     def close(self):
         self._buffer.close()
 
-    def _checkClosed(self):
-        if self._buffer.closed:
-            raise ValueError("I/O operation on closed file.")
+    @property
+    def closed(self):
+        return self._buffer.closed
 
     def readable(self):
         return True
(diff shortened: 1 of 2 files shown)


cojenco (commented):
Thanks Andrew for working on this! 

Could we update the unit tests, maybe an extra line of assertion in `test_close` and `test_write`? Also, do we want to apply the same changes to BlobReader? Seems like it uses the same inherited implementation.

frankyn (approved):

PR #918, opened 2022-11-28
docs: Add samples to multithread branch

andrewsg:
The core question I want to ask in this review is: is the interface for download_many_to_path and upload_many_from_filenames too complex? And especially, are the parameter names as clear and simple as we can make them?

### DIFF
--- a/samples/snippets/snippets_test.py
+++ b/samples/snippets/snippets_test.py
@@ -630,7 +630,7 @@ def test_transfer_manager_snippets(test_bucket, capsys):
     BLOB_NAMES = [
         "test.txt",
         "test2.txt",
-        "/blobs/test.txt",
+        "blobs/test.txt",
         "blobs/nesteddir/test.txt",
     ]
     BIG_BLOB_NAME = "bigblob.txt"
@@ -642,18 +642,18 @@ def test_transfer_manager_snippets(test_bucket, capsys):
         # Create dirs and nested dirs
         for name in BLOB_NAMES:
             relpath = os.path.dirname(name)
-            os.makedirs("{}/{}".format(uploads, relpath), exist_ok=True)
+            os.makedirs(os.path.join(uploads, relpath), exist_ok=True)
 
         # Create files with nested dirs to exercise directory handling.
         for name in BLOB_NAMES:
-            with open("{}/{}".format(uploads, name), "w") as f:
+            with open(os.path.join(uploads, name), "w") as f:
                 f.write(name)
         # Also create one somewhat bigger file.
-        with open("{}/{}".format(uploads, BIG_BLOB_NAME), "wb") as f:
+        with open(os.path.join(uploads, BIG_BLOB_NAME), "wb") as f:
             f.write(TEST_DATA_24_BYTES * SIZE_MULTIPLIER)
 
         storage_transfer_manager.upload_many_blobs_with_transfer_manager(
-            test_bucket.name, ALL_NAMES, root="{}/".format(uploads)
+            test_bucket.name, ALL_NAMES, source_directory="{}/".format(uploads)
         )
         out, _ = capsys.readouterr()
 
@@ -662,7 +662,7 @@ def test_transfer_manager_snippets(test_bucket, capsys):
 
     with tempfile.TemporaryDirectory() as downloads:
         # First let's download the bigger file in chunks.
-        big_destination_path = "{}/chunkeddl.txt".format(downloads)
+        big_destination_path = os.path.join(downloads, "chunkeddl.txt")
         storage_transfer_manager.download_blob_chunks_concurrently_with_transfer_manager(
             test_bucket.name,
             BIG_BLOB_NAME,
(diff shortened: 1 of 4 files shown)


samples/snippets/storage_transfer_manager.py
```diff
@@ -0,0 +1,136 @@
...
+    # full path to the file. Be sure to include the trailing slash. Relative
+    # paths and absolute paths are both accepted. This string is *not* included
+    # in the name of the uploaded blob; it is only used to find the source
+    # files. An empty string means "the current working directory".
+    # root=""
+
+    storage_client = Client()
+    bucket = storage_client.bucket(bucket_name)
+
+    results = transfer_manager.upload_many_from_filenames(bucket, filenames, root=root)
```
danielduhh, line 80: Seems like two upload  use cases: 
1) upload directory to bucket
2) upload list of files to bucket

Is there an easy way for someone to do #1? An alternative here is just to expose the upload_many and download_many apis, and have the samples show how to do #1 and #2 for both use cases. This could help simplify things.

Also, `source_directory` could be an alt for "root", and `destination_directory` an alt for download_many "path_root"

samples/snippets/storage_transfer_manager.py
```diff
@@ -0,0 +1,136 @@
...
+
+    blob = bucket.blob(blob_name)
+
+    # Open the local file in binary write mode.
+    local_file = open(local_filename, "wb")
+
+    # Unlike other transfer manager functions, which handle multiple files and
+    # return exceptions in a list, this function will simply raise any exception
+    # it encounters, and has no return value.
+    transfer_manager.download_chunks_concurrently_to_file(
```
danielduhh, line 125: Thoughts on call this download_ file_obj  or something simpler? It's basically a "smart" version download that handles the range reads for you.

danielduhh (commented):
Just added a couple opinions on the api surface

samples/snippets/storage_transfer_manager.py
andrewsg: I struggled with that nomenclature for sure, and ended up on this which is quite verbose. I'd be happy to change it. My main goals is differentiation between it and other download methods, and consistency with the existing "to_file" or "to_filename" naming pattern. Over the years I've seen many types of software that try to do multithreaded downloads and they're all named awkwardly (like this) or by weird brand names.

"download_to_file_threaded" or "download_chunks_to_file" could work if we wanted to slim it down a touch.

samples/snippets/storage_transfer_manager.py
andrewsg: Agree on renaming root and path_root. Funny how things that make sense to me when I write them don't make as much sense when I read them again later.

I'll craft a separate sample for uploading a directory. At a minimum it's only one different line of code, but because of the nuances involved it will need some space for explanation.

samples/snippets/storage_transfer_manager.py
```diff
@@ -0,0 +1,136 @@
...
+        blob, local_file, chunk_size=chunk_size
+    )
+
+    # If we've gotten this far, it must have been successful.
+
+    number_of_chunks = -(blob.size // -chunk_size)  # Ceiling division
```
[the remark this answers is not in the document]
andrewsg: I would rather not use it as the result, because all of our download methods have a None response right now and I hope to change that to something uniformly useful (across all methods including the non-TM ones) by pulling from HTTP headers in the next major release.

samples/snippets/storage_transfer_manager.py
```diff
@@ -0,0 +1,136 @@
...
+    # in the name of the uploaded blob; it is only used to find the source
+    # files. An empty string means "the current working directory".
+    # root=""
+
+    storage_client = Client()
+    bucket = storage_client.bucket(bucket_name)
+
+    results = transfer_manager.upload_many_from_filenames(bucket, filenames, root=root)
+
+    for name, result in zip(filenames, results):
```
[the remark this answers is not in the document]
andrewsg: Prefer to leave it alone so that if uploads change to have a result later, it will be uniform across tm and non-tm, similar to downloads as discussed.

samples/snippets/storage_transfer_manager.py
andrewsg: Rename complete.

andrewsg: Added directory upload snippet; PTAL.

samples/snippets/snippets_test.py
```diff
@@ -609,7 +620,99 @@ def test_batch_request(test_bucket):
...
+        )
+        out, _ = capsys.readouterr()
+
+        for name in ALL_NAMES:
+            assert "Uploaded {}".format(name) in out
+
+    with tempfile.TemporaryDirectory() as downloads:
+        # First let's download the bigger file in chunks.
+        big_destination_path = os.path.join(downloads, "chunkeddl.txt")
+        storage_transfer_manager.download_blob_chunks_concurrently_with_transfer_manager(
```
frankyn, line 666: Adding a comment to see this test removed; along with sample

samples/snippets/storage_transfer_manager.py
```diff
@@ -0,0 +1,208 @@
...
+        # The results list is either `None` or an exception for each filename in
+        # the input list, in order.
+
+        if isinstance(result, Exception):
+            print("Failed to upload {} due to exception: {}".format(name, result))
+        else:
+            print("Uploaded {} to {}.".format(name, bucket.name))
+
+
+def upload_directory_with_transfer_manager(bucket_name, directory):
```
frankyn, line 62: In a follow-up PR please split these samples into respective files

frankyn (changes requested):

PR #1001, opened 2023-03-06
chore: update benchmarking for SSB onboarding

cojenco:
- support cloud monitoring to onboard SSB, as outlined in internal guide go/gcs-sdk-bench
- add range read using w1r3 (workload 2)
- update input parameters and output
- modularize benchmarking script to support multiple workloads


tests/perf/_perf_utils.py
```diff
@@ -0,0 +1,193 @@
...
+            else:
+                size = object_size
+
+            # If size is greater than the defined threshold, report in MiB/s, otherwise report in KiB/s.
+            if size >= SSB_SIZE_THRESHOLD_BYTES:
+                throughput = size / 1024 / 1024 / (elapsed_time_us / 1_000_000)
+            else:
+                throughput = size / 1024 / (elapsed_time_us / 1_000_000)
+
+            cloud_monitoring_output = (
```
danielduhh, line 146: Hi @BrennaEpp, could you please verify this?

tests/perf/_perf_utils.py
```diff
@@ -0,0 +1,217 @@
...
+            print(cloud_monitoring_output)
+
+
+def cleanup_directory_tree(directory):
+    """Clean up directory tree on disk."""
+    try:
+        shutil.rmtree(directory)
+    except Exception as e:
+        logging.exception(f"Caught an exception while deleting local directory\n {e}")
+    logging.info("Successfully removed local directory")
```
andrewsg, line 174: Won't this result in an INFO-level log of success even if the attempt is a failure as per the exception caught above?

tests/perf/profile_w1r3.py
```diff
@@ -0,0 +1,229 @@
...
+
+def WRITE(bucket, blob_name, checksum, size, args, **kwargs):
+    """Perform an upload and return latency."""
+    blob = bucket.blob(blob_name)
+    file_path = f"{os.getcwd()}/{uuid.uuid4().hex}"
+    # Create random file locally on disk
+    with open(file_path, "wb") as file_obj:
+        file_obj.write(os.urandom(size))
+
+    start_time = time.monotonic_ns()
```
andrewsg, line 38: Usually I just use time.time(), is that not sufficient for these cases?

cojenco: Mainly using this in performance profiling for accuracy. The monotonic clock is not affected by system clock updates, and `time.monotonic_ns()` avoids the precision loss caused by the `time.time()` float type

tests/perf/_perf_utils.py
cojenco: good catch; updated!

andrewsg (approved):
Thanks!

PR #1002, opened 2023-03-07
feat: Add multiprocessing and chunked downloading to transfer manager

andrewsg:


andrewsg:
System tests are pending

google/cloud/storage/transfer_manager.py
```diff
@@ -16,27 +16,70 @@
...
+            kwargs["worker_type"] = THREAD
+            kwargs["max_workers"] = threads
+            return func(*args, **kwargs)
+        else:
+            return func(*args, **kwargs)
+
+    return convert_threads_or_raise
+
+
+@_deprecate_threads_param
```
danielduhh, line 73: Are you planning to remove this for GA, or support it forever?

google/cloud/storage/transfer_manager.py
```diff
@@ -16,27 +16,70 @@
...
 def upload_many(
     file_blob_pairs,
     skip_if_exists=False,
     upload_kwargs=None,
-    threads=4,
+    threads=None,
     deadline=None,
     raise_exception=False,
+    worker_type=PROCESS,
+    max_workers=8,
```
danielduhh, line 82: nit: the default worker value could be a variable (similar to default_chunk_size)

google/cloud/storage/transfer_manager.py
```diff
@@ -65,14 +111,9 @@ def upload_many(
...
     :param threads:
-        The number of threads to use in the worker pool. This is passed to
-        `concurrent.futures.ThreadPoolExecutor` as the `max_worker`; refer
-        to standard library documentation for details.
-
-        The performance impact of this value depends on the use case, but
-        generally, smaller files benefit from more threads and larger files
-        don't benefit from more threads. Too many threads can slow operations,
-        especially with large files, due to contention over the Python GIL.
+        ***DEPRECATED*** Sets `worker_type` to THREAD and `max_workers` to the
```
danielduhh, line 114: IIUC, if someone sets a value for this, worker_type automatically changes to "THREAD"? The pydocs specifically mention threads as an efficient solution for smaller files, and the only way to write to file handlers -- but this mentions the param as deprecated. I think this could create some confusion.

Another suggestion is to fully support both (without the deprecation warning) but set WORKERS as the default, or plan to turn down threads altogether.

danielduhh (commented):
Nice! Just had a couple thoughts..

google/cloud/storage/transfer_manager.py
andrewsg: Planning to remove it for GA, hence the loud warning.

google/cloud/storage/transfer_manager.py
andrewsg: Well, the idea is to grandfather in people who are already using the `threads` keyword argument and ease them into the new system of specifying worker_type and max_workers. So I would encourage people using `threads` to instead use `worker_type` and `max_workers` with the same values. I will clarify the docstring here.

cojenco (commented):
It's awesome to see both process and thread workers being supported 🎉 Still making my way through the PR :) A few questions

andrewsg:
System test is deadlocking due to https://github.com/grpc/grpc/issues/31885, investigating workaround

google/cloud/storage/transfer_manager.py
```diff
@@ -495,7 +698,226 @@ def download_many_to_path(
...
+        and the default is a conservative number that should work okay in most
+        cases without consuming excessive resources.
+
+    :raises: :exc:`concurrent.futures.TimeoutError` if deadline is exceeded.
+    """
+
+    if download_kwargs is None:
+        download_kwargs = {}
+    forced_start = download_kwargs.pop("start", 0)
+    forced_end = download_kwargs.pop("end", None)
```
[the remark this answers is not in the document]
andrewsg: Nah, the "if forced_end" on that line will prevent it from tripping up here.

google/cloud/storage/transfer_manager.py
```diff
@@ -92,6 +133,40 @@ def upload_many(
...
+        recreated in each subprocess. The serialization of the Client object
+        for use in subprocesses is an approximation and may not capture every
+        detail of the Client object, especially if the Client was modified after
+        its initial creation or if `Client._http` was modified in any way.
+
+        THREAD worker types are observed to be relatively efficient for
+        operations with many small files, but not for operations with large
+        files. PROCESS workers are recommended for large file operations.
+
+        PROCESS workers do not support writing to file handlers. Please refer
```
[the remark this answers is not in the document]
andrewsg: Good point... but I hesitate to take it away given it's already part of the preview feature.

google/cloud/storage/transfer_manager.py
```diff
@@ -495,7 +702,226 @@ def download_many_to_path(
...
+        end = min(forced_end + 1, blob.size) if forced_end else blob.size
+        while cursor < end:
+            start = cursor
+            cursor = min(cursor + chunk_size, end)
+            futures.append(
+                executor.submit(
+                    _download_and_write_chunk_in_place,
+                    maybe_pickled_blob,
+                    filename,
+                    offset=(forced_start * -1),
```
cojenco, line 826: Could you help me understand how the offset is computed here?

google/cloud/storage/transfer_manager.py
```diff
@@ -495,7 +702,226 @@ def download_many_to_path(
...
+
+    In some environments, using this feature with mutiple processes will result
+    in faster downloads of large files.
+
+    Using this feature with multiple threads is unlikely to improve download
+    performance under normal circumstances due to Python interpreter threading
+    behavior. The default is therefore to use processes instead of threads.
+
+    Checksumming (md5 or crc32c) is not supported for chunked operations. Any
+    `checksum` parameter passed in to download_kwargs will be ignored.
```
cojenco, line 733: Good call out. Where are we ignoring the `checksum` parameter?

google/cloud/storage/transfer_manager.py
```diff
@@ -495,7 +702,226 @@ def download_many_to_path(
...
+    else:
+        blob = pickle.loads(maybe_pickled_blob)
+    with open(
+        filename, "rb+"
+    ) as f:  # Open in mixed read/write mode to avoid truncating or appending
+        f.seek(offset + start)
+        return blob.download_to_file(f, start=start, end=end, **download_kwargs)
+
+
+def _call_method_on_maybe_pickled_blob(
```
cojenco, line 857: Could we add docstring comments for posterity? There's a whole lot packed in this helper and how it's used :)

tests/system/test_transfer_manager.py
```diff
@@ -75,10 +92,100 @@ def test_upload_many_skip_if_exists(
...
+        transfer_manager.download_chunks_concurrently(
+            download_blob,
+            full_filename,
+            chunk_size=chunk_size,
+            deadline=DEADLINE,
+        )
+        with open(full_filename, "rb") as file_obj:
+            assert _base64_md5hash(file_obj) == source_file["hash"]
+
+        # Now test for case where last chunk is exactly 1 byte.
```
cojenco, line 153: Thanks for the thorough tests! I especially like this one; really helped check the cursor moving through the last byte.

cojenco (commented):
A few questions after a second pass; otherwise LGTM, thanks!

google/cloud/storage/transfer_manager.py
andrewsg: Explained this offline, but for posterity, checksum is not sent by the server for ranged reads, and chunked downloads are all ranged reads, so none of the responses include a checksum. As a result, the resumable media code silently ignores the checksum parameter.

google/cloud/storage/transfer_manager.py
andrewsg: Might be deleting this anyways if we decide ranged reads are out of scope. If not, I should document this entire section better as there are a couple of tricky spots.

andrewsg:
@cojenco @danielduhh responded to feedback, PTAL. Thanks!

danielduhh (approved):

ddelgrosso1 (approved):
From a functionality standpoint this LGTM.

PR #1029, opened 2023-05-02
docs: remove threads in transfer manager samples

### LINKED ISSUE #1028: docs: deprecate the use of the "threads" parameter in transfer manager samples
"threads" is deprecated. Processes are now available and furthermore are default. The analogous parameter is named max_workers.

Sample functions using transfer_manager (aside from the chunked download sample) should accept "processes" instead of "threads" in their signature. Also, the default should be "8" instead of "4".

When actually calling the function, they can be called with max_workers instead of threads, using that passed-in value.

MiaCY:
- Remove deprecated argument "threads" in transfer_manager samples
- Use "processes" in transfer_manager samples and set default number to be 8
- Call "max_workers" when calling transfer_manager functions
- Refactor transfer_manager sample tests accordingly

Fixes #1028  🦕

### DIFF
--- a/samples/snippets/storage_transfer_manager_upload_directory.py
+++ b/samples/snippets/storage_transfer_manager_upload_directory.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 
-def upload_directory_with_transfer_manager(bucket_name, source_directory, threads=4):
+def upload_directory_with_transfer_manager(bucket_name, source_directory, processes=8):
     """Upload every file in a directory, including all files in subdirectories.
 
     Each blob name is derived from the filename, not including the `directory`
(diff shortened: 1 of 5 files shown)


dandhlee:
fyi: `docs` keyword triggers a new release and will be included in the release notes - if this is intended then please revert my title change, otherwise I don't think it's necessary.

dandhlee (approved):

samples/snippets/storage_transfer_manager_upload_directory.py
```diff
@@ -30,12 +30,13 @@ def upload_directory_with_transfer_manager(bucket_name, source_directory, thread
     # working directory".
     # source_directory=""
 
-    # The number of threads to use for the operation. The performance impact of
-    # this value depends on the use case, but generally, smaller files benefit
-    # from more threads and larger files don't benefit from more threads. Too
-    # many threads can slow operations, especially with large files, due to
-    # contention over the Python GIL.
-    # threads=4
+    # The maximum number of worker processes that should be used to handle the
```
andrewsg, line 33: Since this sample doesn't mention the worker type constants, let's try omitting them.

"The maximum number of processes to use for the operation. The performance impact of this value depends on the use case, but smaller files usually benefit from a higher number of processes. Each additional process occupies some CPU and memory resources until finished."

andrewsg (changes requested):

samples/snippets/storage_transfer_manager_upload_directory.py
MiaCY: Omitted worker type in comments related to processes. PTAL.

andrewsg (approved):
Thanks!

andrewsg:
@dandhlee Thanks Dan. Since the samples are public-facing I think they could stand to be in the release notes. I'll merge with docs: for now and if the release notes look messy I'll edit them at launch.

PR #1038, opened 2023-05-10
chore: add option to delete benchmarking bucket

cojenco:
Update benchmarking script with an added option to delete the GCS bucket or not.

With this option, the current SSB workloads will use the same bucket and avoid deletion/recreation in each run.


andrewsg (approved):
This intentionally changes the default behavior from delete to no-delete, right? Then LGTM. Thanks!

cojenco:
> This intentionally changes the default behavior from delete to no-delete, right? Then LGTM. Thanks!

That's right, defaults to no-delete 👍

PR #1043, opened 2023-05-17
feat: allow exceptions to be included in batch responses

### LINKED ISSUE #793: Batch context manager throws only the last exception occurred and no response is returned

#### Environment details

  - MacOS 
  - Python version: `3.8.9`
  - pip version: `22.0.4`
  - `google-cloud-storage` version: `2.3.0`

#### Steps to reproduce

  1.  Send Batch delete request for non existent objects in a bucket
 

#### Code example

```python
 storage_client = storage.Client.from_service_account_json('key.json')
 bucket_obj_1 = storage_client.bucket("bucket")
 with storage_client.batch():
      bucket_obj_1.delete_blobs(blobs=["1.txt", "2.txt", "3.txt"])
```

#### Stack trace
```
  File "****", line 72, in <module>
    bucket_obj_1.delete_blobs(blobs=["1.txt", "2.txt", "3.txt"])
  File "****/lib/python3.8/site-packages/google/cloud/storage/batch.py", line 288, in __exit__
    self.finish()
  File "****/lib/python3.8/site-packages/google/cloud/storage/batch.py", line 274, in finish
    self._finish_futures(responses)
  File "****/lib/python3.8/site-packages/google/cloud/storage/batch.py", line 250, in _finish_futures
    raise exceptions.from_http_response(exception_args)
google.api_core.exceptions.NotFound: 404 BATCH contentid://None: No such object: bucket/3.txt

```

In the above example, the batch request has been made with 3 blobs. However, the exception message only shows details about the last error occurred and this is evident in the implementation of the Batch context manager as mentioned in the below code snippet.

<pre>
    def _finish_futures(self, responses):
        """Apply all the batch responses to the futures created.

        :type responses: list of (headers, payload) tuples.
        :param responses: List of headers and payloads from each response in
                          the batch.

        :raises: :class:`ValueError` if no requests have been deferred.
        """
        # If a bad status occurs, we track it, but don't raise an exception
        # until all futures have been populated.
        exception_args = None

        if len(self._target_objects) != len(responses):  # pragma: NO COVER
            raise ValueError("Expected a response for every request.")

        for target_object, subresponse in zip(self._target_objects, responses):
            if not 200 <= subresponse.status_code < 300:
                <b>exception_args = exception_args or subresponse</b>
            elif target_object is not None:
                try:
                    target_object._properties = subresponse.json()
                except ValueError:
                    target_object._properties = subresponse.content

        if exception_args is not None:
            raise exceptions.from_http_response(exception_args)
</pre>

Since, each blob exception need to be handled individually, it is becoming difficult to handle exceptions. Any suggestions on how to workaround this issue?
Also, is there a way to retrieve the response objects of individual blobs so that the caller can process and handle it accordingly

cojenco:
This adds a flag `raise_exception` to `batch.finish()`, which provides an option to include exceptions in batch responses in a non-backwards-incompatible way
- when set to False, adds exceptions to the list of return responses
- defaults to True, raises the final exception (current behavior)

The flag is also added to the Batch constructor and factory method `client.batch()`

Fixes #793 🦕

### DIFF
--- a/google/cloud/storage/batch.py
+++ b/google/cloud/storage/batch.py
@@ -228,7 +228,7 @@ class Batch(Connection):
         _, body = payload.split("\n\n", 1)
         return dict(multi._headers), body, timeout
 
-    def _finish_futures(self, responses, raise_exception):
+    def _finish_futures(self, responses, raise_exception=True):
         """Apply all the batch responses to the futures created.
 
         :type responses: list of (headers, payload) tuples.
--- a/google/cloud/storage/batch.py
+++ b/google/cloud/storage/batch.py
@@ -135,11 +135,11 @@ class Batch(Connection):
     :param client: The client to use for making connections.
 
     :type raise_exception: bool
-    :param raise_exception: (Optional) Defaults to True.
-    If True, instead of adding exceptions to the list of return responses,
-    the final exception will be raised. Note that exceptions are
-    unwrapped after all operations are complete in success or failure,
-    and only the last exception is raised.
+    :param raise_exception:
+        (Optional) Defaults to True. If True, instead of adding exceptions
+        to the list of return responses, the final exception will be raised.
+        Note that exceptions are unwrapped after all operations are complete
+        in success or failure, and only the last exception is raised.
     """
 
     _MAX_BATCH_SIZE = 1000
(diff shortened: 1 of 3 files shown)


google/cloud/storage/batch.py
```diff
@@ -219,24 +228,34 @@ def _prepare_batch_request(self):
         _, body = payload.split("\n\n", 1)
         return dict(multi._headers), body, timeout
 
-    def _finish_futures(self, responses):
+    def _finish_futures(self, responses, raise_exception):
```
andrewsg, line 231: This should default to true according to the docstring; it's important that it defaults to true so that anything that is already using it won't break when we push out the change.

andrewsg (changes requested):
Looks promising, but might need tests - even if coverage is clear for some reason we should still try to test the functionality.

google/cloud/storage/batch.py
cojenco: updated and added tests

andrewsg (approved):

PR #1047, opened 2023-05-23
feat: add gsutil_uri property on Blob

### LINKED ISSUE #896: missing uri property on Blob()
Can be constructed but would help to get the URI natively as a property of `Blob()` e.g. `gs://my_bucket/path/my_blob`
This is consistent with instantiating a blob using its URI: `my_bucket.from_string(uri)`

### LINKED ISSUE #15984: missing uri property on Blob()
Can be constructed but would help to get the URI natively as a property of `Blob()` e.g. `gs://my_bucket/path/my_blob`
This is consistent with instantiating a blob using its URI: `my_bucket.from_string(uri)`

shani-suthar:
Please review it if this request is acceptable. Thanks! @andrewsg

### Pull Request

- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-storage/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Fixes googleapis/google-cloud-python#15984  🦕


holtskinner (approved):

andrewsg:
Apologies for letting this sit for so long. When the PR was originally filed, we were beginning to work on support for universe domains internally, and that feature conflicted with this implementation. As non-default universe domains need to be supported going forward, I can't accept this PR as it is, but I will open a bug to track the work going forward and do my best to incorporate this idea into the final product. Thank you.

PR #1052, opened 2023-05-25
Chore: refactor client.download_blob_to_file

MiaCY:
- Create new method blob._prep_and_do_download
- Refactor business logic of client.download_blob_to_file into blob._prep_and_do_download
- Refactor blob.download_to_file, blob.download_to_filename, and blob.download_as_bytes to call blob._prep_and_do_download
- Refactor related blob and client unit tests


google/cloud/storage/blob.py
```diff
@@ -3932,6 +3932,159 @@ def open(
...
+        if_generation_match=None,
+        if_generation_not_match=None,
+        if_metageneration_match=None,
+        if_metageneration_not_match=None,
+        timeout=_DEFAULT_TIMEOUT,
+        checksum="md5",
+        retry=DEFAULT_RETRY,
+    ):
+        
+        """Download the contents of a blob object or blob URI into a file-like object.
```
andrewsg, line 3953: This isn't accurate (this function only accepts file_obj)

google/cloud/storage/client.py
```diff
@@ -48,15 +50,16 @@
 from google.cloud.storage.bucket import Bucket, _item_to_blob, _blobs_page_start
 from google.cloud.storage.blob import (
     Blob,
-    _get_encryption_headers,
+    # _get_encryption_headers,
```
andrewsg, line 53: Before merging, be sure to remove this comment and the above comments. In general, prefer to delete rather than to comment out - with git, it's easy to restore whatever was there earlier even if you delete it.

andrewsg (commented):
We may later have to tune download_to_filename() to put the business logic of opening the file and handling exceptions behind a private barrier, but to keep things simple let's save that for after this PR.

andrewsg (changes requested):
Other than the unresolved naming question, looks good!

google/cloud/storage/blob.py
```diff
@@ -957,7 +957,7 @@ def _do_download(
 
             This private method does not accept ConditionalRetryPolicy values
             because the information necessary to evaluate the policy is instead
-            evaluated in client.download_blob_to_file().
+            evaluated in blob._download_blob().
```
cojenco, line 960: Let's also update L907 with the new private method.

tests/unit/test_client.py
```diff
@@ -1652,7 +1661,7 @@ def test_download_blob_to_file_with_failure(self):
         grmp_response = InvalidResponse(raw_response)
         credentials = _make_credentials(project=project)
         client = self._make_one(credentials=credentials)
-        blob = mock.create_autospec(Blob)
+        blob = self._make_blob(name="blob_name", bucket="source_name")
```
cojenco, line 1664: Note that the bucket parameter here takes a Bucket object, instead of a string

tests/unit/test_client.py
```diff
@@ -1639,9 +1639,18 @@ def test_create_bucket_w_name_only(self):
...
+
+        blob = Blob(*args, **kw)
+
+        return blob
+
     def test_download_blob_to_file_with_failure(self):
         from google.resumable_media import InvalidResponse
-        from google.cloud.storage.blob import Blob
+
+        # from google.cloud.storage.blob import Blob
```
cojenco, line 1653: Remember to remove this before merging, thanks!

cojenco (commented):
The updated tests using the mock.patch.object context looks good 🙌 
Left a few comments. Let me know if you have any questions.

google/cloud/storage/blob.py
```diff
@@ -3932,6 +3932,159 @@ def open(
     :rtype: str or ``NoneType``
     """
 
+    def _download_blob(
```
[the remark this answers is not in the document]
MiaCY: renamed new private method to blob._prep_and_do_download

tests/unit/test_client.py
MiaCY: refactored to `bucket=None`

andrewsg (approved):
Good work!

PR #1053, opened 2023-05-26
fix: extend wait for bucket metadata consistency in system tests

andrewsg:


tests/system/conftest.py
```diff
@@ -165,12 +167,20 @@ def signing_bucket(storage_client, signing_bucket_name):
...
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
 def default_ebh_bucket_name():
-    return _helpers.unique_name("gcp-systest-default-ebh")
+    # Keep track of how many ebh buckets have been created so we can get a
+    # clean one each rerun. "unique_name" is unique per test iteration, not
+    # per test rerun.
+    global ebh_bucket_iteration
+    ebh_bucket_iteration += 1
```
m-strzelczyk, line 176: Why not simply use `uuid.uuid4().hex[:10]` to have unique names without tracking?

andrewsg: I figured if we need to examine the artifacts or their logs later, having them in order would make it easier to determine which one we're looking for.

cojenco (changes requested):
Reviewing the flaky tests related to this caching change, let's also apply `await_config_changes_propagate` or appropriate retries around bucket metadata patches. Otherwise looks good, thank you!

i.e. https://github.com/googleapis/python-storage/blob/main/tests/system/test_bucket.py#L693-L694
i.e. https://github.com/googleapis/python-storage/blob/main/tests/system/test_bucket.py#L921-L938

andrewsg:
> https://github.com/googleapis/python-storage/blob/main/tests/system/test_bucket.py#L693-L694

Surprisingly this error turned out to be related to the pytest-rerunfailures process and was not the root cause. It's fixed by changing the bucket to not be session-shared.

> https://github.com/googleapis/python-storage/blob/main/tests/system/test_bucket.py#L921-L938

I'll take a look at this one now. Thanks!

andrewsg:
PTAL @cojenco

cojenco (approved):

PR #1054, opened 2023-05-26
chore: update noxfile to use verbose output, py3.9 by default

andrewsg:


cojenco:
Thank you for updating this! We'll also need to change the .kokoro config files for the session to be found, i.e. https://github.com/googleapis/python-storage/blob/main/.kokoro/presubmit/system-3.8.cfg

PR #1063, opened 2023-06-21
Chore: refactor blob.upload_from_file

MiaCY:
- Create new method blob._prep_and_do_upload
- Refactor business logic of blob.upload_from_file into blob._prep_and_do_upload
- Refactor blob.upload_from_file and blob.upload_from_filename to call blob._prep_and_do_upload


google/cloud/storage/blob.py
```diff
@@ -2386,7 +2386,7 @@ def _do_upload(
 
         return response.json()
 
-    def upload_from_file(
+    def _prep_and_do_upload(
```
andrewsg, line 2389: Let's just call this _do_upload since there's no other method we're disambiguating with in this case.

andrewsg (changes requested):
Oh, this is a straightforward and minimal approach. I think this is a good idea. Consider reducing the size of the private method's docstring a bit, possibly just by saying "see upload_from_file docstring for more info" if needed. Compare it to the other major private methods we have - is the docstring way too complex compared to those or is it similar? If it's similar or you think it would be confusing to reduce the size, we can just leave it as is.

I'm going to have to make unrelated changes to this area of the code soonish, so on the week of the 10th of July, let's look at merging the upload and download refactors into the main branch so there are no conflicts.

google/cloud/storage/blob.py
andrewsg: Correction: looks like we already have one of those so this is fine as it is.

PR #1082, opened 2023-07-18
feature: add 'command' argument to private upload/download interface

MiaCY:
- Add 'command' argument to blob._prep_and_do_download, blob._prep_and_do_upload, and other methods called by them:

  - blob._do_upload
  - blob._do_multipart_upload
  - blob._do_resumable_upload
  - blob._initiate_resumable_upload
  - blob._get_upload_arguments
  - _helpers._get_default_headers

- Add doc string regarding 'command' argument to above methods.
- Refactor related blob tests accordingly.


google/cloud/storage/blob.py
```diff
@@ -1709,6 +1709,10 @@ def _get_upload_arguments(self, client, content_type):
         :type content_type: str
         :param content_type: Type of content being uploaded (or :data:`None`).
 
+        :type command: str
+        :param command: (Optional) Information that can be passed into default
```
andrewsg, line 1713: (Optional) Information about which interface for upload/download was used, to be included in the X-Goog-API-Client header for traffic analysis purposes. Please leave as None unless otherwise directed.

andrewsg (changes requested):

andrewsg (approved):

PR #1096, opened 2023-07-26
feature: utilize new private interface to pass 'command' through transfer manager

MiaCY:
Download:
- Create new method blob._handle_filename_and_download()
- Refactor business logic of blob.download_to_filename() into blob._handle_filename_and_download()
- Refactor blob.download_to_filename() to call blob._handle_filename_and_download()
- Refactor transfer_manager.download_many() to call  blob._handle_filename_and_download() or blob._prep_and_do_download()
- Add 'command' to download_kwargs of transfer_manager.download_many() and transfer_manager.download_chunks_concurrently()

Upload:
- Create new method blob._handle_filename_and_upload()
- Refactor business logic of blob.upload_from_filename() into blob._handle_filename_and_upload()
- Refactor blob.upload_from_filename() to call blob._handle_filename_and_upload()
- Refactor transfer_manager.upload_many() to call  blob._handle_filename_and_upload() or blob._prep_and _do_upload()
- Addd 'command' to upload_kwargs of transfer_manager.upload_many()

 Other:
- Refactor transfer manager unit tests accordingly
- Edit docstrings of related methods


google/cloud/storage/transfer_manager.py
```diff
@@ -244,21 +245,17 @@ def download_many(
     :type blob_file_pairs: List(Tuple('google.cloud.storage.blob.Blob', IOBase or str))
     :param blob_file_pairs:
         A list of tuples of blob and a file or filename. Each blob will be
-        downloaded to the corresponding blob by using blob.download_to_file() or
-        blob.download_to_filename() as appropriate.
+        downloaded to the corresponding blob by using blob._prep_and_do_download() or blob._handle_filename_and_download() as appropriate.
```
andrewsg, line 248: Hmm, this is the tricky part. We want to avoid clients having to peek at private methods if possible. It's a little awkward, but how about this:

"downloaded to the corresponding blob by using APIs identical to blob.download_to_file() or blob.download_to_filename() as appropriate"

Additionally keep using blob.download_to_filename() in the "note" below.

Then, don't change the download_kwargs section at all, because the arguments should be identical.

The intention here is to keep the users focused on public methods, and also to keep the private methods functionally identical to the public methods wherever possible.

google/cloud/storage/transfer_manager.py
```diff
@@ -453,9 +452,8 @@ def upload_many_from_filenames(
     :type upload_kwargs: dict
     :param upload_kwargs:
         A dictionary of keyword arguments to pass to the upload method. Refer
-        to the documentation for blob.upload_from_file() or
-        blob.upload_from_filename() for more information. The dict is directly
-        passed into the upload methods and is not validated by this function.
+        to the documentation for blob._prep_and_do_upload() or
```
andrewsg, line 455: Use similar language as above.

andrewsg (changes requested):

tests/unit/test_transfer_manager.py
```diff
@@ -242,15 +248,15 @@ def test_download_many_with_filenames():
...
+        blob_mock._handle_filename_and_download.return_value = FAKE_RESULT
 
     results = transfer_manager.download_many(
         BLOB_FILE_PAIRS,
         download_kwargs=DOWNLOAD_KWARGS,
         worker_type=transfer_manager.THREAD,
     )
     for (mock_blob, file) in BLOB_FILE_PAIRS:
-        mock_blob.download_to_filename.assert_any_call(file, **DOWNLOAD_KWARGS)
+        mock_blob._handle_filename_and_download.assert_any_call(file, **DOWNLOAD_KWARGS)
```
cojenco, line 259: Similar for the downloads tests... Instead of adding `"command": "tm.download_many"` to `DOWNLOAD_KWARGS`, assert against a newly created `EXPECTED_DOWNLOAD_KWARGS` that includes the command

tests/unit/test_transfer_manager.py
```diff
@@ -538,13 +544,13 @@ def test_download_chunks_concurrently():
             worker_type=transfer_manager.THREAD,
         )
     for x in range(MULTIPLE):
-        blob_mock.download_to_file.assert_any_call(
+        blob_mock._prep_and_do_download.assert_any_call(
             mock.ANY,
             **DOWNLOAD_KWARGS,
```
cojenco, line 549: Hmmmm I'd imagine `_prep_and_do_download()` here would be called with an overriden value of`"command": "tm.download_sharded"`. Let's take a look at this together:)

cojenco (commented):
Looks great, thanks Mia!

cojenco (approved):

andrewsg (approved):

PR #1110, opened 2023-08-16
chore: Add region tags for Transfer Manager samples

andrewsg:


samples/snippets/storage_transfer_manager_download_all_blobs.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# [START storage_transfer_manager_download_all_blobs]
```
cojenco, line 16: Just a nit, there will be extra blank lines wrapping the code snippet in sample browser with this line after/before the region tags.

cojenco (approved):
A question on wrapping blank lines, otherwise LGTM, thanks!

PR #1119, opened 2023-09-18
feat: add gccl-gcs-cmd field to X-Goog-API-Client header for Transfer Manager calls

andrewsg:


google/cloud/storage/_helpers.py
```diff
@@ -599,19 +599,32 @@ def _get_default_headers(
...
 
-    Args:
-        user_agent (str): The user-agent for requests.
-    Returns:
-        Dict: The headers to be used for the request.
+    :type user_agent: str
+    :param user_agent: The user-agent for requests.
+
+        :type command: str
+        :param command:
```
cojenco, line 610: nit: indentation?

another nit: any particular reason why this PR is tagged as a fix?

tests/unit/test_transfer_manager.py
```diff
@@ -769,7 +769,7 @@ def test_upload_chunks_concurrently_with_metadata_and_encryption():
             "Accept": "application/json",
             "Accept-Encoding": "gzip, deflate",
             "User-Agent": "agent",
-            "X-Goog-API-Client": "agent gccl-invocation-id/{}".format(invocation_id),
+            "X-Goog-API-Client": f"agent gccl-invocation-id/{invocation_id} gccl-gcs-cmd/tm.upload_sharded",
```
cojenco, line 772: 👍

cojenco (approved):
🎉🎉🎉

google/cloud/storage/_helpers.py
andrewsg: I wasn't sure how to tag it. It's not customer-facing so it's not a feat. But it seems bigger than a chore.

Fix is for point releases, feat is for minor releases. I think this is more of a point release thing, so fix. I could be convinced otherwise.

PR #1120, opened 2023-09-18
feat: support configurable retries in upload_chunks_concurrently

andrewsg:


google/cloud/storage/transfer_manager.py
```diff
@@ -1075,6 +1096,7 @@ def _upload_part(
         checksum=checksum,
         headers=headers,
     )
+    part._retry_strategy = _api_core_retry_to_resumable_media_retry(retry)
```
cojenco, line 1099: I think this would work, thanks for adding! Curious what is the idempotency for XML API multipart uploads? I see in the documentation that preconditions are not supported, but didn't see any remarks on idempotency.

cojenco (approved):

google/cloud/storage/transfer_manager.py
andrewsg: Part uploads and finalize are fully idempotent because they're associated with an upload ID.

Initialize is not fully idempotent but since an extra upload is empty and therefore not billed for storage, the consequences of occasionally getting an extra upload are near zero. I made the decision reliability in this case is better than quibbling over the details of the init call.

PR #1121, opened 2023-09-19
feat: add support for custom headers

cojenco:
Add support for custom headers on a per-client basis
Reference - https://cloud.google.com/storage/docs/troubleshooting#custom-headers 

The following sample adds a custom header to the requests made by the client
```

client = storage.Client(
    extra_headers={
        "custom-header": "custom-value"
    }
)
```

Fixes internal b/294091512 🦕


andrewsg (approved):
Would be nice to have integration tests as well but since we don't typically snoop on outbound requests in our integration tests, and since custom headers don't usually have an impact on API behavior that is easily observable at the level our tests run, it may be difficult. Consider creative ways to implement an integration test if any come to mind; otherwise, good to go. Thanks.

PR #1138, opened 2023-10-06
feat: add crc32c_checksum argument to download_chunks_concurrently

andrewsg:


google/cloud/storage/transfer_manager.py
```diff
@@ -1205,6 +1304,38 @@ def _get_pool_class_and_requirements(worker_type):
...
+
+            # Zero pad base_crc32c. To conserve memory, do so with only
+            # MAX_CRC32C_ZERO_ARRAY_SIZE at a time. Reuse the zeroes array where
+            # possible.
+            padded = 0
+            while padded < size:
+                desired_zeroes_size = min((size - padded), MAX_CRC32C_ZERO_ARRAY_SIZE)
+                # resize zeroes array, unless we already have the correct size.
+                zeroes = (
+                    bytes(desired_zeroes_size)
```
frankyn, line 1324: what if you did `zeroes[:desired_zeroes_size]`  instead creating a new block?

google/cloud/storage/transfer_manager.py
```diff
@@ -732,6 +751,8 @@ def download_chunks_concurrently(
     deadline=None,
     worker_type=PROCESS,
     max_workers=DEFAULT_MAX_WORKERS,
+    *,
```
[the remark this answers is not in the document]
andrewsg: This is special Python 3 syntax that marks everything after that point as "keyword argument only". The reason I do that for extra arguments is so that I don't have to worry about ordering or reordering them in future revisions. It doesn't need a name.

google/cloud/storage/transfer_manager.py
andrewsg: That's a good idea. I'll experiment with it.

tests/unit/test_transfer_manager.py
```diff
@@ -567,6 +566,34 @@ def test_download_chunks_concurrently():
...
+    blob_mock._prep_and_do_download.side_effect = write_to_file
+
+    with mock.patch("google.cloud.storage.transfer_manager.open", mock.mock_open()):
+        transfer_manager.download_chunks_concurrently(
+            blob_mock,
+            FILENAME,
+            chunk_size=CHUNK_SIZE,
+            download_kwargs=DOWNLOAD_KWARGS,
+            worker_type=transfer_manager.THREAD,
+            crc32c_checksum=True,
```
cojenco, line 593: Could we also add a case that tests crc32c_checksum raises DataCorruption?

andrewsg: Funny coincidence, I just realized from cover that that was missing and added it as you were typing this.

frankyn (approved):

tests/unit/test_transfer_manager.py
```diff
@@ -973,3 +1044,31 @@ def test__call_method_on_maybe_pickled_blob():
...
+
+    with mock.patch(
+        "google.cloud.storage.transfer_manager.open", mock.mock_open()
+    ) as open_mock:
+        wrapper = transfer_manager._ChecksummingSparseFileWrapper(FILENAME, 0, True)
+        wrapper.write(b"abcdefgh")
+        handle = open_mock()
+        handle.write.assert_called_with(b"abcdefgh")
+        wrapper.write(b"ijklmnop")
+        assert wrapper.crc == google_crc32c.value(b"abcdefghijklmnop")
```
cojenco, line 1073: 👍👍

cojenco (approved):

PR #1161, opened 2023-10-11
feat: add skip_if_exists to download_many

### LINKED ISSUE #990: add an option to sync in the new transfer manager
thanks a lot for the new Transfer manager, it was a long wait but worth it

it would be really nice if we can have an option for sync, it means if the file was already download, don't download it again

andrewsg:
Fixes #990 🦕

### DIFF
--- a/google/cloud/storage/transfer_manager.py
+++ b/google/cloud/storage/transfer_manager.py
@@ -758,7 +758,7 @@ def download_many_to_path(
         raise_exception=raise_exception,
         worker_type=worker_type,
         max_workers=max_workers,
-        skip_if_exists=skip_if_exists
+        skip_if_exists=skip_if_exists,
     )
 
 
--- a/google/cloud/storage/transfer_manager.py
+++ b/google/cloud/storage/transfer_manager.py
@@ -353,7 +353,7 @@ def download_many(
     :type skip_if_exists: bool
     :param skip_if_exists:
         Before downloading each blob, check if the file for the filename exists;
-        if it does, skip that blob. This only works for filenames.
+        if it does, skip that blob.
 
     :raises: :exc:`concurrent.futures.TimeoutError` if deadline is exceeded.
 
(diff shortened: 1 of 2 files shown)


google/cloud/storage/transfer_manager.py
```diff
@@ -696,6 +709,11 @@ def download_many_to_path(
         and the default is a conservative number that should work okay in most
         cases without consuming excessive resources.
 
+    :type skip_if_exists: bool
+    :param skip_if_exists:
+        Before downloading each blob, check if the file for the filename exists;
+        if it does, skip that blob. This only works for filenames.
```
cojenco, line 715: nit: probably can remove last sentence since this method only applies to file paths (str)

google/cloud/storage/transfer_manager.py
```diff
@@ -329,6 +331,11 @@ def download_many(
...
```
cojenco, line 337: I think it's worth calling out that this is referring to `blob_file_pairs`  - can only be a list of tuples of blob and a filename for this to work. 

Will leave it up to you if you want to add a ValueError validation here. What I'm thinking is that a user may miss this note and not know why downloads are still repeated when passing in file-like objects

cojenco (approved):

google/cloud/storage/transfer_manager.py
andrewsg: Open file-like objects definitionally represent extant files, so hopefully no user should expect this feature to affect them. I'll pass on the validation for now; mixed filename/file handler lists are possible (honestly they cause only trouble so I should have banned them, but too late for that now) and it's complex to look ahead and only raise ValueError if all of them are identical.

google/cloud/storage/transfer_manager.py
andrewsg: fixed

PR #1163, opened 2023-10-12
docs: update formatting and wording in transfer_manager docstrings

andrewsg:


google/cloud/storage/transfer_manager.py
```diff
@@ -785,10 +792,11 @@ def download_chunks_concurrently(
     performance under normal circumstances due to Python interpreter threading
     behavior. The default is therefore to use processes instead of threads.
 
+    :type bucket: :class:`google.cloud.storage.bucket.Bucket`
     :param bucket:
         The bucket which contains the blobs to be downloaded
```
cojenco, line 797: Seems like no bucket param for this method

cojenco (changes requested):

google/cloud/storage/transfer_manager.py
andrewsg: Wow, you're right. Good catch!!

cojenco (approved):
🎉 🎉

PR #1170, opened 2023-10-20
fix: Blob.from_string parse gcs uri with regex

### LINKED ISSUE #1107: Blob.from_string() should be able to parse all valid gcs uri
`Blob.from_string` is terminating parsing of object names at hash(`#`) character. 

## Test Code (minimalistic)
```python3
blob = storage.Blob.from_string('gs://my-bucket/my/problamatic/object#name')
print(blob.name)
# my/problamatic/object
```
## Expected
Object name should be `my/problamatic/object#name`
## Actual
Object name found is `my/problamatic/object`

## Temporary Workaround
```python3
import re
GS_PATTERN = re.compile(r"gs://(?P<bucket_name>[^/]+)/(?P<object_name>.+)")

m = GS_PATTERN.match('gs://my-bucket/my/problamatic/object#name')
blob = storage.Bucket(name=m.group('bucket_name'), client=None).blob(m.group('object_name'))
print(blob.name)
# my/problamatic/object#name
```

## Version
`google_cloud_storage-2.10.0-py2.py3-none-any.whl`

## Suggestion
`urlsplit` is not a good utility for parsing gcs uris. I have seen problems with `%` char, trailing `?`, ... etc with `urlsplit`. Most of those characters are valid in context of gcs object names.

https://github.com/googleapis/python-storage/blob/663edc5e803f327f9c4eba86a75206d2e46c8536/google/cloud/storage/blob.py#L406

cojenco:
Fixes #1107 🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -134,7 +134,9 @@ _DOWNLOAD_AS_STRING_DEPRECATED = (
     "Blob.download_as_string() is deprecated and will be removed in future. "
     "Use Blob.download_as_bytes() instead."
 )
-
+_GS_URL_REGEX_PATTERN = re.compile(
+    r"(?P<scheme>gs)://(?P<bucket_name>[a-z0-9_.-]+)/(?P<object_name>.+)"
+)
 
 _DEFAULT_CHUNKSIZE = 104857600  # 1024 * 1024 B * 100 = 100 MB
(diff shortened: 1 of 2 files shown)


andrewsg (approved):

Ark-kun:
This change should likely be highlighted as a breaking change in the storage library since the new code starts raising errors for strings that worked before.

PR #1176, opened 2023-11-04
fix: use native namespace to avoid pkg_resources warnings

### LINKED ISSUE #1000: DeprecationWarning for pkg_resources.declare_namespace usage in google.cloud
Starting from `setuptools==67.3.0`, it has started raising DeprecationWarning. Some libraries treat them as errors during testing.

See: https://setuptools.pypa.io/en/latest/history.html#v67-3-0.

```python
/lib/python3.11/site-packages/pkg_resources/__init__.py:2804: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google.cloud')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)
```

This could be fixed by switching to new layout as specified in [PEP 420](https://peps.python.org/pep-0420/), by dropping `__init__.py`.

See similar discussion in https://github.com/matplotlib/matplotlib/issues/25244 and https://github.com/pypa/setuptools/pull/3434#issuecomment-1435697632.

andrewsg:
Fixes googleapis/python-storage#1000 🦕


parthea:
For other `google-cloud-*` libraries, we'll follow the same change in https://github.com/googleapis/google-auth-library-python/pull/1205 which

parthea (commented):
Please also remove `namespace_packages=` in setup.py and add a packaging test similar to https://github.com/googleapis/google-auth-library-python/pull/1205/files

andrewsg:
@parthea Test will be added after dependencies are updated, see #1179 . Took care of the setup.py updates. PTAL

parthea (approved):

PR #1180, opened 2023-11-08
fix: support virtual_hosted_style and bucket_bound_hostname in v2 signing

### LINKED ISSUE #1031: generate_signed_url does not work for virtual_hosted_style=True

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/python-storage/issues
  - Search StackOverflow: https://stackoverflow.com/questions/tagged/google-cloud-platform+python

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS type and version: Arch Linux
  - Python version: Python 3.9.13
  - pip version: `pip 21.3.1 `
  - `google-cloud-storage` version: 2.3.0

#### Steps to reproduce

  1. Call `blob.generate_signed_url(..., virtual_hosted_style=True)`
  2. Use the generated URL

#### Code example

```python
blob = bucket.blob("example")
blob.generate_signed_url(virtual_hosted_style=True)
```

#### Stack trace
```xml
<Error>
<Code>SignatureDoesNotMatch</Code>
<Message>
The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.
</Message>
<StringToSign>
GET 1683172800 /bucket/object
</StringToSign>
</Error>
```

I believe this is due to that the generated URL is https://bucket.storage.googleapis.com/object so it signs `/object` instead of `/bucket/object`

#### Workaround
```python
out = blob.generate_signed_url()
out = out.replace(
    "https://storage.googleapis.com/{}/".format(bucket_name),
    "https://{}.storage.googleapis.com/".format(bucket_name),
)
```

cojenco:
Cloud Storage recommends using V4 over V2 signing. Moreover, `virtual_hosted_style` and `bucket_bound_hostname` were intended to work primarily for v4 signing. 

The main caveat being that v2 signing requires the resource path to include the bucket-name. This patch handles the special cases (non path-styles) for v2 signing

Fixes #1031 🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -578,7 +578,11 @@ class Blob(_PropertyMixin):
             resource = f"/{self.bucket.name}/{quoted_name}"
 
         if virtual_hosted_style or bucket_bound_hostname:
-            resource = f"/{quoted_name}"
+            if version == "v4":
+                resource = f"/{quoted_name}"
+            else:
+                # v2 signing requires the resource path to include the bucket name
+                resource = f"/{self.bucket.name}/{quoted_name}"
 
         if credentials is None:
             client = self._require_client(client)
@@ -600,7 +604,7 @@ class Blob(_PropertyMixin):
             else:
                 headers.update(encryption_headers)
 
-        return helper(
+        signed_url = helper(
             credentials,
             resource=resource,
             expiration=expiration,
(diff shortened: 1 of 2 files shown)


andrewsg:
As discussed, looks good, but before we merge and release let's take a look at what upcoming work in this code area will look like.

cojenco:
Closing as obsolete

PR #1188, opened 2023-11-27
feat: support object retention lock

### LINKED ISSUE #1173: Add support for object retention
Please implement the object retention feature. The following operations must be supported:

- Create a bucket with object retention enabled by using the `enableObjectRetention` bucket create boolean option at bucket create time
- Get the object retention enabled status of a bucket
- Create an object with retention enabled (on an retention enabled bucket) by populating the `retention` object at object create time (retention object has the fields `mode` which can be Unlocked or Locked, and `retainUntilTime` which is a timestamp)
- Update retention for an existing object with a patch operation, by passing the `overrideUnlockedRetention` boolean patch option
- Get the retention policy of an object
- Remove retention for an existing object the same way as above

Details in go/gcs-object-retention-client-request

You can use the [java implementation](https://github.com/googleapis/java-storage/pull/2277) as a reference.

The kokoro test project should be allowlisted shortly, let me know if you need access to an allowlisted project for your own testing. 

We're aiming to get this in by the November code freeze, let me know if you need any help.

cojenco:
This adds support for [object retention lock](https://cloud.google.com/storage/docs/object-lock)

Fixes #1173


JesseLovelace (approved):

google/cloud/storage/_helpers.py
```diff
@@ -326,12 +327,21 @@ def patch(
         :type retry: google.api_core.retry.Retry or google.cloud.storage.retry.ConditionalRetryPolicy
         :param retry:
             (Optional) How to retry the RPC. See: :ref:`configuring_retries`
+
+        :type override_unlocked_retention: bool
+        :param override_unlocked_retention:
+            (Optional) Applicable for objects that have an unlocked retention configuration.
```
andrewsg, line 333: I can't say I fully understand this explanation yet. Is it in line with other language docstrings?

google/cloud/storage/blob.py
```diff
@@ -4797,3 +4809,126 @@ def _add_query_parameters(base_url, name_value_pairs):
     query = parse_qsl(query)
     query.extend(name_value_pairs)
     return urlunsplit((scheme, netloc, path, urlencode(query), frag))
+
+
+class Retention(dict):
+    """Map an object's retention configuration.
+
+    :type blob: :class:`Blob`
```
andrewsg, line 4817: Do we use this model of a separate class that links to a Blob in this way for any other classes?

google/cloud/storage/_helpers.py
cojenco: I agree it's quite confusing. As part of the feature, a new boolean parameter `overrideUnlockedRetention` is added to bucket patch and update operations, as shown in the [official documentation](https://cloud.google.com/storage/docs/json_api/v1/objects/patch#parameters). 

Would it be more clear to update the docstrings to something along the lines of:
```
overrideUnlockedRetention must be set to true if the operation includes a
retention property that changes the mode from Unlocked to Locked, 
reduces the retainUntilTime, or removes the retention configuration from the object.
```

google/cloud/storage/blob.py
cojenco: Not to Blob, but we use this same model in [Bucket for `IAMConfiguration` ](https://github.com/googleapis/python-storage/blob/9e4d1d8b44e55a1a10032b844932a8f5140c68fe/google/cloud/storage/bucket.py#L2212-L2220)

cojenco: Reading through the prd and [client request](https://goto.google.com/gcs-object-retention-client-request), I think it makes more sense to use a dataclass for the retention object. IMO, it will also provide more flexibility for further expansions and changes (which I have a feeling there may be in the future) Thoughts?

cojenco: Discussed offline and we decided to move forward with this dataclass implementation that is align with Bucket IAMConfiguration. This also gives us flexibility to expand on the retention object if/when new fields are introduced.

PR #1207, opened 2024-01-03
chore: fix get RPO sample

tritone:
There was an extra line here that doesn't make sense. Fixing based on external user feedback.

Fixes b/318488404 🦕

### DIFF
--- a/samples/snippets/storage_get_rpo.py
+++ b/samples/snippets/storage_get_rpo.py
@@ -35,8 +35,6 @@ def get_rpo(bucket_name):
 
     storage_client = storage.Client()
     bucket = storage_client.bucket(bucket_name)
-
-    bucket.rpo = RPO_DEFAULT
     rpo = bucket.rpo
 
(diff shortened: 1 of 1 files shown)


samples/snippets/storage_get_rpo.py
```diff
@@ -35,8 +35,6 @@ def get_rpo(bucket_name):
 
     storage_client = storage.Client()
     bucket = storage_client.bucket(bucket_name)
```
cojenco, line 37:
```suggestion
    bucket = storage_client.get_bucket(bucket_name)
```

cojenco (commented):
Thanks @tritone  for the fix! We'll also need to update to calling `get_bucket()` to actually make a metadata GET.

samples/snippets/storage_get_rpo.py
tritone: changed; thanks!

cojenco (approved):
Re-running kokoro as the auth library fix just released, thank you!

PR #1211, opened 2024-01-18
samples: replace deprecated method

cojenco:
Replace deprecated method `download_as_string` in samples; related to internal bug b/320588834


samples/snippets/storage_download_into_memory.py
```diff
@@ -37,7 +37,7 @@ def download_blob_into_memory(bucket_name, blob_name):
     # any content from Google Cloud Storage. As we don't need additional data,
     # using `Bucket.blob` is preferred here.
     blob = bucket.blob(blob_name)
-    contents = blob.download_as_string()
+    contents = blob.download_as_bytes()
 
     print(
         "Downloaded storage object {} from bucket {} as the following string: {}.".format(
```
andrewsg, line 43: Let's update this to "the following bytestring" or "the following bytes object".

We need to verify this prints correctly even when the bytestring is non-ascii. The test should include an example where that is the case. Maybe we should also decode the bytestring before it is printed, assuming it is utf-8?

andrewsg (changes requested):

samples/snippets/storage_download_into_memory.py
cojenco: Good catch; updated sample with decoding

andrewsg (approved):

PR #1212, opened 2024-01-18
feat: support custom universe domains/TPC

andrewsg:


JesseLovelace (approved):
Thanks Andrew!

google/cloud/storage/client.py
```diff
@@ -135,32 +140,79 @@ def __init__(
...
+
+        if client_options and client_options.universe_domain:
+            self._universe_domain = client_options.universe_domain
+        else:
+            self._universe_domain = None
+
+        storage_emulator_override = _get_storage_emulator_override()
+        api_endpoint_override = _get_api_endpoint_override()
+
+        # Determine the api endpoint. The rules are as follows:
```
cojenco, line 159: Thanks for adding comments here. Could you share more about how/why this order is determined? 

What happens when a user specifies both `universe_domain` and `api_endpoint_override`? I read in one of (the many) docs that it should be possible for a user to specify the endpoint override in addition to the universe. I guess it’s alright as long as endpoint override takes precedence… is my understanding correct?

google/cloud/storage/_http.py
```diff
@@ -31,10 +37,11 @@ class Connection(_http.JSONConnection):
     :param client_info: (Optional) instance used to generate user agent.
 
     :type api_endpoint: str
-    :param api_endpoint: (Optional) api endpoint to use.
+    :param api_endpoint: (Optional) api endpoint to use. This may be a universe
+        domain if one was specified.
```
cojenco, line 41: Does this mean that a universe domain can also be set via `Connection.api_endpoint`, in addition to `client_options.universe_domain`?

google/cloud/storage/blob.py
```diff
@@ -565,25 +572,38 @@ def generate_signed_url(
...
 
         # If you are on Google Compute Engine, you can't generate a signed URL
         # using GCE service account.
         # See https://github.com/googleapis/google-auth-library-python/issues/50
         if virtual_hosted_style:
-            api_access_endpoint = f"https://{self.bucket.name}.storage.googleapis.com"
+            api_access_endpoint = _virtual_hosted_style_base_url(
+                api_access_endpoint, self.bucket.name
+            )
+            resource = f"/{quoted_name}"
```
cojenco, line 596: There are v2 signing changes put on hold for this. Do we want to integrate #1180 in this code area altogether? If not, then we can add it afterwards.

cojenco (commented):
Looks great, thanks Andrew! Just a few minor questions

google/cloud/storage/client.py
andrewsg: Yes, it's explicitly ok if `universe_domain` and `api_endpoint_override` are both enabled. The latter takes precedence and there is no special casing - `universe_domain` is still set for all purposes other than determining the endpoint.

The order was determined by reading the TPC design doc very carefully.

google/cloud/storage/_http.py
andrewsg: This docstring is confusing. I'll rewrite it. Basically, users should never interact with this directly, but universe_domain being set may affect how our system sets this parameter.

google/cloud/storage/blob.py
andrewsg: Good question, let's talk about in next meeting and go over the changes.

google/cloud/storage/blob.py
```diff
@@ -537,21 +543,22 @@ def generate_signed_url(
...
 
         :type bucket_bound_hostname: str
         :param bucket_bound_hostname:
-            (Optional) If passed, then construct the URL relative to the
-            bucket-bound hostname.  Value can be a bare or with scheme, e.g.,
-            'example.com' or 'http://example.com'.  See:
-            https://cloud.google.com/storage/docs/request-endpoints#cname
+            (Optional) If passed, then construct the URL relative to the bucket-bound hostname.
+            Value can be a bare or with scheme, e.g., 'example.com' or 'http://example.com'.
+            Incompatble with api_access_endpoint and virtual_hosted_style.
```
BrennaEpp, line 552: Nit: typo
```suggestion
            Incompatible with api_access_endpoint and virtual_hosted_style.
```

BrennaEpp (approved):
LGTM - thank you Andrew!

Please pull in https://github.com/googleapis/conformance-tests/pull/89 and make required changes to your testing code to test the new cases (this can be in a separate PR).

cojenco (approved):

PR #1243, opened 2024-03-22
fix: cancel upload when BlobWriter exits with exception

### LINKED ISSUE #1228: BlobWriter should abort multipart upload during exception handling
Hi 👋

When an error occurs during a multipart upload, `BlobWriter.__exit__` is called with exception context. When that is the case, the multipart upload should be aborted. Currently, `BlobWriter.__exit__` [blindly calls](https://github.com/python/cpython/blob/v3.12.2/Lib/_pyio.py#L499) `BlobWriter.close`, which in turn finishes (commits) the multipart upload. The user then ends up with a partial blob on gcs, whereas the multipart upload should have been aborted.

#### Steps to reproduce

- init a [`BlobWriter`](https://github.com/googleapis/python-storage/blob/v2.14.0/google/cloud/storage/fileio.py#L228)
- write some bytes that are to be uploaded
- raise an error that should result in aborting the upload

#### Code example

```python
with blob.open('wb') as fp:
    fp.write(b'first chunk')  # not yet uploaded
    fp.write(b'big chunk' * 1024 ** 8)  # uploaded
    raise ValueError('SIGTERM received')

# the new blob should not have been created on gcs
```

For reference: [`smart_open.s3.MultiPartWriter.__exit__`](https://github.com/piskvorky/smart_open/blob/v6.4.0/smart_open/s3.py#L970)

ddelange:

Fixes #1228

### DIFF
--- a/tests/system/test_fileio.py
+++ b/tests/system/test_fileio.py
@@ -14,6 +14,9 @@
 # limitations under the License.
 
 
+import pytest
+
+from google.cloud.storage.fileio import CHUNK_SIZE_MULTIPLE
 from .test_blob import _check_blob_hash
 
 
@@ -76,3 +79,40 @@ def test_blobwriter_and_blobreader_text_mode(
         assert text_data[:100] == reader.read(100)
         assert 0 == reader.seek(0)
         assert reader.read() == text_data
+
+
+def test_blobwriter_exit(
+    shared_bucket,
+    blobs_to_delete,
+    service_account,
+):
+    blob = shared_bucket.blob("NeverUploaded")
+
+    # no-op when nothing was uploaded yet
+    with pytest.raises(ValueError, match="SIGTERM received"):
+        with blob.open("wb") as writer:
+            writer.write(b"first chunk")  # not yet uploaded
+            raise ValueError("SIGTERM received")  # no upload to cancel in __exit__
+    # blob should not exist
+    assert not blob.exists()
+
+    # unhandled exceptions should cancel the upload
+    with pytest.raises(ValueError, match="SIGTERM received"):
+        with blob.open("wb", chunk_size=CHUNK_SIZE_MULTIPLE) as writer:
+            writer.write(b"first chunk")  # not yet uploaded
+            writer.write(bytes(CHUNK_SIZE_MULTIPLE))  # uploaded
+            raise ValueError("SIGTERM received")  # upload is cancelled in __exit__
+    # blob should not exist
+    assert not blob.exists()
+
+    # handled exceptions should not cancel the upload
+    with blob.open("wb", chunk_size=CHUNK_SIZE_MULTIPLE) as writer:
+        writer.write(b"first chunk")  # not yet uploaded
+        writer.write(bytes(CHUNK_SIZE_MULTIPLE))  # uploaded
+        try:
+            raise ValueError("This is fine")
+        except ValueError:
+            pass  # no exception context passed to __exit__
+    blobs_to_delete.append(blob)
+    # blob should have been uploaded
+    assert blob.exists()
--- a/README.rst
+++ b/README.rst
@@ -72,6 +72,12 @@ setup.py file. Applications which do not import directly from
 `google-resumable-media` can safely disregard this dependency. This backwards
 compatibility feature will be removed in a future major version update.
 
+Miscellaneous
+~~~~~~~~~~~~~
+
+- The BlobWriter class now attempts to terminate an ongoing resumable upload if
+  the writer exits with an exception.
+
 Quick Start
 -----------
 
--- a/docs/storage/exceptions.rst
+++ b/docs/storage/exceptions.rst
@@ -1,5 +1,5 @@
 Exceptions
-~~~~~~~~~
+~~~~~~~~~~
 
 .. automodule:: google.cloud.storage.exceptions
   :members:
--- a/google/cloud/storage/fileio.py
+++ b/google/cloud/storage/fileio.py
@@ -437,6 +437,19 @@ class BlobWriter(io.BufferedIOBase):
             self._upload_chunks_from_buffer(1)
         self._buffer.close()
 
+    def terminate(self):
+        """Cancel the ResumableUpload."""
+        if self._upload_and_transport:
+            upload, transport = self._upload_and_transport
+            transport.delete(upload.upload_url)
+        self._buffer.close()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is not None:
+            self.terminate()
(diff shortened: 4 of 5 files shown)


ddelange:
@andrewsg can you link me to the relevant documentation section(s) to update? do you have specific wording in mind?

andrewsg:
Hi @ddelange,

Thanks for your submission. I've retargeted the PR to the 3.0 branch. For safety the test runner has to be activated manually for external PRs, so I've started the tests running now.

Three things:

- Please add the following text to the README file:

```
Miscellaneous
~~~~~~~~~~~~~

- The BlobWriter class now attempts to terminate an ongoing resumable upload if
  the writer exits with an exception.
```
 
- Somehow a typo in the new docs file escaped the CI last round. Could you please add a single ~ character to the title underline in the file `docs/storage/exceptions.rst`, to clear the "title underline isn't long enough" warning? It isn't related to this PR, but it's too trivial to make a separate PR for if we can clean it up here.

- The system test has this error:

```
        # no-op when nothing was uploaded yet
        with pytest.raises(ValueError, match="SIGTERM received"):
            with blob.open("wt") as writer:
>               writer.write(b'first chunk')  # not yet uploaded
E               TypeError: write() argument must be str, not bytes
```

- I don't see it yet but there will probably be a coverage issue, as features have to be exercised by unit tests. Please add a unit test.

ddelange:
Hi @andrewsg 👋

I rebased my commit to the new target branch, and fixed points 1 2 and 3. Point 3 was my typo in the tests, causing them to exit early. With the test passing, point 4 should also be covered (literally :]).

andrewsg:
Thank you for doing all of that! Unfortunately, system tests exist alongside the unit tests but they aren't a replacement for unit tests in the coverage requirements for this package, so we still will need unit tests covering the lines mentioned for the PR to pass.

andrewsg:
Looks like there's another system test issue: 
```
shared_bucket = , blobs_to_delete = []
service_account = 

    def test_blobwriter_exit(
        shared_bucket,
        blobs_to_delete,
        service_account,
    ):
        blob = shared_bucket.blob("NeverUploaded")
    
        # no-op when nothing was uploaded yet
        with pytest.raises(ValueError, match="SIGTERM received"):
            with blob.open("wb") as writer:
                writer.write(b"first chunk")  # not yet uploaded
                raise ValueError("SIGTERM received")  # no upload to cancel in __exit__
        # blob should not exist
        assert not blob.exists()
    
        # unhandled exceptions should cancel the upload
        with pytest.raises(ValueError, match="SIGTERM received"):
            with blob.open("wb") as writer:
                writer.write(b"first chunk")  # not yet uploaded
>               writer.write(b"big chunk" * 1024 ** 8)  # uploaded
E               OverflowError: cannot fit 'int' into an index-sized integer
```

Let me know if you need help with running the system tests locally.

ddelange:
Thanks for the heads up. I'll write a new unit test case. Regarding the OverflowError: can you suggest a cleaner way to trigger the threshold where the first part is uploaded (multipart upload created)? Is there a certain amount of bytes stored in a constant I can import? To change the line to something like `b'0' * MIN_PART_UPLOAD_SIZE`?

tests/system/test_fileio.py
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 
+import pytest
+
```
ddelange, line 18:
```suggestion

from google.cloud.storage.fileio import CHUNK_SIZE_MULTIPLE
```

tests/system/test_fileio.py
```diff
@@ -76,3 +78,41 @@ def test_blobwriter_and_blobreader_text_mode(
...
+    with pytest.raises(ValueError, match="SIGTERM received"):
+        with blob.open("wb") as writer:
+            writer.write(b"first chunk")  # not yet uploaded
+            raise ValueError("SIGTERM received")  # no upload to cancel in __exit__
+    # blob should not exist
+    assert not blob.exists()
+
+    # unhandled exceptions should cancel the upload
+    with pytest.raises(ValueError, match="SIGTERM received"):
+        with blob.open("wb") as writer:
```
ddelange, line 101:
```suggestion
        with blob.open("wb", chunksize=CHUNK_SIZE_MULTIPLE) as writer:
```

tests/system/test_fileio.py
```diff
@@ -76,3 +78,41 @@ def test_blobwriter_and_blobreader_text_mode(
...
+            writer.write(b"first chunk")  # not yet uploaded
+            writer.write(b"big chunk" * 1024 ** 8)  # uploaded
```
ddelange, line 103:
```suggestion
            writer.write(b"0" * CHUNK_SIZE_MULTIPLE)  # uploaded
```

ddelange (commented):
@andrewsg how about this?

andrewsg:
Hi ddlange, yes, overriding the chunk size makes sense to me. Go for it.

ddelange:
@andrewsg ready for review again, now with unit tests

andrewsg:
Thank you. Looks like there are still test failures with the system test and with the unit test. Can you take a look? Run "nox -s unit-3.11" (or whatever version) to run the unit tests, subsequently "nox -s cover" to check coverage, and "nox -s blacken lint" to get lint to pass as well.

ddelange:
@andrewsg thanks for the heads up.
- fixed system test typo `chunksize` -> `chunk_size`
- fixed unit test typo `initiate_resumable_upload` -> `_initiate_resumable_upload` in my and multiple other unit tests in `test_fileio.py` (I guess it was a typo which propagated through copy-paste)

unit tests are green on my machine, but `nox -s cover` still shows `442-445, 448-451` missing, which are the complete functions this PR adds. how can that be? the unit tests I wrote should cover all lines I added 🤔

andrewsg:
Thank you for catching the unit test issue in the other tests. It would have been avoided if I'd used the "spec" features of the mocking library properly; my mistake. I appreciate you fixing it in the entire file.

I don't know why coverage doesn't work on your machine - it's finicky - but it is clear in CI so all is good with the unit tests and coverage.

The system test is still failing. The part of the test that was causing an error previously due to expanding too aggressively, which you fixed by using the chunk size as the size to write, is repeated further down in the test.

ddelange:
> which you fixed by using the chunk size as the size to write, is repeated further down in the test

thanks for the catch, fixed now!

ddelange:
@andrewsg can you run CI again?

andrewsg (approved):

andrewsg:
@ddelange Checks pass. Thank you for your contribution, and again thanks for the heads-up about the testing issue.

PR #1249, opened 2024-04-03
fix: Reload properties when creating `Blob.from_string()`

### LINKED ISSUE #1248: `Blob.content_type` is `None` when created `from_string()`
### Example Code

```py
gcs_uri = "gs://github-repo/rag/intro_multimodal_rag/google-10k-sample-14pages.pdf"
gcs_client = storage.Client()
blob = storage.Blob.from_string(gcs_uri, gcs_client)

print(blob.content_type) # Outputs `None`
```

This isn't the case when loading blob from a local file or listing from the bucket.

holtskinner:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-storage/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Fixes #1248 🦕

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -416,7 +416,8 @@ class Blob(_PropertyMixin):
             raise ValueError("URI pattern must be gs://bucket/object")
         bucket = Bucket(client, name=match.group("bucket_name"))
         blob = cls(match.group("object_name"), bucket)
-        blob.reload(client=client)
+        if client:
+            blob.reload(client=client)
         return blob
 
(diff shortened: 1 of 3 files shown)


andrewsg:
Sorry, can't accept at present due to rational in linked bug. Thank you for your interest, though. Feel free to reach out to discuss usage to avoid this issue in another way.

PR #1257, opened 2024-04-12
feat: introduce OpenTelemetry Tracing decorators

cojenco:
part 1 introduce OTel tracing module and decorators

Fixes internal b/330754859

### DIFF
--- a/google/cloud/storage/_http.py
+++ b/google/cloud/storage/_http.py
@@ -18,6 +18,7 @@ import functools
 from google.cloud import _http
 from google.cloud.storage import __version__
 from google.cloud.storage import _helpers
+from google.cloud.storage._opentelemetry_tracing import create_span
 
 
 class Connection(_http.JSONConnection):
@@ -65,14 +66,25 @@ class Connection(_http.JSONConnection):
 
     def api_request(self, *args, **kwargs):
         retry = kwargs.pop("retry", None)
-        kwargs["extra_api_info"] = _helpers._get_invocation_id()
+        invocation_id = _helpers._get_invocation_id()
+        kwargs["extra_api_info"] = invocation_id
+        span_attributes = {
+            "gccl-invocation-id": invocation_id,
+        }
         call = functools.partial(super(Connection, self).api_request, *args, **kwargs)
-        if retry:
-            # If this is a ConditionalRetryPolicy, check conditions.
-            try:
-                retry = retry.get_retry_policy_if_conditions_met(**kwargs)
-            except AttributeError:  # This is not a ConditionalRetryPolicy.
-                pass
+        with create_span(
+            name="Storage.Connection.api_request",
+            attributes=span_attributes,
+            client=self._client,
+            api_request=kwargs,
+            retry=retry,
+        ):
             if retry:
-                call = retry(call)
-        return call()
+                # If this is a ConditionalRetryPolicy, check conditions.
+                try:
+                    retry = retry.get_retry_policy_if_conditions_met(**kwargs)
+                except AttributeError:  # This is not a ConditionalRetryPolicy.
+                    pass
+                if retry:
+                    call = retry(call)
+            return call()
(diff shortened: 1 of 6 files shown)


google/cloud/storage/_http.py
```diff
@@ -65,14 +66,25 @@ def __init__(self, client, client_info=None, api_endpoint=None):
...
+        }
         call = functools.partial(super(Connection, self).api_request, *args, **kwargs)
-        if retry:
-            # If this is a ConditionalRetryPolicy, check conditions.
-            try:
-                retry = retry.get_retry_policy_if_conditions_met(**kwargs)
-            except AttributeError:  # This is not a ConditionalRetryPolicy.
-                pass
+        with create_span(
+            name="Storage.Connection.api_request",
```
andrewsg, line 76: How did you arrive at this name layout?

google/cloud/storage/_opentelemetry_tracing.py
```diff
@@ -0,0 +1,101 @@
...
+    ENABLE_OTEL_TRACES_ENV_VAR, _DEFAULT_ENABLE_OTEL_TRACES_VALUE
+)
+logger = logging.getLogger(__name__)
+
+try:
+    from opentelemetry import trace
+
+    HAS_OPENTELEMETRY = True
+
+    logger.info(f"HAS_OPENTELEMETRY is {HAS_OPENTELEMETRY}")
```
andrewsg, line 27: This looks a little terse, is it intended to be user-facing? Seems like we could put it outside of the try/except if so.

google/cloud/storage/_opentelemetry_tracing.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+@contextmanager
+def create_span(
+    name, attributes=None, client=None, api_request=None, retry=None, **kwargs
+):
+    """Creates a context manager for a new span and set it as the current span
+    in the configured tracer. If no configuration exists yields None."""
+    if not HAS_OPENTELEMETRY or not enable_otel_traces:
+        yield None
```
andrewsg, line 52: Out of curiosity, what happens if we omit the yield statement and just return?

google/cloud/storage/_opentelemetry_tracing.py
```diff
@@ -0,0 +1,101 @@
...
+    # Yield new span.
+    with tracer.start_as_current_span(
+        name=name, kind=trace.SpanKind.CLIENT, attributes=final_attributes
+    ) as span:
+        try:
+            yield span
+        except api_exceptions.GoogleAPICallError as error:
+            span.set_status(trace.Status(trace.StatusCode.ERROR))
+            span.record_exception(error)
+            raise error
```
andrewsg, line 66: I believe if we "raise" without the "error" it will preserve the traceback without including this bit, which might be more legible.

google/cloud/storage/_http.py
cojenco: This follows the span name convention of ​​$package.$service/$method. Basically, the span name should identify the work represented by the span, in this case, it should include the RPC method name. Internal go/gcs-sdk-tracing-span-attributes is tracking the semantic conventions for spans.

google/cloud/storage/_opentelemetry_tracing.py
cojenco: Yep only the logger.debug message is needed; Removed this line :)

google/cloud/storage/_opentelemetry_tracing.py
cojenco: Without the yield statement, the context manager will panic that the generator didn’t yield.

google/cloud/storage/_opentelemetry_tracing.py
cojenco: Good call. I’ve updated this, thanks!

andrewsg (approved):

PR #1333, opened 2024-08-05
Fix: Properly escape URL construction for XML MPU API

andrewsg:
cc: @mattlombana

### DIFF
--- a/tests/system/test_transfer_manager.py
+++ b/tests/system/test_transfer_manager.py
@@ -245,6 +245,8 @@ def test_upload_chunks_concurrently(shared_bucket, file_data, blobs_to_delete):
     chunk_size = 5 * 1024 * 1024  # Minimum supported by XML MPU API
     assert os.path.getsize(filename) > chunk_size  # Won't make a good test otherwise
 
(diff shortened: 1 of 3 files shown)


tests/system/test_transfer_manager.py
```diff
@@ -418,3 +420,58 @@ def test_upload_chunks_concurrently_with_kms(
...
+    source_file = file_data["big"]
+    filename = source_file["path"]
+    blob_name = "../example_bucket/mpu_file"
+    upload_blob = shared_bucket.blob(blob_name)
+    chunk_size = 5 * 1024 * 1024  # Minimum supported by XML MPU API
+    assert os.path.getsize(filename) > chunk_size  # Won't make a good test otherwise
+
+    blobs_to_delete.append(upload_blob)
+
+    # If the blob name is not quoted at all, this will result in a 403.
```
tritone, line 437: nit: I would call this "URL encoded" rather than "quoted" probably?

tritone (approved):

tritone (approved):

PR #1335, opened 2024-08-06
fix: Changed name of methods `Blob.from_string()` and `Bucket.from_string()` to `from_uri()`

holtskinner:
- Other methods use "string" to refer to the actual content of the Blob, not the uri. The updated method name is more intuitive for developers.
- Marked `from_string()` as deprecated and mapped functionality to `from_uri()`.

Original reasoning for `from_string()` naming in [#7693](https://github.com/googleapis/google-cloud-python/issues/7693)


andrewsg (approved):

andrewsg:
Thank you!

PR #1336, opened 2024-08-06
feat: Add support for `https://storage.googleapis.com/` URLs in `Blob.from_string()`

### LINKED ISSUE #1299: Support gs public url at `Blob.from_string`
Hi,

I am working on a project that has one Node.js and one Python codebase, both reading cloud storage objects via URIs (`gs://bucket/object` or `https://storage.googleapis.com/...`). 

I noticed that the [JavaScript implementation](https://github.com/googleapis/nodejs-storage/blob/18eef67892af6c7311a12d4c22579e240149eb02/src/file.ts#L2409) supports both URI schemas, but the [Python implementation](https://github.com/googleapis/python-storage/blob/c2ab0e035b179a919b27c7f50318472f14656e00/google/cloud/storage/blob.py#L391) only supports `gs://` URIs. 

Although it is simple to work around this limitation, is there any plan to support public URL (`https://storage.googleapis.com/...`) in the Python implementation as well? I can PR it as well, it is pretty trivial imo.

holtskinner:
Allow creating `Blob`s from `https://storage.googleapis.com` URLs.

Example:

```py
Blob.from_string("https://storage.googleapis.com/bucket/object")
```

Fixes #1299

### DIFF
--- a/google/cloud/storage/blob.py
+++ b/google/cloud/storage/blob.py
@@ -137,7 +137,7 @@ _DOWNLOAD_AS_STRING_DEPRECATED = (
     "Use Blob.download_as_bytes() instead."
 )
 _GS_URL_REGEX_PATTERN = re.compile(
-    r"(?P<scheme>gs|https)://(?:storage\.(?:mtls\.)?googleapis\.com/)?(?P<bucket_name>[a-z0-9_.-]+)/(?P<object_name>.+)"
+    r"(?P<scheme>gs)://(?P<bucket_name>[a-z0-9_.-]+)/(?P<object_name>.+)"
 )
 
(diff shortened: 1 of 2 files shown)


andrewsg:
@holtskinner Sorry for the delay. I'm not sure this is safe to for TPC customers. We might need universe domain special casing. I will check.

holtskinner:
> @holtskinner Sorry for the delay. I'm not sure this is safe for TPC users. We might need universe domain special casing. I will check.

Note - TPC (Trusted Partner Cloud)? https://www.businessinsider.com/google-cloud-europe-asia-data-sovereignty-gdpr-trusted-partner-cloud-2022-9

---

Would adding the option to use the MTLS link work as well?

holtskinner:
Closing out after offline discussion about alignment with other languages.

PR #1338, opened 2024-08-08
Fix: Add regression test for range read retry issue and bump dependency to fix

andrewsg:


sydney-munro:
nit: I will trust your judgement for your library but since its a test/dependency thing shouldn't it be `test: Add regression test for range read retry issue and bump dependency to fix` or `dep: Add regression test for range read retry issue and bump dependency to fix` or the like. In java fix gets picked up in release notes.

sydney-munro (approved):

PR #1346, opened 2024-09-20
feat: add integration test for universe domain

cojenco:


tests/system/conftest.py
```diff
@@ -331,3 +331,56 @@ def keyring(storage_client, kms_bucket, kms_client):
         except exceptions.NotFound:
             key = {"purpose": purpose}
             kms_client.create_crypto_key(keyring_path, key_name, key)
+
+
+@pytest.fixture(scope="function")
+def test_universe_domain():
+    if _helpers.test_universe_domain is None:
+        pytest.skip("TEST_UNIVERSE_DOMAIN not set in environment.")
```
JesseLovelace, line 339: Does this cause the test to fail, or to succeed without running?

If it's the latter, we may end up never running the tests if secret manager permissions get messed up and not get notified about it

andrewsg (approved):
Thanks!

tests/system/conftest.py
cojenco: It's the latter, but nox, the python testing package we use, will report if the test gets skipped. I was able to verify in nox that this test is running and passing.

JesseLovelace (approved):

PR #1357, opened 2024-10-02
fix: do not spam the log with checksum related INFO messages when downloading using transfer_manager

### LINKED ISSUE #1358: `download_chunks_concurrently` spams the log with INFO messages about checksum not being returned
#### Environment details

  - OS type and version: MacOS 14.7
  - Python version: 3.10.15
  - pip version: 24.2
  - `google-cloud-storage` version: 2.18.2

#### Steps to reproduce

  1. Use `transfer_manager.download_chunks_concurrently` API to download a large file from GCS
  2. A lot of INFO log messages should appear (for ~1GB file I get 30+):

```
INFO google.resumable_media._helpers - No MD5 checksum was returned from the service while downloading ...
(which happens for composite objects), so client-side content integrity checking is not being performed.
```

#### Code example

```python
url = "gs://..." # large object
client = storage.Client(
    credentials=credentials,
    project=project,
)
blob = Blob.from_string(url, client=client)
transfer_manager.download_chunks_concurrently(
    blob,
    filename,
)
```

I'm going to prepare a PR fixing this.

rafalh:
`download_chunks_concurrently` function does not allow to set `checksum` field in `download_kwargs`. It also does not set it on its own so it takes the default value of `"md5"` (see `Blob._prep_and_do_download`). Because ranged downloads do not return checksums it results in a lot of INFO messages (tens/hundreds):
```
INFO google.resumable_media._helpers - No MD5 checksum was returned from the service while downloading ...
(which happens for composite objects), so client-side content integrity checking is not being performed.
```
To fix it set the `checksum` field to `None` which means no checksum checking for individual chunks. Note that `transfer_manager` has its own checksum checking logic (enabled by `crc32c_checksum` argument)

Fixes #1358 🦕

### DIFF
--- a/google/cloud/storage/transfer_manager.py
+++ b/google/cloud/storage/transfer_manager.py
@@ -885,6 +885,7 @@ def download_chunks_concurrently(
             "'checksum' is in download_kwargs, but is not supported because sliced downloads have a different checksum mechanism from regular downloads. Use the 'crc32c_checksum' argument on download_chunks_concurrently instead."
         )
 
+    download_kwargs["checksum"] = None
     download_kwargs["command"] = "tm.download_sharded"
 
     # We must know the size and the generation of the blob.
--- a/google/cloud/storage/transfer_manager.py
+++ b/google/cloud/storage/transfer_manager.py
@@ -885,6 +885,7 @@ def download_chunks_concurrently(
             "'checksum' is in download_kwargs, but is not supported because sliced downloads have a different checksum mechanism from regular downloads. Use the 'crc32c_checksum' argument on download_chunks_concurrently instead."
         )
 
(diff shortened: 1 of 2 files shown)


cojenco (commented):
I think this is a reasonable patch to avoid passing in the default checksum `md5` in the subsequent call `blob._prep_and_do_download`. @andrewsg thoughts/concerns to this change?

@rafalh Could you update the corresponding unit tests in https://github.com/googleapis/python-storage/blob/main/tests/unit/test_transfer_manager.py?

andrewsg:
This makes sense to me. Thank you for your submission!

cojenco (approved):
LGTM, thank you for working on this!

PR #1362, opened 2024-10-11
fix: add type hints to blob and bucket

rinarakaki:
The goal of this PR: to pass `pyright google/cloud/storage/blob.py`


andrewsg:
We need more information and consideration before we can accept this, but we'll keep this in mind as a starting point going forward. Please reach out if you're available to discuss further.

PR #1365, opened 2024-10-17
Feat: Add restore_bucket and handling for soft-deleted buckets

### LINKED ISSUE #1347: Support for Bucket Restore
Please add support for Bucket Restore, as specified in [go/soft-delete-phase-2-client-libraries](http://goto.google.com/soft-delete-phase-2-client-libraries)

andrewsg:
fixes #1347

### DIFF
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -749,19 +749,6 @@ class Bucket(_PropertyMixin):
         if generation is not None:
             return int(generation)
 
-    @generation.setter
-    def generation(self, value):
-        """Set the generation of this bucket.
-
-        Only buckets with soft delete policies have generations.
-
-        See: https://cloud.google.com/storage/docs/managing-turbo-replication
-
-        :type value: int
-        :param value: The generation to set.
-        """
-        self._patch_property("generation", value)
-
     @property
     def soft_delete_time(self):
         """If this bucket has been soft-deleted, returns the time at which it became soft-deleted.
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -629,8 +629,7 @@ class Bucket(_PropertyMixin):
 
     :type generation: int
     :param generation: (Optional) If present, selects a specific revision of
-                       this bucket. Only buckets with soft delete policies
-                       have generations.
+                       this bucket.
     """
 
     _MAX_OBJECTS_FOR_ITERATION = 256
@@ -740,10 +739,7 @@ class Bucket(_PropertyMixin):
 
         :rtype: int or ``NoneType``
         :returns: The generation of the bucket or ``None`` if the bucket's
-                  resource has not been loaded from the server. Only buckets
-                  with soft delete policies have generations; buckets without
-                  such a policy will also have ``None`` in this field. See
-                  the Bucket.soft_delete_policy property.
+                  resource has not been loaded from the server.
         """
         generation = self._properties.get("generation")
         if generation is not None:
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -779,8 +779,6 @@ class Bucket(_PropertyMixin):
     def _query_params(self):
         """Default query parameters."""
         params = super()._query_params
-        if self.generation is not None:
-            params["generation"] = self.generation
         return params
 
     @classmethod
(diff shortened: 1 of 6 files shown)


tests/system/test_client.py
```diff
@@ -215,3 +216,46 @@ def test_client_universe_domain(
...
+    assert soft_deleted_bucket.generation == generation
+    assert soft_deleted_bucket.soft_delete_time is not None
+    assert soft_deleted_bucket.hard_delete_time is not None
+
+    # Restore the bucket.
+    restored_bucket = _helpers.retry_429_503(storage_client.restore_bucket)(
+        name, generation=generation
+    )
+    assert restored_bucket.name == name
+    assert restored_bucket.generation == generation
```
cojenco, line 259: Does a restored bucket retain the same generation it had before deletion/restoration? I'd expect the generation to change although I couldn't find any clarification in the user guide.

google/cloud/storage/bucket.py
```diff
@@ -726,6 +734,53 @@ def user_project(self):
...
+    @property
+    def generation(self):
+        """Retrieve the generation for the bucket.
+
+        :rtype: int or ``NoneType``
+        :returns: The generation of the bucket or ``None`` if the bucket's
+                  resource has not been loaded from the server. Only buckets
+                  with soft delete policies have generations; buckets without
+                  such a policy will also have ``None`` in this field. See
+                  the Bucket.soft_delete_policy property.
```
cojenco, line 746: Could we check if this is still true? I see go/gcs-soft-delete-bucket-restores-user-guide states that generation exists for every bucket, regardless of whether soft delete is enabled.

google/cloud/storage/client.py
```diff
@@ -798,17 +809,28 @@ def _bucket_arg_to_bucket(self, bucket_or_name):
                  str, \
             ]):
                 The bucket resource to pass or name to create.
+            generation (Optional[long]):
```
cojenco, line 812:
```suggestion
            generation (Optional[int]):
```

google/cloud/storage/client.py
```diff
@@ -859,6 +884,19 @@ def get_bucket(
                 See the retry.py source code and docstrings in this package (google.cloud.storage.retry) for
                 information on retry types and how to configure them.
 
+            generation (Optional[long]):
```
cojenco, line 887:
```suggestion
            generation (Optional[int]):
```

google/cloud/storage/client.py
```diff
@@ -1480,6 +1530,57 @@ def list_buckets(
             retry=retry,
         )
 
+    def restore_bucket(
+        self,
```
cojenco, line 1534: Do we want to add the optional query param `projection`? Seems like the API accepts 
`projection=noAcl` or `full`. It would be nice if there were ready to publish docs to refer to.

google/cloud/storage/client.py
```diff
@@ -1480,6 +1530,57 @@ def list_buckets(
...
+        bucket_name,
+        generation,
+        if_metageneration_match=None,
+        if_metageneration_not_match=None,
+        timeout=_DEFAULT_TIMEOUT,
+        retry=DEFAULT_RETRY,
```
cojenco, line 1540: Let's check with the API folks... it sounds more like `DEFAULT_RETRY_IF_GENERATION_SPECIFIED`?

cojenco (commented):
Appreciate you working on this while dealing with the whole credentials/secrets situation. I have some questions... and was wondering if we have the most recent user guide/docs?

tests/system/test_client.py
andrewsg: It retains the same one.

google/cloud/storage/bucket.py
andrewsg: Good catch. Not sure how I got my wires so crossed here. It does appear to exist regardless of whether soft delete is enabled. I'll fix it.

google/cloud/storage/client.py
andrewsg: Added

google/cloud/storage/client.py
```diff
@@ -798,17 +809,28 @@ def _bucket_arg_to_bucket(self, bucket_or_name):
                  str, \
             ]):
                 The bucket resource to pass or name to create.
+            generation (Optional[int]):
+                The bucket generation. Only buckets with soft delete policies
+                have generations. If generation is specified, bucket_or_name
+                must be a name (str).
```
cojenco, line 816: Let's remove this line too "Only buckets with soft delete policies have generations."

tests/system/test_client.py
cojenco: ahhh good to know! Could we link to the docs when there's a chance, thanks!

cojenco (approved):
Thanks Andrew!

PR #1371, opened 2024-10-30
samples: add OTel Tracing quickstart

cojenco:


samples/snippets/storage_trace_quickstart.py
```diff
@@ -0,0 +1,79 @@
...
+    # The contents to upload to the file
+    # contents = "The quick brown fox jumps over the lazy dog."
+
+    # In this sample, we use Google Cloud Trace to export the OpenTelemetry
+    # traces: https://cloud.google.com/trace/docs/setup/python-ot
+    # Choose and configure the exporter for your set up accordingly.
+
+    # Configure the sampling rate to control trace ingestion volume.
+    # Trace sampling rate must be in the range [0.0,1.0].
+    tracing_rate = 1.0
```
frankyn, line 52: Is there a rule of thumb for tracing rate?

samples/snippets/storage_trace_quickstart.py
```diff
@@ -0,0 +1,79 @@
...
+Sample that exports OpenTelemetry Traces collected from the Storage client to Cloud Trace.
+"""
+
+
+def run_quickstart(bucket_name, blob_name, contents):
+    # [START storage_enable_otel_tracing]
+
+    from opentelemetry import trace
+    from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+    from opentelemetry.instrumentation.requests import RequestsInstrumentor
```
frankyn, line 29: call out that this is optional for requests spans

samples/snippets/storage_trace_quickstart.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Export to Google Cloud Trace.
+    tracer_provider.add_span_processor(BatchSpanProcessor(CloudTraceSpanExporter()))
+    trace.set_tracer_provider(tracer_provider)
+
+    # Optional yet recommended: enable traces emitted from the requests HTTP library.
+    RequestsInstrumentor().instrument(tracer_provider=tracer_provider)
+
+    # Instantiate a storage client and perform a write and read workload.
+    storage_client = storage.Client()
```
frankyn, line 66: Do you need a parent span that groups the requests being made?

samples/snippets/storage_trace_quickstart.py
cojenco: It truly depends on the user application and how users want to balance cost vs trace volume. For quickstart purposes, I think we want to set to 1.0 to ensure users can see the traces in Cloud Trace console.

samples/snippets/storage_trace_quickstart.py
cojenco: updated

samples/snippets/storage_trace_quickstart.py
cojenco: Updated to create a root span that groups all the requests, thanks!

frankyn (approved):

samples/snippets/requirements.txt
```diff
@@ -3,3 +3,6 @@ google-cloud-storage==2.19.0
 pandas===1.3.5; python_version == '3.7'
 pandas===2.0.3; python_version == '3.8'
 pandas==2.2.3; python_version >= '3.9'
+opentelemetry-exporter-gcp-trace
+opentelemetry-propagator-gcp
+opentelemetry-instrumentation-requests
```
andrewsg, line 8: Can you please add a newline to the end of this file?

samples/snippets/storage_trace_quickstart.py
```diff
@@ -0,0 +1,83 @@
...
+    # The ID of your GCS bucket
+    # bucket_name = "your-bucket-name"
+    # The ID of your GCS object
+    # blob_name = "your-object-name"
+    # The contents to upload to the file
+    # contents = "The quick brown fox jumps over the lazy dog."
+
+    # In this sample, we use Google Cloud Trace to export the OpenTelemetry
+    # traces: https://cloud.google.com/trace/docs/setup/python-ot
+    # Choose and configure the exporter for your set up accordingly.
```
andrewsg, line 49: Copy: I think "set up" in this case should be "setup" but in this case let's use the more formal "configuration" or "environment" instead.

samples/snippets/storage_trace_quickstart.py
```diff
@@ -0,0 +1,83 @@
...
+# limitations under the License.
+
+import sys
+
+"""
+Sample that exports OpenTelemetry Traces collected from the Storage client to Cloud Trace.
+"""
+
+
+def run_quickstart(bucket_name, blob_name, contents):
```
andrewsg, line 24: "contents" is an unusual name for the value here. We use "data" in the library.

andrewsg (changes requested):

samples/snippets/requirements.txt
cojenco: Done

samples/snippets/storage_trace_quickstart.py
cojenco: Updated, thanks!

PR #1383, opened 2024-11-16
Feat: Add "auto" checksum option and make default

andrewsg:


tests/resumable_media/system/requests/test_download.py
```diff
@@ -62,7 +64,7 @@ class CorruptingAuthorizedSession(tr_requests.AuthorizedSession):
     """
 
     EMPTY_MD5 = base64.b64encode(hashlib.md5(b"").digest()).decode("utf-8")
-    crc32c = _helpers._get_crc32c_object()
+    crc32c = google_crc32c.Checksum()
```
cojenco, line 67: For system tests, maybe we can add a parametrized checksum "auto" here?
https://github.com/googleapis/python-storage/blob/d17a15df936ae7797c56659fedfb03bb9f177659/tests/resumable_media/system/requests/test_download.py#L272-L273


Can do the same for an upload integration test here as well
https://github.com/googleapis/python-storage/blob/d17a15df936ae7797c56659fedfb03bb9f177659/tests/resumable_media/system/requests/test_upload.py#L234-L235

cojenco (approved):
LGTM, thanks for adding clear documentation too! Just a question on adding a  parametrized checksum "auto" to the system test.

PR #1385, opened 2024-11-21
feat: Enable custom predicates for media operations

### LINKED ISSUE #1361: Resumable media retry appears to treat `deadline` differently than google.api.core.Retry
#### Environment details

  - Python version: 3.11.10
  - pip version: 24.2
  - `google-cloud-storage` version: 2.18.2

#### Description 
For resumable media calls in cloud storage (ex: [download_as_string](https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/blob.py#L1009)), the provided `google.api.core.Retry` is [converted](https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/_helpers.py#L634) to a `google.resumable_media.RetryStrategy`. However, the meaning of the provided `deadline` appears to differ between these two classes. 

For `google.api.core.Retry`, `deadline` seems to be defined as ["the maximum amount of time a function can block"](https://googleapis.dev/python/google-api-core/2.2.2/_modules/google/api_core/retry.html#:~:text=The%20deadline%20is%20the%20maxmimum%20amount%0Aof%20time%20a%20method%20can%20block.) 

However, for `google.resumable_media.RetryStrategy`, `max_cumulative_retry` (what `deadline` corresponds to) is defined as ["The maximum total amount of time to sleep during the retry process." ](https://github.com/googleapis/google-resumable-media-python/blob/main/google/resumable_media/common.py#L124C49-L125C48)

There appears to be a mismatch between the definitions though. One refers to the total time the function blocks while the other refers to the cumulative time spent sleeping between retries. 

If this is indeed a mismatch is it possible to update the documentation of resumable media operations to clarify that they have a different retry strategy?

andrewsg:
This refactor unifies the old resumable media retry code with the newer google.api_core retry code, enabling custom predicates for media operations.

Fixes #1361


tests/resumable_media/unit/requests/test__helpers.py
```diff
@@ -63,345 +57,3 @@ def test__get_body_w_content_consumed(self):
 
 def _make_response(status_code):
     return mock.Mock(status_code=status_code, spec=["status_code"])
-
-
-def _get_status_code(response):
-    return response.status_code
-
-
-class Test_wait_and_retry(object):
```
cojenco, line 72: Amazing ✨  Do we need to replace / add any retry tests? Or are the retry conformance tests sufficient?

google/cloud/storage/retry.py
```diff
@@ -35,19 +39,34 @@
     requests.ConnectionError,
     requests_exceptions.ChunkedEncodingError,
     requests_exceptions.Timeout,
+    http.client.BadStatusLine,
```
cojenco, line 42: LGTM, thanks for consolidating the retryable errors. Looking at the requests docs `requests.ConnectionError` should be equivalent to `requests.exceptions.ConnectionError` so we're good!

google/cloud/storage/_media/common.py
```diff
@@ -17,108 +17,5 @@
...
 """
 
-import http.client
-
-_SLEEP_RETRY_ERROR_MSG = (
-    "At most one of `max_cumulative_retry` and `max_retries` " "can be specified."
-)
-
 UPLOAD_CHUNK_SIZE = 262144  # 256 * 1024
 """int: Chunks in a resumable upload must come in multiples of 256 KB."""
```
cojenco, line 21: With only this constant remaining, would it make more sense to move this to the upload file?

google/cloud/storage/_media/requests/upload.py
```diff
@@ -79,9 +79,7 @@ def retriable_request():
 
             return result
 
-        return _request_helpers.wait_and_retry(
-            retriable_request, self._get_status_code, self._retry_strategy
-        )
+        return _request_helpers.wait_and_retry(retriable_request, self._retry_strategy)
```
cojenco, line 84: nit: to match download, we probably want to add the retry arg docstrings for MultipartUpload, ResumableUpload etc

google/cloud/storage/blob.py
```diff
@@ -2166,11 +2132,9 @@ def _initiate_resumable_upload(
 
         upload_url = _add_query_parameters(base_url, name_value_pairs)
         upload = ResumableUpload(
-            upload_url, chunk_size, headers=headers, checksum=checksum
+            upload_url, chunk_size, headers=headers, checksum=checksum, retry=retry
```
cojenco, line 2135: Curious will we be changing uploads to have unconditional retry by default as part of 3.0?

google/cloud/storage/blob.py
```diff
@@ -1024,9 +1023,7 @@ def _do_download(
         :type retry: google.api_core.retry.Retry
         :param retry: (Optional) How to retry the RPC. A None value will disable
             retries. A google.api_core.retry.Retry value will enable retries,
-            and the object will configure backoff and timeout options. Custom
-            predicates (customizable error codes) are not supported for media
-            operations such as this one.
+            and the object will configure backoff and timeout options.
```
cojenco, line 1026: nit: Do we want to call out the default retry? I recall we received some feedback around stating the default value in our docstrings

cojenco (commented):
Looking good, very excited about retry unification!

google/cloud/storage/blob.py
andrewsg: Yes, my intention is to make sure that change happens before the release.

google/cloud/storage/blob.py
andrewsg: I would like to incorporate DEFAULT_RETRY into the docs somehow without copy/pasting it into every method. Not sure how to do that yet. I'd prefer not to include it in this change as it's already quite large but it's a good idea.

tests/resumable_media/unit/requests/test__helpers.py
andrewsg: We have 100% coverage. Conformance tests pass. We can't reliably test retries in a system test scenario. So I think it's as tested as we can make it.

google/cloud/storage/_media/common.py
andrewsg: IIRC UPLOAD_CHUNK_SIZE is exposed in __init__.py, so rather than tear that out or modify it to expose part of upload.py I decided to leave it in place.

google/cloud/storage/blob.py
cojenco: Sounds good, let's update that in a separate PR

cojenco (approved):

google/cloud/storage/_media/requests/upload.py
andrewsg: Thanks, done

PR #1431, opened 2025-01-31
feat: add Bucket.move_blob() for HNS-enabled buckets

andrewsg:


google/cloud/storage/bucket.py
```diff
@@ -2236,6 +2237,130 @@ def rename_blob(
...
+        if_generation_match=None,
+        if_generation_not_match=None,
+        if_metageneration_match=None,
+        if_metageneration_not_match=None,
+        if_source_generation_match=None,
+        if_source_generation_not_match=None,
+        if_source_metageneration_match=None,
+        if_source_metageneration_not_match=None,
+        timeout=_DEFAULT_TIMEOUT,
+        retry=DEFAULT_RETRY,
```
cojenco, line 2255: Could you check the retry policy? Looking at copy operation and other language implementations, move object appears to be conditionally idempotent with ifGenerationMatch.

google/cloud/storage/bucket.py
```diff
@@ -2205,6 +2205,7 @@ def rename_blob(
         :rtype: :class:`Blob`
         :returns: The newly-renamed blob.
         """
+
```
cojenco, line 2208: nit: is this extra line needed?

google/cloud/storage/bucket.py
```diff
@@ -2236,6 +2237,130 @@ def rename_blob(
             )
         return new_blob
 
+    @create_trace_span(name="Storage.Bucket.moveBlob")
+    def move_blob(
+        self,
+        blob,
```
cojenco, line 2243: Double checking that this is to follow the rename_blob() / copy_blob() semantics here, correct? As discussed, I guess there are both pros and cons for taking in a blob object vs blob name string

cojenco (commented):
Overall looking good, thanks for working on this. I commented on a few questions.

tests/system/test_bucket.py
```diff
@@ -433,6 +433,37 @@ def test_bucket_copy_blob_w_metageneration_match(
...
+
+    dest = created.move_blob(
+        source,
+        "dest",
+        if_source_generation_match=source.generation,
+        if_source_metageneration_match=source.metageneration,
+    )
+    blobs_to_delete.append(dest)
+
+    assert dest.download_as_bytes() == payload
```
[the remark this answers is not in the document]
andrewsg: Done

google/cloud/storage/bucket.py
andrewsg: You're right, for some reason I got the wrong idea about other language impls. Fixed. Nice catch!

google/cloud/storage/bucket.py
andrewsg: Yes, since we have a close analogue I decided to keep the same interface.

google/cloud/storage/bucket.py
andrewsg: Nope, stray.

cojenco (approved):

PR #1441, opened 2025-02-21
Feat: Add api_key argument to Client constructor

andrewsg:


google/cloud/storage/client.py
```diff
@@ -108,6 +108,12 @@ class Client(ClientWithProject):
     :param extra_headers:
         (Optional) Custom headers to be sent with the requests attached to the client.
         For example, you can add custom audit logging headers.
+
+    :type api_key: string
+    :param api_key:
+        (Optional) An API key. Mutually exclusive with any other credentials.
+        This parameter is an alias for setting `client_options.api_key` and
+        will supercede any api key set in the `client_options` parameter.
```
cojenco, line 116: Thanks for working on this. While I understand that allowing api key-based access is the main goal, is there a specific reason why we need to add a parameter as an alias? 

`client_options` is an existing parameter, and both a dictionary or ClientOption instance is supported, making it quite easy for users to pass it in.  

```
client_options = {"api_key": api_key}
client = Client(client_options=client_options)
```
Could you share more context on the user journey and requirements for this, thanks!

andrewsg: This was the result of a discussion with the PM. Basically, the overarching design philosophy was to make it as easy to use as possible, and this alias is inexpensive for us to maintain and improves usability and legibility.

cojenco (approved):

PR #1443, opened 2025-02-25
chore: ensure full docs are generated through sphinx-docfx-yaml pipeline

cojenco:
Unblock sphinx-docfx-yaml pipeline to generate full documentation by moving the contextmanager decorator within. Filed internal ticket b/394647066 to investigate why sphinx-generated yaml documentation drops parameters when a decorator is present.


andrewsg (approved):
Thanks for doing this. Hopefully we can figure out what went wrong with Sphinx at a later date.

Issue #1458, opened 2025-04-07
Performance gap between GCS Python Client and boto3 for object downloads


dreamtalen:
#### Environment details

  - GCP VM instance: c4a-highcpu-72  
  - OS type and version: Linux, 6.1.0-31-cloud-arm64
  - Python version: Python 3.11.2
  - pip version: pip 23.0.1
  - `google-cloud-storage` version: 2.19.0 & 3.1.0

#### Steps to reproduce

When comparing download performance between Google Cloud Storage Python Client and boto3 (AWS SDK), we observed that GCS client is significantly slower (about 50% slower) than boto3 for downloading the same objects stored in a GCS bucket.

#### GCS Client Implementation (Two methods tested)
1. Using `blob.download_as_bytes()`:
```python
blob = bucket.blob(key)
data = blob.download_as_bytes()
```

2. Using `blob.open()` (10%-30% faster than method 1, but still 50% slower than boto3):
```python
with blob.open("rb") as f:
    data = f.read()
```

#### boto3 Implementation
```python
response = s3_client.get_object(Bucket=bucket_name, Key=key)
data = response['Body'].read()
```

#### Performance Results
With `ThreadPoolExecutor(max_workers=16)`, I got following average throughput downloading 64MB x 1000 objects from GCS bucket to memory:
- boto3 `get_object` : 12 Gbps
- GCS `download_as_bytes()`: 3.2 Gbps in 2.19.0 & 4.2 Gbps in 3.1.0
- GCS `blob.open()`: 4.5 Gbps in both 2.19.0 & 3.1.0

#### Questions
1. Is this performance gap expected?
2. Are there any recommended optimizations or best practices for improving download performance with the GCS Python client?
3. Are there any internal differences in how GCS supports S3-compatible APIs handling downloads that might explain the performance gap?

#### Additional Context
- We've tried various optimizations including:
  - Using `raw_download=True`
  - Configuring connection pools (Ref: https://stackoverflow.com/questions/52653409/increase-connection-pool-size)
  - Using different chunk sizes in `blob.open("rb", chunk_size=xxx)`
- The performance gap remains consistent across multiple test runs

Benchmarking scripts are available at https://github.com/dreamtalen/boto3-benchmark/tree/main/google-cloud-storage

Thanks!

cojenco:
[...]

PR #1493, opened 2025-06-19
feat: Adding support of single shot download

shubham-up-47:
Adding the support of single shot download using a boolean flag.

Some results using the script in the [customer issue](https://github.com/googleapis/python-storage/issues/1458),
* `Download` speed with `single_shot_download=false` (default value) = 123 MB / sec
* `Download` speed with `single_shot_download=true` = 273 MB / sec
* `RawDownload` speed with `single_shot_download=false` (default value) = 131 MB / sec
* `RawDownload` speed with `single_shot_download=true` = 289 MB / sec

### DIFF
--- a/google/cloud/storage/_media/_download.py
+++ b/google/cloud/storage/_media/_download.py
@@ -169,6 +169,7 @@ class Download(DownloadBase):
         headers=None,
         checksum="auto",
         retry=DEFAULT_RETRY,
+        single_shot_download=False,
     ):
         super(Download, self).__init__(
             media_url, stream=stream, start=start, end=end, headers=headers, retry=retry
@@ -182,6 +183,7 @@ class Download(DownloadBase):
         self._expected_checksum = None
         self._checksum_object = None
         self._object_generation = None
+        self._single_shot_download = single_shot_download
 
     def _prepare_request(self):
         """Prepare the contents of an HTTP request.
--- a/google/cloud/storage/_media/_download.py
+++ b/google/cloud/storage/_media/_download.py
@@ -140,7 +140,7 @@ class Download(DownloadBase):
             ``start`` to the end of the media.
         headers (Optional[Mapping[str, str]]): Extra headers that should
             be sent with the request, e.g. headers for encrypted data.
-        checksum Optional([str]): The type of checksum to compute to verify
+        checksum (Optional[str]): The type of checksum to compute to verify
             the integrity of the object. The response headers must contain
             a checksum of the requested type. If the headers lack an
             appropriate checksum (for instance in the case of transcoded or
@@ -157,6 +157,9 @@ class Download(DownloadBase):
             See the retry.py source code and docstrings in this package
             (google.cloud.storage.retry) for information on retry types and how
             to configure them.
+        single_shot_download (Optional[bool]): If true, download the object in a single request.
+            Caution: Enabling this will increase the memory overload for your application.
+            Please enable this as per your use case.
 
     """
 
--- a/google/cloud/storage/_media/_download.py
+++ b/google/cloud/storage/_media/_download.py
@@ -179,11 +179,11 @@ class Download(DownloadBase):
             self.checksum = (
                 "crc32c" if _helpers._is_crc32c_available_and_fast() else "md5"
             )
+        self.single_shot_download = single_shot_download
         self._bytes_downloaded = 0
         self._expected_checksum = None
         self._checksum_object = None
         self._object_generation = None
-        self._single_shot_download = single_shot_download
 
     def _prepare_request(self):
         """Prepare the contents of an HTTP request.
--- a/google/cloud/storage/_media/requests/download.py
+++ b/google/cloud/storage/_media/requests/download.py
@@ -139,7 +139,10 @@ class Download(_request_helpers.RequestsMixin, _download.Download):
                 content = response.raw.read()
                 self._stream.write(content)
                 self._bytes_downloaded += len(content)
-                local_checksum_object.update(content)
+                if isinstance(local_checksum_object, _helpers._DoNothingHash):
+                    checksum_object.update(content)
+                else:
+                    local_checksum_object.update(content)
                 response._content_consumed = True
             else:
                 body_iter = response.iter_content(
--- a/google/cloud/storage/_media/requests/download.py
+++ b/google/cloud/storage/_media/requests/download.py
@@ -132,13 +132,22 @@ class Download(_request_helpers.RequestsMixin, _download.Download):
             # the stream is indeed compressed, this will delegate the checksum
             # object to the decoder and return a _DoNothingHash here.
             local_checksum_object = _add_decoder(response.raw, checksum_object)
-            body_iter = response.iter_content(
-                chunk_size=_request_helpers._SINGLE_GET_CHUNK_SIZE, decode_unicode=False
-            )
-            for chunk in body_iter:
-                self._stream.write(chunk)
-                self._bytes_downloaded += len(chunk)
-                local_checksum_object.update(chunk)
+
+            if self._single_shot_download:
+                # This is useful for smaller files, or when the user wants to
+                # download the entire file in one go.
+                content = response.content
+                self._stream.write(content)
+                self._bytes_downloaded += len(content)
+                local_checksum_object.update(content)
+            else:
+                body_iter = response.iter_content(
+                    chunk_size=_request_helpers._SINGLE_GET_CHUNK_SIZE, decode_unicode=False
+                )
+                for chunk in body_iter:
+                    self._stream.write(chunk)
+                    self._bytes_downloaded += len(chunk)
+                    local_checksum_object.update(chunk)
 
         # Don't validate the checksum for partial responses.
         if (
@@ -345,14 +354,22 @@ class RawDownload(_request_helpers.RawRequestsMixin, _download.Download):
             checksum_object = self._checksum_object
 
         with response:
-            body_iter = response.raw.stream(
-                _request_helpers._SINGLE_GET_CHUNK_SIZE, decode_content=False
-            )
-            for chunk in body_iter:
-                self._stream.write(chunk)
-                self._bytes_downloaded += len(chunk)
-                checksum_object.update(chunk)
-            response._content_consumed = True
+            if self._single_shot_download:
+                # This is useful for smaller files, or when the user wants to
+                # download the entire file in one go.
+                content = response.raw.read()
+                self._stream.write(content)
+                self._bytes_downloaded += len(content)
+                checksum_object.update(content)
+            else:
+                body_iter = response.raw.stream(
+                    _request_helpers._SINGLE_GET_CHUNK_SIZE, decode_content=False
+                )
+                for chunk in body_iter:
+                    self._stream.write(chunk)
+                    self._bytes_downloaded += len(chunk)
+                    checksum_object.update(chunk)
+                response._content_consumed = True
 
         # Don't validate the checksum for partial responses.
         if (
--- a/google/cloud/storage/_media/requests/download.py
+++ b/google/cloud/storage/_media/requests/download.py
@@ -133,9 +133,9 @@ class Download(_request_helpers.RequestsMixin, _download.Download):
             # object to the decoder and return a _DoNothingHash here.
             local_checksum_object = _add_decoder(response.raw, checksum_object)
 
+            # This is useful for smaller files, or when the user wants to
+            # download the entire file in one go.
             if self.single_shot_download:
-                # This is useful for smaller files, or when the user wants to
-                # download the entire file in one go.
                 content = response.raw.read(decode_content=True)
                 self._stream.write(content)
                 self._bytes_downloaded += len(content)
@@ -356,9 +356,9 @@ class RawDownload(_request_helpers.RawRequestsMixin, _download.Download):
             checksum_object = self._checksum_object
 
         with response:
+            # This is useful for smaller files, or when the user wants to
+            # download the entire file in one go.
             if self.single_shot_download:
-                # This is useful for smaller files, or when the user wants to
-                # download the entire file in one go.
                 content = response.raw.read()
                 self._stream.write(content)
                 self._bytes_downloaded += len(content)
--- a/google/cloud/storage/_media/requests/download.py
+++ b/google/cloud/storage/_media/requests/download.py
@@ -133,10 +133,10 @@ class Download(_request_helpers.RequestsMixin, _download.Download):
             # object to the decoder and return a _DoNothingHash here.
             local_checksum_object = _add_decoder(response.raw, checksum_object)
 
-            if self._single_shot_download:
+            if self.single_shot_download:
                 # This is useful for smaller files, or when the user wants to
                 # download the entire file in one go.
-                content = response.content
+                content = response.raw.read()
                 self._stream.write(content)
                 self._bytes_downloaded += len(content)
                 local_checksum_object.update(content)
@@ -354,7 +354,7 @@ class RawDownload(_request_helpers.RawRequestsMixin, _download.Download):
             checksum_object = self._checksum_object
 
         with response:
-            if self._single_shot_download:
+            if self.single_shot_download:
                 # This is useful for smaller files, or when the user wants to
                 # download the entire file in one go.
                 content = response.raw.read()
@@ -369,7 +369,7 @@ class RawDownload(_request_helpers.RawRequestsMixin, _download.Download):
                     self._stream.write(chunk)
                     self._bytes_downloaded += len(chunk)
                     checksum_object.update(chunk)
-                response._content_consumed = True
+            response._content_consumed = True
 
         # Don't validate the checksum for partial responses.
         if (
--- a/google/cloud/storage/_media/requests/download.py
+++ b/google/cloud/storage/_media/requests/download.py
@@ -142,7 +142,8 @@ class Download(_request_helpers.RequestsMixin, _download.Download):
                 local_checksum_object.update(content)
             else:
                 body_iter = response.iter_content(
-                    chunk_size=_request_helpers._SINGLE_GET_CHUNK_SIZE, decode_unicode=False
+                    chunk_size=_request_helpers._SINGLE_GET_CHUNK_SIZE,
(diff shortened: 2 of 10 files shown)


google/cloud/storage/_media/requests/download.py
```diff
@@ -132,13 +132,22 @@ def _write_to_stream(self, response):
...
-            )
-            for chunk in body_iter:
-                self._stream.write(chunk)
-                self._bytes_downloaded += len(chunk)
-                local_checksum_object.update(chunk)
+
+            if self._single_shot_download:
+                # This is useful for smaller files, or when the user wants to
+                # download the entire file in one go.
+                content = response.content
```
chandra-siri: Just a headsup : make sure this also works for objects with `content-encoding`:gzip and br

chandra-siri: `response.content` will not download the entire file at once.  
see [response.content code](https://github.com/psf/requests/blob/91a3eabd3dcc4d7f36dd8249e4777a90ef9b4305/src/requests/models.py#L902) which internally calls `iter_content` with `CONTENT_CHUNK_SIZE` (10KiB) little more than `_SINGLE_GET_CHUNK_SIZE ` (8KiB). 

that's the reason you're seeing only marginal improvement in download speed, from 123 -> 136

I suggest using [response.raw.stream](https://github.com/urllib3/urllib3/blob/f80d281fe53dfe93004b856ba8b9adb3fa37da77/src/urllib3/response.py#L1069), response.raw will be urlib3's response object.

chandra-siri (changes requested):

shubham-up-47:
> @shubham-up-47 Please add unit tests.

Done.

google/cloud/storage/_media/requests/download.py
shubham-up-47: Sure, tests in this file [tests/resumable_media/system/requests/test_download.py](https://github.com/googleapis/python-storage/pull/1493/files#diff-1ad8f54c293facaa96176b44aced4805b1c3e7cd02bf5a03401876594b76914e) are ensuring that.

shubham-up-47: Yeah, i didn't know that. Thanks a lot for telling. Updated the code now. Using response.rew.read() in that too now.

google/cloud/storage/blob.py
```diff
@@ -1484,6 +1504,10 @@ def download_as_bytes(
             (google.cloud.storage.retry) for information on retry types and how
             to configure them.
 
+        :type single_shot_download: bool
+        :param single_shot_download:
+            (Optional) If true, download the object in a single request.
```
chandra-siri, line 1518: Can we add a warning to the user: 

`Caution: Enabling this will increase the memory overload for your application. Please enable this as per your user use case.`

Also where ever application in this file.

google/cloud/storage/_media/requests/download.py
```diff
@@ -132,13 +132,24 @@ def _write_to_stream(self, response):
...
-            body_iter = response.iter_content(
-                chunk_size=_request_helpers._SINGLE_GET_CHUNK_SIZE, decode_unicode=False
-            )
-            for chunk in body_iter:
-                self._stream.write(chunk)
-                self._bytes_downloaded += len(chunk)
-                local_checksum_object.update(chunk)
+
+            if self.single_shot_download:
+                # This is useful for smaller files, or when the user wants to
```
chandra-siri: nit: can you move this comment above if statement?

google/cloud/storage/_media/requests/download.py
```diff
@@ -132,13 +132,24 @@ def _write_to_stream(self, response):
...
+                # download the entire file in one go.
+                content = response.raw.read(decode_content=True)
+                self._stream.write(content)
+                self._bytes_downloaded += len(content)
+                local_checksum_object.update(content)
+                response._content_consumed = True
```
chandra-siri, line 143: why this is required ., just curious why it wasn't needed before?

tests/system/test_blob.py
```diff
@@ -1149,3 +1149,23 @@ def test_object_retention_lock(storage_client, buckets_to_delete, blobs_to_delet
...
     blob.patch(override_unlocked_retention=True)
     assert blob.retention.mode is None
+
+
+def test_blob_download_as_bytes_single_shot_download(
+    shared_bucket, blobs_to_delete, file_data, service_account
+):
+    blob_name = f"download-single-shot-{uuid.uuid4().hex}"
+    info = file_data["simple"]
+    payload = None
```
chandra-siri: this is not required.

tests/unit/test_blob.py
```diff
@@ -1612,9 +1623,16 @@ def _download_to_file_helper(
 
         with mock.patch.object(blob, "_prep_and_do_download"):
             if raw_download:
-                blob.download_to_file(file_obj, raw_download=True, **extra_kwargs)
+                blob.download_to_file(
+                    file_obj,
+                    raw_download=True,
+                    single_shot_download=single_shot_download,
+                    **extra_kwargs,
```
chandra-siri, line 1630: `**extra_kwargs` should take care of `single_shot_download=single_shot_download`, isn't it ?

tests/unit/test_blob.py
```diff
@@ -1884,11 +1927,14 @@ def test_download_to_filename_notfound(self):
...
                 checksum="auto",
                 retry=DEFAULT_RETRY,
+                single_shot_download=False,
             )
             stream = blob._prep_and_do_download.mock_calls[0].args[0]
             self.assertEqual(stream.name, filename)
 
-    def _download_as_bytes_helper(self, raw_download, timeout=None, **extra_kwargs):
+    def _download_as_bytes_helper(
+        self, raw_download, timeout=None, single_shot_download=False, **extra_kwargs
```
chandra-siri, line 1936: the default value is `False` , why you're adding explicitly ? . It just adds more "diffs" to PR.

chandra-siri (changes requested):

google/cloud/storage/_media/requests/download.py
shubham-up-47: As discussed offline, this is marking the content consumed, which isn't updated by default (when we are reading the raw content).

shubham-up-47:
> Nit: Also please follow https://www.conventionalcommits.org/en/v1.0.0/#specification , we don't need (storage) , unlike C++ where it's a mono-repo

Done. Noted.

tests/unit/test_blob.py
shubham-up-47: No, it is passing the default False value of `single_shot_download` otherwise.

Test it failing without this assignment, checked it now.

tests/unit/test_blob.py
shubham-up-47: No, this is a separate helper function, and it doesn't know about the `single_shot_download` parameter, that's why it needs to me added here.

Test is failing without this parameter definition,
self = <tests.unit.test_blob.Test_Blob testMethod=test_download_as_bytes_wo_single_shot_download>, raw_download = False
timeout = None, extra_kwargs = {'retry': None, 'single_shot_download': False}, blob_name = 'blob-name'
client = <NonCallableMagicMock spec='Client' id='140577857904928'>
bucket = <tests.unit.test_blob._Bucket object at 0x7fdad5489250>, blob = <Blob: name, blob-name, None>, expected_timeout = 60

    def _download_as_bytes_helper(
        self, raw_download, timeout=None, **extra_kwargs
    ):
        blob_name = "blob-name"
        client = self._make_client()
        bucket = _Bucket(client)
        blob = self._make_one(blob_name, bucket=bucket)
    
        with mock.patch.object(blob, "_prep_and_do_download"):
            if timeout is None:
                expected_timeout = self._get_default_timeout()
                fetched = blob.download_as_bytes(
                    raw_download=raw_download,
>                   single_shot_download=single_shot_download,
                    **extra_kwargs,
                )
E               NameError: name 'single_shot_download' is not defined

tests/unit/test_blob.py:1948: NameError
=================================================== short test summary info ===================================================
FAILED tests/unit/test_blob.py::Test_Blob::test_download_as_byte_w_custom_timeout - NameError: name 'single_shot_download' is not defined
FAILED tests/unit/test_blob.py::Test_Blob::test_download_as_bytes_no_retry - NameError: name 'single_shot_download' is not defined
FAILED tests/unit/test_blob.py::Test_Blob::test_download_as_bytes_w_custom_timeout - NameError: name 'single_shot_download' is not defined
FAILED tests/unit/test_blob.py::Test_Blob::test_download_as_bytes_w_raw - NameError: name 'single_shot_download' is not defined
FAILED tests/unit/test_blob.py::Test_Blob::test_download_as_bytes_w_single_shot_download - NameError: name 'single_shot_download' is not defined
FAILED tests/unit/test_blob.py::Test_Blob::test_download_as_bytes_wo_raw - NameError: name 'single_shot_download' is not defined
FAILED tests/unit/test_blob.py::Test_Blob::test_download_as_bytes_wo_single_shot_download - NameError: name 'single_shot_download' is not defined
================================================ 7 failed, 374 passed in 4.93s ================================================

google/cloud/storage/blob.py
shubham-up-47: Done.

google/cloud/storage/_media/requests/download.py
shubham-up-47: Done.

tests/system/test_blob.py
shubham-up-47: Removed.

google/cloud/storage/blob.py
```diff
@@ -1047,13 +1048,20 @@ def _do_download(
             See the retry.py source code and docstrings in this package
             (google.cloud.storage.retry) for information on retry types and how
             to configure them.
+
+        :type single_shot_download: bool
+        :param single_shot_download:
+            (Optional) If true, download the object in a single request.
+            Caution: Enabling this will increase the memory overload for your application.
+            Please enable this as per your user use case.
```
chandra-siri: remove `user`

google/cloud/storage/blob.py
```diff
@@ -1222,6 +1232,12 @@ def download_to_file(
             (google.cloud.storage.retry) for information on retry types and how
             to configure them.
 
+        :type single_shot_download: bool
+        :param single_shot_download:
+            (Optional) If true, download the object in a single request.
+            Caution: Enabling this will increase the memory overload for your application.
+            Please enable this as per your user use case.
```
chandra-siri: remove `user` same here and everywhere else

chandra-siri (dismissed):

google/cloud/storage/blob.py
shubham-up-47: Done.

google/cloud/storage/blob.py
shubham-up-47: Done.

chandra-siri (dismissed):

danielduhh (commented):
Design question: why put this behind an opt-in flag? Is there a non backwards compatible change that you're concerned about? Or is the flag temporary?

This requires a customer to take an extra step in order to benefit from the performance optimization.

shubham-up-47:
> Design question: why put this behind an opt-in flag? Is there a non backwards compatible change that you're concerned about? Or is the flag temporary?
> 
> This requires a customer to take an extra step in order to benefit from the performance optimization.

We wanted to keep the existing flow as it is and add the support of single_shot_download, that's why keeping its default value as False currently.

Later if everything works fine, then we will make the default value of single_shot_download flag as true.

google/cloud/storage/_media/requests/download.py
cojenco: I'm also curious why this is added. It seems like we're modifying the requests library's internal state management. Is this not handled automatically in the read call?

google/cloud/storage/_media/requests/download.py
```diff
@@ -132,13 +132,24 @@ def _write_to_stream(self, response):
...
-            )
-            for chunk in body_iter:
-                self._stream.write(chunk)
-                self._bytes_downloaded += len(chunk)
-                local_checksum_object.update(chunk)
+
+            # This is useful for smaller files, or when the user wants to
+            # download the entire file in one go.
+            if self.single_shot_download:
+                content = response.raw.read(decode_content=True)
```
cojenco, line 139: With decode_content=True on raw.read(), could you double check
(a) there are system tests testing compressed files with checksum enabled
(b) users can expect the same results with compressed files in both code paths (single_shot_download=True and single_shot_download=False)

tests/resumable_media/system/requests/test_download.py
```diff
@@ -306,6 +323,28 @@ def test_download_to_stream(self, add_files, authorized_transport):
             assert stream.getvalue() == actual_contents
             check_tombstoned(download, authorized_transport)
 
+    def test_single_shot_download_to_stream(self, add_files, authorized_transport):
```
cojenco: Could you also add checksums to this test as well?

@pytest.mark.parametrize("checksum", ["auto", "md5", "crc32c", None])

tests/resumable_media/system/requests/test_download.py
```diff
@@ -286,6 +286,23 @@ def test_download_full(self, add_files, authorized_transport, checksum):
             assert self._read_response_content(response) == actual_contents
             check_tombstoned(download, authorized_transport)
 
+    @pytest.mark.parametrize("checksum", ["auto", "md5", "crc32c", None])
+    def test_single_shot_download_full(self, add_files, authorized_transport, checksum):
```
cojenco, line 290: Please add tests to `TestRawDownload` as well

google/cloud/storage/_media/_download.py
```diff
@@ -169,6 +169,7 @@ def __init__(
         headers=None,
         checksum="auto",
         retry=DEFAULT_RETRY,
+        single_shot_download=False,
```
cojenco, line 175: nit: could you update the docstrings here as well

google/cloud/storage/blob.py
```diff
@@ -1222,6 +1232,12 @@ def download_to_file(
             (google.cloud.storage.retry) for information on retry types and how
             to configure them.
 
+        :type single_shot_download: bool
+        :param single_shot_download:
+            (Optional) If true, download the object in a single request.
+            Caution: Enabling this will increase the memory overload for your application.
+            Please enable this as per your use case.
```
cojenco, line 1242: Thanks for adding the note on memory consumption. 

Based on the design doc, I understand we are introducing single shot download support in multiple phases. Note that if and when we want to have `single_shot_download` defaulted to True, that would be a breaking change and would require a major version bump, so something to keep in mind when planning for next version changes

google/cloud/storage/blob.py
```diff
@@ -1047,13 +1048,20 @@ def _do_download(
             See the retry.py source code and docstrings in this package
             (google.cloud.storage.retry) for information on retry types and how
             to configure them.
+
+        :type single_shot_download: bool
+        :param single_shot_download:
+            (Optional) If true, download the object in a single request.
+            Caution: Enabling this will increase the memory overload for your application.
+            Please enable this as per your use case.
```
cojenco, line 1056: nit: although this is an internal method, consider adding a note that `single_shot_download` is only supported in `Download` and `RawDownload` classes aka when chunk_size is set to None (default value)

cojenco (commented):
Left a few questions. Please also add system tests under [TestRawDownload](https://github.com/googleapis/python-storage/blob/main/tests/resumable_media/system/requests/test_download.py#L424) as well, thanks!

google/cloud/storage/blob.py
danielduhh: @cojenco how is that a breaking change?

google/cloud/storage/_media/requests/download.py
shubham-up-47: It is required when we are consuming the raw content (i.e. response.raw.read line), then we will have to explicitly mark the content as consumed (as we are doing [here](https://github.com/googleapis/python-storage/blob/129455b4290dfc75349f5529d167809df40be783/google/cloud/storage/_media/requests/download.py#L355) in `RawDownload` class).

tests/resumable_media/system/requests/test_download.py
shubham-up-47: Done.

tests/resumable_media/system/requests/test_download.py
shubham-up-47: Actually `TestRawDownload` is inheriting the `TestDownload` class, that's why all the tests written in the `TestDownload` class are automatically running for the `TestRawDownload` class.

https://screenshot.googleplex.com/4GYUPuXSXVkT7uh

google/cloud/storage/_media/requests/download.py
shubham-up-47: The `test_single_shot_download_to_stream` in the [file](https://github.com/googleapis/python-storage/pull/1493/files#diff-1ad8f54c293facaa96176b44aced4805b1c3e7cd02bf5a03401876594b76914e), takes care of these 2 scenarios because,
*  it is testing with and without checksum
* and ALL_FILES contains both compressed and decompressed files

google/cloud/storage/_media/_download.py
shubham-up-47: Done.

google/cloud/storage/blob.py
shubham-up-47: Done.

google/cloud/storage/blob.py
shubham-up-47: Sure @cojenco, will keep these things in mind.

shubham-up-47:
> Left a few questions. Please also add system tests under [TestRawDownload](https://github.com/googleapis/python-storage/blob/main/tests/resumable_media/system/requests/test_download.py#L424) as well, thanks!

Actually TestRawDownload is inheriting the TestDownload class, that's why all the tests written in the TestDownload class are automatically running for the TestRawDownload class.

https://screenshot.googleplex.com/4GYUPuXSXVkT7uh

chandra-siri (approved):

cojenco (approved):

ShreyasSinha (approved):

PR #1522, opened 2025-08-18
feat: Indicate that md5 is used as a CRC

xnox:
MD5 in storage helpers is used as a CRC function for
non-cryptographically secure purposes. Ensure that md5 is initiated
with `usedforsecurity=False` to ensure that Python in FIPS mode can
fetch MD5 implementation for such non cryptographically secure
purpose.

This is no effective change on non-FIPS mode Python installations.

This improves compatibility with most FIPS mode Python installations.

### DIFF
--- a/google/cloud/storage/_helpers.py
+++ b/google/cloud/storage/_helpers.py
@@ -21,6 +21,7 @@ import base64
 import datetime
 from hashlib import md5
 import os
+import sys
 from urllib.parse import urlsplit
 from urllib.parse import urlunsplit
 from uuid import uuid4
@@ -536,7 +537,10 @@ def _base64_md5hash(buffer_object):
     :rtype: str
     :returns: A base64 encoded digest of the MD5 hash.
     """
-    hash_obj = md5()
+    if sys.version_info >= (3, 9):
+        hash_obj = md5(usedforsecurity=False)
+    else:
+        hash_obj = md5()
     _write_buffer_to_hash(buffer_object, hash_obj)
     digest_bytes = hash_obj.digest()
     return base64.b64encode(digest_bytes)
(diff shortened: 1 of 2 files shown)


xnox:
@chandra-siri can you please approve this workflows to run?

This is currently blocking multiple deployments to access GCP storage when using Python in FIPS mode.

google/cloud/storage/_helpers.py
```diff
@@ -536,7 +536,7 @@ def _base64_md5hash(buffer_object):
     :rtype: str
     :returns: A base64 encoded digest of the MD5 hash.
     """
-    hash_obj = md5()
+    hash_obj = md5(usedforsecurity=False)
     _write_buffer_to_hash(buffer_object, hash_obj)
     digest_bytes = hash_obj.digest()
     return base64.b64encode(digest_bytes)
```
chandra-siri, line 546: This seem to work only for python versions >= 3.9 , could you add that check ? 

something like 
```python
if sys.version_info >= (3, 9):
  hash_obj = md5(usedforsecurity=False)
else:
  hash_obj = md5
```

chandra-siri (changes requested):

chandra-siri (dismissed):

xnox:
```
1909 passed, 26 warnings in 22.95s
nox > Session unit-3.13 was successful.
```

There are many other warnings about other mocks which print a lot of noise, but hopefully all of these are ok.

The _MD5 mock is now updated and the test pass with at least one python version, I do not have older versions of python readily available to me.

chandra-siri (approved):

Pulkit0110 (approved):

xnox:
CI looks green! Just need a second reviewer now.

PR #1525, opened 2025-08-19
fix: Send entire object checksum at Resumable Upload Initiation Request

### LINKED ISSUE #1524: Python SDK should send entire object checksum at Resumable Upload Initiation Request
More details in internal b/434685905

chandra-siri:
Send entire object checksum at Resumable Upload Initiation Request.

Entire object checksum is provided while blob instantiation. The same will be used as data integrity checks and upload will succeed only when server computed checksum and user provided checksum match.

Fixes #1524 🦕

### DIFF
--- a/google/cloud/storage/bucket.py
+++ b/google/cloud/storage/bucket.py
@@ -876,6 +876,7 @@ class Bucket(_PropertyMixin):
         :type generation: long
         :param generation: (Optional) If present, selects a specific revision of
                            this object.
+
         :type crc32c_checksum: str
         :param crc32c_checksum:
             (Optional) If set, the CRC32C checksum of the blob's content.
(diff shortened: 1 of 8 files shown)


google/cloud/storage/bucket.py
```diff
@@ -873,6 +877,14 @@ def blob(
         :param generation: (Optional) If present, selects a specific revision of
                            this object.
 
+        :type crc32c_checksum: str
+        :param crc32c_checksum:
+            (Optional) If set, the CRC32C checksum of the blob's content.
```
shubham-up-47, line 882: How this crc32c_checksum value is set?

shubham-up-47 (approved):

google/cloud/storage/bucket.py
chandra-siri: User will set this while creating blob.

```python
    blob = bucket.blob(destination_blob_name, crcr32c_checksum='user_calculated_value')
```

PR #1545, opened 2025-09-17
feat: Add async bidiRpc files in python-storage

chandra-siri:
these files will be removed once https://github.com/googleapis/python-api-core/pull/836 gets submitted.


google/cloud/storage/_experimental/asyncio/bidi_base.py
```diff
@@ -0,0 +1,80 @@
...
+# limitations under the License.
+
+"""Base class for bi-directional streaming RPC helpers."""
+
+
+class BidiRpcBase:
+    """A base class for consuming a bi-directional streaming RPC.
+
+    This maps gRPC's built-in interface which uses a request iterator and a
+    response iterator into a socket-like :func:`send` and :func:`recv`. This
```
Pulkit0110, line 20: This class doesn't have the `:func:send` and `:func:recv`.

google/cloud/storage/_experimental/asyncio/bidi_base.py
```diff
@@ -0,0 +1,80 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Base class for bi-directional streaming RPC helpers."""
+
+
+class BidiRpcBase:
```
Pulkit0110, line 16: Are there any use cases where this class would be used directly by the users? If not, should we consider making it an abstract base class?

google/cloud/storage/_experimental/asyncio/bidi_async.py
```diff
@@ -0,0 +1,230 @@
...
+      requests to a gRPC stream from a Queue.
+
+    This generator takes asynchronous requests off a given queue and yields them
+    to gRPC.
+
+    This helper is useful when you have an indeterminate, indefinite, or
+    otherwise open-ended set of requests to send through a request-streaming
+    (or bidirectional) RPC.
+
+    The reason this is necessary
```
Pulkit0110, line 37: nit: remove the extra line. And looks like something is missing in the sentence e.g. `user have control on the when they would want to send`

google/cloud/storage/_experimental/asyncio/bidi_async.py
```diff
@@ -0,0 +1,230 @@
...
+            easily restarting streams that require some initial configuration
+            request.
+    """
+
+    def __init__(self, queue: asyncio.Queue, initial_request=None):
+        self._queue = queue
+        self._initial_request = initial_request
+        self.call = None
+
+    def _is_active(self):
```
Pulkit0110, line 76: It's not clear who's/what activeness it checks/return. I think it returns whether the gRPC stream is active?

google/cloud/storage/_experimental/asyncio/bidi_async.py
```diff
@@ -0,0 +1,230 @@
...
+    async def close(self):
+        """Closes the stream."""
+        if self.call is None:
+            return
+
+        await self._request_queue.put(None)
+        self.call.cancel()
+        self._request_generator = None
+        self._initial_request = None
+        self._callbacks = []
```
Pulkit0110, line 191: I think we should remove the lines
```
self._initial_request = None
self._callbacks = []
```

If the user wants to restart the stream with same configuration, then it won't be possible or they'll have to create a new object/stream.

google/cloud/storage/_experimental/asyncio/bidi_base.py
```diff
@@ -0,0 +1,80 @@
...
+                    to start the RPC.
+        initial_request (Union[protobuf.Message,
+                Callable[[], protobuf.Message]]): The initial request to
+            yield. This is useful if an initial request is needed to start the
+            stream.
+        metadata (Sequence[Tuple(str, str)]): RPC metadata to include in
+            the request.
+    """
+
+    def __init__(self, start_rpc, initial_request=None, metadata=None):
```
Pulkit0110, line 39: we should add a check that `start_rpc `must not be `None`

Pulkit0110 (changes requested):

Pulkit0110 (approved):

PR #1546, opened 2025-09-17
feat(experimental): add AsyncAbstractObjectStream

chandra-siri:
this will be the parent class for AsyncReadObjectStream and AsyncWriteObjectStream


Pulkit0110 (commented):
nit: add `experimental` in the PR title

google/cloud/storage/_experimental/asyncio/async_abstract_object_stream.py
Pulkit0110, line 1: Please add license in this file

google/cloud/storage/_experimental/asyncio/async_abstract_object_stream.py
```diff
@@ -0,0 +1,36 @@
...
+
+class _AsyncAbstractObjectStream(abc.ABC):
+    """
+    Abstract class for both ReadObjectStream as well as WriteObjectStream.
+
+    Attributes will include
+    1. bucket_name
+    2. object_name
+    3. generation_number (if given)
+
```
Pulkit0110, line 30: refactor the docstring to make it consistent with other python sdk files.

google/cloud/storage/_experimental/asyncio/async_abstract_object_stream.py
```diff
@@ -0,0 +1,36 @@
...
+
+    """
+
+    def __init__(self, bucket_name, object_name, generation_number=None):
```
Pulkit0110: nit: add type hints

google/cloud/storage/_experimental/asyncio/async_abstract_object_stream.py
```diff
@@ -0,0 +1,36 @@
...
+        super().__init__()
```
Pulkit0110, line 47: should we add a runtime check to make sure that `bucket_name` and `object_name` are never `None`.

google/cloud/storage/_experimental/asyncio/async_abstract_object_stream.py
```diff
@@ -0,0 +1,36 @@
...
+        self.bucket_name = bucket_name
+        self.object_name = object_name
+        self.generation_number = generation_number
+
+    @abc.abstractmethod
+    async def open(self):
+        raise NotImplementedError("Subclasses should implement this method.")
```
Pulkit0110: I think raising the error is redundant as the decorator `@abc.abstractmethod` itself does that. You can use` pass` instead.

google/cloud/storage/_experimental/asyncio/async_abstract_object_stream.py
```diff
@@ -0,0 +1,36 @@
+import abc
+
+
+class _AsyncAbstractObjectStream(abc.ABC):
```
Pulkit0110, line 19: Question: Would this class act as the abstract base class for any  `ReadObjectStream` or `WriteObjectStream`? If yes, then I think `recv()` and `send()` methods shouldn't be a part of this class. Reason is, it would make sense and work perfectly fine for the bidi RPCs but for non-bidi RPC both of these methods might not make sense (e.g. if I have to only read an object using async client, then the class should only need and implement the `recv()` method).

Pulkit0110 (changes requested):

google/cloud/storage/_experimental/asyncio/async_abstract_object_stream.py
chandra-siri: done.

google/cloud/storage/_experimental/asyncio/async_abstract_object_stream.py
chandra-siri: _" If I have to only read an object using async client, then the class should only need and implement the recv()"_


This isn't correct. `recv()` fetches response from server-to-client stream. it should be present for both reads and writes.

google/cloud/storage/_experimental/asyncio/async_abstract_object_stream.py
chandra-siri: `bucket_name` and `object_name` can be `None` when user opens gRPC stream with `read_handle`

chandra-siri: you're right, bucket_name and object_name cannot be None. Added a check.

Pulkit0110 (approved):

PR #1551, opened 2025-09-18
feat(zb-experimental): implement download_ranges

chandra-siri:
Downloads multiple byte ranges from the object into the buffers provided by user.

This interface provides option to download multiple ranges of a GCS  `Object`
    concurrently.

Example usage

```python

        client = AsyncGrpcClient().grpc_client
        mrd = await AsyncMultiRangeDownloader.create_mrd(
            client, bucket_name="chandrasiri-rs", object_name="test_open9"
        )
        my_buff1 = open('my_fav_file.txt', 'wb')
        my_buff2 = BytesIO()
        my_buff3 = BytesIO()
        my_buff4 = any_object_which_provides_BytesIO_like_interface()
        results_arr, error_obj = await mrd.download_ranges(
            [
                (0, 100, my_buff1),
                (100, 20, my_buff2),
                (200, 123, my_buff3),
                (300, 789, my_buff4),
            ]
        )
        if error_obj:
            print("Error occurred: ")
            print(error_obj)
            print(
                "please issue call to `download_ranges` with updated"
                "`read_ranges` based on diff of (bytes_requested - bytes_written)"
            )

        for result in results_arr:
            print("downloaded bytes", result)
```


google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
```diff
@@ -148,18 +188,71 @@ async def open(self) -> None:
...
+                    results[read_id].bytes_written += len(data)
+
+                    if object_data_range.range_end:
+                        del read_id_to_writable_buffer_dict[
+                            object_data_range.read_range.read_id
+                        ]
+            except Exception as exc:
+                exception = exc
+                break
+        return results, exception
```
Pulkit0110: Any specific reason why we're returning the exception instead of raising it?

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
```diff
@@ -38,21 +70,29 @@ class AsyncMultiRangeDownloader:
         mrd = await AsyncMultiRangeDownloader.create_mrd(
             client, bucket_name="chandrasiri-rs", object_name="test_open9"
         )
-        my_buff1 = BytesIO()
+        my_buff1 = open('my_fav_file.txt', 'wb')
```
Pulkit0110, line 73: Question: in this example, what are the pros and cons of using `open()` (which is synchronous) and `aiofiles` (which is asynchronous)?

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
```diff
@@ -148,18 +188,71 @@ async def open(self) -> None:
...
 
-
-        Raises:
-            NotImplementedError: This method is not yet implemented.
         """
-        raise NotImplementedError("TODO")
+        if len(read_ranges) > 1000:
+            raise Exception(
+                "Invalid input - length of read_ranges cannot be more than 1000"
+            )
```
Pulkit0110, line 212: Question: Is the limit of 1000 enforced by the backend or is it created by us?

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
```diff
@@ -148,18 +188,71 @@ async def open(self) -> None:
...
+        for i in range(0, len(read_ranges), _MAX_READ_RANGES_PER_BIDI_READ_REQUEST):
+            read_ranges_segment = read_ranges[
+                i : i + _MAX_READ_RANGES_PER_BIDI_READ_REQUEST
+            ]
+
+            read_ranges_for_bidi_req = []
+            for j, read_range in enumerate(read_ranges_segment):
+                read_id = i + j
+                read_id_to_writable_buffer_dict[read_id] = read_range[2]
+                bytes_requested = read_range[1]
```
Pulkit0110, line 225: should it be `bytes_requested = read_range[1] - read_range[0]+1` or update the docstring that the tuple holds start_byte, length and buffer?

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
```diff
@@ -148,18 +188,71 @@ async def open(self) -> None:
...
+                    raise Exception("None response received, something went wrong.")
+
+                for object_data_range in response.object_data_ranges:
+                    if object_data_range.read_range is None:
+                        raise Exception("Invalid response, read_range is None")
+
+                    data = object_data_range.checksummed_data.content
+                    read_id = object_data_range.read_range.read_id
+                    buffer = read_id_to_writable_buffer_dict[read_id]
+                    buffer.write(data)
```
Pulkit0110: I think it should be `await buffer.write(data)` else it will block the program.

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
```diff
@@ -148,18 +188,71 @@ async def open(self) -> None:
...
+                read_ranges_for_bidi_req.append(
+                    _storage_v2.ReadRange(
+                        read_offset=read_range[0],
+                        read_length=bytes_requested,
+                        read_id=read_id,
+                    )
+                )
+            await self.read_obj_str.send(
+                _storage_v2.BidiReadObjectRequest(read_ranges=read_ranges_for_bidi_req)
+            )
```
Pulkit0110, line 236: What if `read_obj_str.send` call fails? I think we should put it inside a try-catch block or may be put the entire for loop inside a try catch block.

Pulkit0110 (changes requested):

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
chandra-siri: currently it's sync open, which as you said, will be blocking.

hmm.. haven't thought of async buffers, yet. We can discuss about this and scope it in future work items.

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
chandra-siri: backend enforces a limit of 100 ranges per `BidiReadRequest`. However, if the ranges > 100 and we split them across multiple `BidiReadRequests` to maximum of 1000 (ie 10 requests) . This may change subject to experimentation. The reason - we're making a first recv() call only after all the send() calls are made. We need to do some experimentation to figure this out.

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
chandra-siri: When I wrote the doc string it was `start_byte` & `end_byte` but modified it during implementation to keep it consistent with other SDKs. (C++ , Go). Modified the doc string as well.

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
chandra-siri: buffer is sync currently. Providing support for async buffer is new scope. 

as discussed in this [comment](https://github.com/googleapis/python-storage/pull/1551/files/e571dd1e4e8160b23eedc310f05218cc491f9310#r2378417101) let's discuss and prioritize this work item.

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
chandra-siri: because the caller is expected to handle partial failure scenario as of now. which can be detected by results objects' `bytes_requested` and `bytes_written` attributes.

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
chandra-siri: I've haven't kept try-catch block because there's no need to handle it, if `.send` fails error should be raised and reported to caller.

chandra-siri:
@Pulkit0110 Replied to your comments, please review further. Feel free to resolve if comments sounds good.

tests/unit/asyncio/test_async_multi_range_downloader.py
Pulkit0110, line 1: We should add more test cases to cover scenarios like `len(read_ranges) > 1000`,  `len(read_ranges) > 100 and len(read_ranges) < 1000`.

Pulkit0110 (changes requested):

tests/unit/asyncio/test_async_multi_range_downloader.py
chandra-siri: yes, it'll be added in follow up PRs

chandra-siri: actually pushed a commit , just now

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
chandra-siri: Discussed offline and agreed to raise Exception instead.

google/cloud/storage/_experimental/asyncio/async_multi_range_downloader.py
chandra-siri: Discussed offline and will be taking this as future work item.

Pulkit0110 (approved):

PR #1553, opened 2025-09-24
fix: Redact sensitive data from OTEL traces and fix env var parsing

rajeevpodar:
fix: Redact sensitive data from OTEL traces and fix env var parsing
### Summary

This PR addresses two key improvements:

1. **Redaction of Sensitive Data in OTEL Traces:**  
   - Ensures that sensitive information (such as authentication tokens, secret values, etc.) is not captured or exposed in OpenTelemetry traces.
   - Applies systematic filtering and redaction to both HTTP request and response data included in trace spans.

2. **Robust Environment Variable Parsing:**  
   - Fixes and simplifies the logic for parsing environment variables to avoid misconfigurations and potential runtime errors.

### Details

- The implementation closely follows the [OpenTelemetry HTTP semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/), ensuring that all HTTP spans generated by the library are compliant.
- All relevant HTTP attributes in traces are set according to the specification, while omitting or redacting values that are considered sensitive.
- Includes tests and documentation updates to reflect the improved behavior.

### Motivation

- **Security:** Preventing leakage of sensitive information via telemetry is essential for compliance and user trust.
- **Standardization:** Aligning with OpenTelemetry semantic conventions improves interoperability with observability tools and makes traces easier to understand and analyze.

### Related Links

- [OpenTelemetry HTTP Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/)

---

Please let me know if further changes or clarifications are needed.

### DIFF
--- a/google/cloud/storage/_opentelemetry_tracing.py
+++ b/google/cloud/storage/_opentelemetry_tracing.py
@@ -115,8 +115,8 @@ def _set_api_request_attr(request, client):
     if request.get("path"):
         full_url = client._connection.build_api_url(request.get("path"))
         attr.update(_get_opentelemetry_attributes_from_url(full_url, strip_query=True))
-    if request.get("timeout"):
-        attr["connect_timeout,read_timeout"] = request.get("timeout")
+    if "timeout" in request:
+        attr["connect_timeout,read_timeout"] = str(request.get("timeout"))
     return attr
 
 
@@ -131,8 +131,16 @@ def _get_opentelemetry_attributes_from_url(
 ):
     """Helper to assemble OpenTelemetry span attributes from a URL."""
     u = urlparse(url)
+    netloc = u.netloc
+    # u.hostname is always lowercase. We parse netloc to preserve casing.
+    # netloc format: [userinfo@]host[:port]
+    if "@" in netloc:
+        netloc = netloc.split("@", 1)[1]
+    if ":" in netloc and not netloc.endswith("]"):  # Handle IPv6 literal
+        netloc = netloc.split(":", 1)[0]
+
     attributes = {
-        "server.address": u.hostname,
+        "server.address": netloc,
         "server.port": u.port,
         "url.scheme": u.scheme,
         "url.path": u.path,
--- a/google/cloud/storage/_opentelemetry_tracing.py
+++ b/google/cloud/storage/_opentelemetry_tracing.py
@@ -18,7 +18,7 @@ import logging
 import os
 
 from contextlib import contextmanager
-
+from urllib.parse import urlparse
 from google.api_core import exceptions as api_exceptions
 from google.api_core import retry as api_retry
 from google.cloud.storage import __version__
@@ -28,7 +28,15 @@ from google.cloud.storage.retry import ConditionalRetryPolicy
 ENABLE_OTEL_TRACES_ENV_VAR = "ENABLE_GCS_PYTHON_CLIENT_OTEL_TRACES"
 _DEFAULT_ENABLE_OTEL_TRACES_VALUE = False
 
-enable_otel_traces = os.environ.get(
+
+def _parse_bool_env(name: str, default: bool = False) -> bool:
+    val = os.environ.get(name, None)
+    if val is None:
+        return default
+    return str(val).strip().lower() in {"1", "true", "yes", "on"}
+
+
+enable_otel_traces = _parse_bool_env(
     ENABLE_OTEL_TRACES_ENV_VAR, _DEFAULT_ENABLE_OTEL_TRACES_VALUE
 )
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


google/cloud/storage/_opentelemetry_tracing.py
```diff
@@ -28,7 +28,15 @@
 ENABLE_OTEL_TRACES_ENV_VAR = "ENABLE_GCS_PYTHON_CLIENT_OTEL_TRACES"
 _DEFAULT_ENABLE_OTEL_TRACES_VALUE = False
 
-enable_otel_traces = os.environ.get(
+
+def _parse_bool_env(name: str, default: bool = False) -> bool:
+    val = os.environ.get(name, None)
+    if val is None:
+        return default
+    return str(val).strip().lower() in {"1", "true", "yes", "on"}
```
chandra-siri, line 36: nit: coercing to str is not required. Just  `val.strip().lower()` should work fine because output from [`os.environ.get`](https://docs.python.org/3/library/os.html#os.environ) is always str or None. If None, it's caught in line35 and `default` is returned.

tests/unit/test__opentelemetry_tracing.py
chandra-siri, line 1: We need to add one more unit test . It should test that tracing are disabled when a user sets `ENABLE_GCS_PYTHON_CLIENT_OTEL_TRACES` to False (or any value other than {1, true, on, yes} ) 

ie negative unit test of [`test_enable_trace_yield_span`](https://github.com/googleapis/python-storage/blob/d44de76184c48927ec94294a729f961e75a897ce/tests/unit/test__opentelemetry_tracing.py#L79)

google/cloud/storage/_opentelemetry_tracing.py
```diff
@@ -105,15 +113,37 @@ def _set_api_request_attr(request, client):
...
     if request.get("path"):
-        path = request.get("path")
-        full_path = f"{client._connection.API_BASE_URL}{path}"
-        attr["url.full"] = full_path
-    if request.get("timeout"):
-        attr["connect_timeout,read_timeout"] = request.get("timeout")
+        full_url = client._connection.build_api_url(request.get("path"))
+        attr.update(_get_opentelemetry_attributes_from_url(full_url, strip_query=True))
+    if "timeout" in request:
+        attr["connect_timeout,read_timeout"] = str(request.get("timeout"))
```
chandra-siri, line 119: there's no need to change timeout section of the code and convert into str()

tests/unit/test__opentelemetry_tracing.py
```diff
@@ -147,15 +147,19 @@ def test_get_final_attributes(setup, setup_optin):
         "rpc.system": "http",
         "user_agent.original": f"gcloud-python/{__version__}",
         "http.request.method": "GET",
-        "url.full": "https://testOtel.org/foo/bar/baz",
-        "connect_timeout,read_timeout": (100, 100),
+        "server.address": "testOtel.org",
+        "url.path": "/foo/bar/baz",
+        "url.scheme": "https",
+        "connect_timeout,read_timeout": str((100, 100)),
```
chandra-siri, line 167: if we don't change `timeout` param in the production file, we need not change the same here in test file.

chandra-siri (changes requested):

google/cloud/storage/_opentelemetry_tracing.py
rajeevpodar: It can vary as per the pre-processing environment variables at startup and populating os.environ with boolean values (True/False) instead of strings.

When I run this code without str(val), it crashes with an AttributeError: 'bool' object has no attribute 'strip' when the environment variable is set to True or False.

The str(val) is a deliberate safeguard to make this function robust against both strings (like "1", "true") and booleans (True), both of which are present in our live environment.

Here are the test report

- ERROR tests/unit/test__opentelemetry_tracing.py::test_enable_trace_yield_span - AttributeError: 'bool' object has no attribute 'strip'
- ERROR tests/unit/test__opentelemetry_tracing.py::test_enable_trace_call - AttributeError: 'bool' object has no attribute 'strip'
- ERROR tests/unit/test__opentelemetry_tracing.py::test_enable_trace_error - AttributeError: 'bool' object has no attribute 'strip'
- ERROR tests/unit/test__opentelemetry_tracing.py::test_get_final_attributes - AttributeError: 'bool' object has no attribute 'strip'
- ERROR tests/unit/test__opentelemetry_tracing.py::test_set_conditional_retry_attr - AttributeError: 'bool' object has no attribute 'strip'

google/cloud/storage/_opentelemetry_tracing.py
rajeevpodar: Ditto as [comment](https://github.com/googleapis/python-storage/pull/1553/files#r2377821106)

tests/unit/test__opentelemetry_tracing.py
rajeevpodar: Ignoring as per [comment](https://github.com/googleapis/python-storage/pull/1553/files#r2377821106)

tests/unit/test__opentelemetry_tracing.py
rajeevpodar: Added

chandra-siri (dismissed):

chandra-siri (approved):

PR #1554, opened 2025-09-25
feat(zb-experimental): Add support for closing MRD

chandra-siri:
* add `is_stream_open` and check it's value before doing any of the following operation. `open` , `close` , `send` ,`recv`
* Instantiate `socket_like_rpc` inside open, so that it is a new object every time we open. 
   * Otherwise, consider this scenario where a user is opening after closing then `initial_request` would be None because close sets it to None.


google/cloud/storage/_experimental/asyncio/async_read_object_stream.py
```diff
@@ -83,32 +83,41 @@ def __init__(
...
         self.read_handle = response.read_handle
 
+        self._is_stream_open = True
+
     async def close(self) -> None:
         """Closes the bidi-gRPC connection."""
+        if not self._is_stream_open:
+            raise ValueError("Stream is not open")
         await self.socket_like_rpc.close()
+        self._is_stream_open = False
```
Pulkit0110, line 120: should we interchange the sequence of these statements?

Reason is, if somehow `self.socket_like_rpc.close()` fails and throws error, the `_is_stream_open` would still be true.

tests/unit/asyncio/test_async_read_object_stream.py
```diff
@@ -27,16 +27,42 @@
 _TEST_READ_HANDLE = b"test-read-handle"
 
 
+async def instantiate_read_obj_stream(mock_client, mock_cls_async_bidi_rpc, open=True):
+    """Helper to instance an instance of _AsyncReadObjectStream and open it by default."""
```
Pulkit0110: nit: should be `Helper to create an instance of/instantiate`

Pulkit0110 (changes requested):

Pulkit0110 (dismissed):

google/cloud/storage/_experimental/asyncio/async_read_object_stream.py
chandra-siri: > Reason is, if somehow `self.socket_like_rpc.close()` fails and throws error, the `_is_stream_open` would still be true.

in this case because the close operation failed, `self._is_stream_open` should be False. why do you think the opposite ?

tests/unit/asyncio/test_async_read_object_stream.py
chandra-siri: done.

Pulkit0110 (approved):

PR #1595, opened 2025-10-29
fix: Use separate header object for each upload in Transfer Manager MPU

MattIrv:
I didn't see test failures due to this (are the transfer manager tests running continuously?) but in our own repo (https://github.com/GoogleCloudPlatform/gcs-connector-for-pytorch/pull/209) a copied implementation of this caused upload failures (with errors like `google.cloud.storage.exceptions.DataCorruption: Checksum mismatch: checksum calculated by client and server did not match. Error code: BadDigest, Error message: The MD5 you specified in Content-MD5 or x-goog-hash did not match what we computed., Error details: The specified CRC (-1631273713) does not match what we computed (1851139595)
`) because the headers dict gets updated in [upload.py](https://github.com/googleapis/python-storage/blob/2da90fc12394eceaf97df718f3e37d42e8f674a4/google/cloud/storage/_media/requests/upload.py#L766) but passing the headers dict to `executor.submit` here causes all threads to share the same object.


chandra-siri (approved):
Thanks for your fix!

ShreyasSinha (approved):

chandra-siri:
> @chandra-siri I don't see a way to re-run the approvers check so that I can submit this. Are you able to do so? Thanks!

done!

PR #1596, opened 2025-10-29
feat: provide option to update user_agent

### LINKED ISSUE #1327: Make it possible to update / add to the user-agent for an existing client object
**Is your feature request related to a problem? Please describe.**

My team is using the Python storage client. When we initialize a new client, we set the user-agent as part of the client_info passed [here](https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/client.py#L119). However, we'd like users to be able to pass in their own storage clients in order to do things like custom authentication or manage other options. Unfortunately the user-agent does not get set in this case, and there doesn't seem to be any way to edit or append to it for an existing client object.

 **Describe the solution you'd like**

Provide a method or field on the [Client object](https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/client.py#L66) that we can use to edit or append to the user-agent for an existing Client object.

 **Describe alternatives you've considered**

It may be possible for us to somehow duplicate the credentials / options / transport from an existing Client and create a new one and set the user-agent, but that seems error-prone. 

 **Additional context**

See [our existing code on GitHub](https://github.com/GoogleCloudPlatform/dataflux-pytorch/blob/main/dataflux_pytorch/dataflux_mapstyle_dataset.py#L106) for an example of how we set the user-agent for a new client today.

chandra-siri:
feat: provide option to update `user_agent`, fixes #1327

Once the Client is instantiated let SDK users append `user_agent` . 

Please Note: This doesn't let you remove `user_agent` provided by upstream application.


chandra-siri:
fixes #1327

google/cloud/storage/client.py
```diff
@@ -297,6 +297,18 @@ def universe_domain(self):
...
+        """Update the user-agent string for this client.
+
+        :type user_agent: str
+        :param user_agent: The string to add to the user-agent.
+        """
+        existing_user_agent = self._connection._client_info.user_agent
+        if existing_user_agent is None:
+            self._connection.user_agent = user_agent
+        else:
+            self._connection.user_agent = f"{user_agent} {existing_user_agent}"
```
MattIrv, line 310: On line 306 above we get the existing user_agent from `self._connection._client_info` but here we're setting it directly on `self._connection`. Does that matter?

I see that in our existing code to work around this we set it on `self._connection._client_info`: https://github.com/GoogleCloudPlatform/dataflux-client-python/blob/e2a455a94ddaa2b2f7cd1a3e560018fd1b56badc/dataflux_core/user_agent.py#L20

MattIrv (approved):

google/cloud/storage/client.py
chandra-siri: > On line 306 above we get the existing user_agent from self._connection._client_info but here we're setting it directly on self._connection. Does that matter?

No, it shouldn't matter, both are same. I'm using the [setter property](https://github.com/googleapis/python-cloud-core/blob/c26e587a79b1905f4d64e409b853d9e50bbb6c17/google/cloud/_http/__init__.py#L98-L100) to assign the value.

suni72 (approved):

PR #1620, opened 2025-11-16
feat(zb-experimental): implement append

chandra-siri:
feat(zb-experimental): implement append


google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
```diff
@@ -31,6 +31,10 @@
 )
 
 
+_MAX_CHUNK_SIZE_BYTES = 2 * 1024 * 1024  # 2 MiB
+_MAX_BUFFER_SIZE_BYTES = 150 * 1024 * 1024  # 16 MiB
```
Pulkit0110: What is the expected value here? `_MAX_BUFFER_SIZE_BYTES` is set to 150 MiB but the comment says 16 MiB?

google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
```diff
@@ -143,7 +147,35 @@ async def open(self) -> None:
...
+                        content=data[start_idx:end_idx]
+                    ),
+                )
+            )
+            chunk_size = end_idx - start_idx
+            self.offset += chunk_size
+            bytes_to_flush += chunk_size
+            if bytes_to_flush >= _MAX_BUFFER_SIZE_BYTES:
+                await self.flush()
+                bytes_to_flush = 0
```
Pulkit0110, line 203: What if `bytes_to_flush` never exceeds `_MAX_BUFFER_SIZE_BYTES`? example: append(1mb) is called?

google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
```diff
@@ -143,7 +147,35 @@ async def open(self) -> None:
...
+            start_idx = end_idx
```
Pulkit0110, line 204: Also, I think the `close()` function should check if there's any data that isn't flushed

tests/unit/asyncio/test_async_appendable_object_writer.py
```diff
@@ -267,3 +264,141 @@ async def test_finalize(mock_write_object_stream, mock_client):
...
+    writer.flush = mock.AsyncMock()
+
+    data1 = b"a" * (_MAX_CHUNK_SIZE_BYTES + 10)
+    await writer.append(data1)
+
+    data2 = b"b" * (_MAX_CHUNK_SIZE_BYTES + 20)
+    await writer.append(data2)
+
+    total_data_length = len(data1) + len(data2)
+    assert writer.offset == total_data_length
```
Pulkit0110, line 465: we should also assert how many time flush called?

Pulkit0110 (changes requested):

google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
chandra-siri: then flush will never be called from client. (server periodically flushes them at min(60s, 256MiB) , whichever occurs first)

google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
chandra-siri: corrected it, my bad. Used it for testing.

tests/unit/asyncio/test_async_appendable_object_writer.py
chandra-siri: Done.

google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
chandra-siri: added flush before close()

Pulkit0110 (approved):

PR #1637, opened 2025-12-04
feat: Auto enable mTLS when supported certificates are detected

agrawalradhika-cell:
The Python SDK will use a hybrid approach for mTLS enablement:

If the GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable is set (either true or false or any value), the SDK will respect that setting. This is necessary for test scenarios and users who need to explicitly control mTLS behavior.
If the GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable is not set, the SDK will automatically enable mTLS only if it detects Managed Workload Identity (MWID) or X.509 Workforce Identity Federation (WIF) certificate sources. In other cases where the variable is not set, mTLS will remain disabled.

### DIFF
--- a/tests/unit/gapic/storage_v2/test_storage.py
+++ b/tests/unit/gapic/storage_v2/test_storage.py
@@ -148,12 +148,19 @@ def test__read_environment_variables():
     with mock.patch.dict(
         os.environ, {"GOOGLE_API_USE_CLIENT_CERTIFICATE": "Unsupported"}
     ):
-        with pytest.raises(ValueError) as excinfo:
-            StorageClient._read_environment_variables()
-    assert (
-        str(excinfo.value)
-        == "Environment variable `GOOGLE_API_USE_CLIENT_CERTIFICATE` must be either `true` or `false`"
-    )
+        if not hasattr(google.auth.transport.mtls, "should_use_client_cert"):
+            with pytest.raises(ValueError) as excinfo:
+                StorageClient._read_environment_variables()
+            assert (
+                str(excinfo.value)
+                == "Environment variable `GOOGLE_API_USE_CLIENT_CERTIFICATE` must be either `true` or `false`"
+            )
+        else:
+            assert StorageClient._read_environment_variables() == (
+                False,
+                "auto",
+                None,
+            )
 
     with mock.patch.dict(os.environ, {"GOOGLE_API_USE_MTLS_ENDPOINT": "never"}):
         assert StorageClient._read_environment_variables() == (False, "never", None)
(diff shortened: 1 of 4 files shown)


chandra-siri:
kokoro failure should be fixed once this is merged - https://github.com/googleapis/python-storage/pull/1635

tests/unit/gapic/storage_v2/test_storage.py
```diff
@@ -176,6 +183,105 @@ def test__read_environment_variables():
...
+    if not hasattr(google.auth.transport.mtls, "should_use_client_cert"):
+        with mock.patch.dict(
+            os.environ, {"GOOGLE_API_USE_CLIENT_CERTIFICATE": "FALSE"}
+        ):
+            assert StorageClient._use_client_cert_effective() is False
+
+    # Test case 9: Test when `should_use_client_cert` is unavailable and the
+    # `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is not set.
+    # In this case, the method should return False, which is the default value.
+    if not hasattr(google.auth.transport.mtls, "should_use_client_cert"):
```
daniel-sanche, line 252: I see a lot of these checks. We generally want to avoid conditionals inside test logic, because we don't want to accidentally skip an important assertion.

This should be broken out into separate tests, with a skip statement based on the version of the auth library

tests/unit/gapic/storage_v2/test_storage.py
```diff
@@ -148,12 +148,19 @@ def test__read_environment_variables():
     with mock.patch.dict(
         os.environ, {"GOOGLE_API_USE_CLIENT_CERTIFICATE": "Unsupported"}
     ):
-        with pytest.raises(ValueError) as excinfo:
-            StorageClient._read_environment_variables()
-    assert (
-        str(excinfo.value)
-        == "Environment variable `GOOGLE_API_USE_CLIENT_CERTIFICATE` must be either `true` or `false`"
-    )
+        if not hasattr(google.auth.transport.mtls, "should_use_client_cert"):
```
daniel-sanche, line 151: I mentioned this in a different PR, but we should probably be checking based on the auth library version, to catch regressions in the future

tests/unit/gapic/storage_v2/test_storage.py
```diff
@@ -733,6 +828,119 @@ def test_storage_client_get_mtls_endpoint_and_cert_source(client_class):
...
+        (
+            # With workloads not present in config, mTLS is disabled.
+            {
+                "version": 1,
+                "cert_configs": {},
+            },
+            None,
+        ),
+    ]
+    if hasattr(google.auth.transport.mtls, "should_use_client_cert"):
```
daniel-sanche, line 920: again, we should avoid all these conditionals if possible. Can these be broken out into a new test, that is only run on the proper versions of auth?

daniel-sanche (changes requested):

tests/unit/gapic/storage_v2/test_storage.py
agrawalradhika-cell: Met offline with Daniel, this is change which will be duplicate from gapic-generator-python changes for mTLS support https://github.com/googleapis/gapic-generator-python/pull/2472, so comment can be ignored for now

tests/unit/gapic/storage_v2/test_storage.py
agrawalradhika-cell: Met offline with Daniel, this is change which will be duplicate from gapic-generator-python changes for mTLS support https://github.com/googleapis/gapic-generator-python/pull/2472, so comment can be ignored for now

daniel-sanche (approved):
You can ignore my comments on the auto-generated code, those should have been left in the gapic-generator PR

Hand-wrritten changes LGTM

tests/unit/gapic/storage_v2/test_storage.py
agrawalradhika-cell: Same as above

Linchin (approved):

PR #1664, opened 2025-12-15
feat(storage): Add AsyncConnection class and unit tests

googlyrahman:
Add AsyncConnection class and unit tests


google/cloud/storage/_http.py
chandra-siri, line 1: no need to change this file I guess.

google/cloud/storage/_experimental/asyncio/async_connection.py
chandra-siri, line 1: * did you lookout for async version of `https://github.com/googleapis/python-cloud-core/blob/main/google/cloud/_http/__init__.py#L227`  in `python-cloud-core` library? , if it already exits, then we're reinventing the wheel here.  

* If i'm not wrong this file is essentially an async version of that right ?  why didn't you opt for creating a common class for sync and async connection just like - https://github.com/googleapis/python-storage/pull/1655/changes ?

chandra-siri (changes requested):

google/cloud/storage/_experimental/asyncio/async_connection.py
googlyrahman: As per my knowledge, There is no async version of this client in `python-cloud-core`, The library doesn't even consists of "async" as keyword which is basic unit for async code, See https://github.com/search?q=repo%3Agoogleapis%2Fpython-cloud-core%20%22async%22&type=code, With that said, let me know if there is any existing library which we can utilize.

Regarding the decision not to create an interface, I have a few reasons for this approach and would value your feedback:
- Inheritance Complexity (Main Reason): The current sync implementation has a complex inheritance chain: Connection (Python SDK) $\rightarrow$ JSONConnection (python-cloud-core) $\rightarrow$ Connection (python-cloud-core).Defining a base abstract class in our library for implementations that exist in an external library creates fragility. If the upstream library changes without our knowledge, it causes breakage.
- Not a Public Interface: Unlike Client, Bucket, or Blob, this component is internal, so an interface seems unnecessary.
- Limited Utility: Even if we implemented an interface, it would likely only abstract a single method (api_request), which adds little value.
- It would require changes in the `python-cloud-core` repository, lengthening the development process.
- I prefer making least changes to existing sync client to avoid blast radius, and this should not impact us because down the line as per our plan the sync will eventually be wrapper on the async client. 

Let me know your thoughts about it?

google/cloud/storage/_http.py
googlyrahman: This is just to share the UserAgent Syntax with AsyncConnection and SyncConnection

google/cloud/storage/_experimental/asyncio/async_connection.py
chandra-siri: resolving , as per our offline discussion.

chandra-siri (approved):

PR #1709, opened 2026-01-16
feat(ZonalBuckets): add support for `generation=0` to avoid overwriting existing objects and add `is_stream_open` support 

chandra-siri:
feat(ZonalBuckets): add support for `generation=0` to prevent overwriting existing objects
feat(ZonalBuckets): add `is_stream_open` property to AsyncAppendableObjectWriter for stream status check

### DIFF
--- a/google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
+++ b/google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
@@ -97,13 +97,16 @@ class AsyncAppendableObjectWriter:
         :param object_name: The name of the GCS Appendable Object to be written.
 
         :type generation: int
-        :param generation: (Optional) If present, selects a specific revision of
-                            that object.
-                            If None, a new object is created.
-                            If None and Object already exists then it'll will be
-                            overwritten.
-                            To avoid overwriting existing object use
-                            `generation=0` which will fail if object exists.
+        :param generation: (Optional) If present, creates writer for that
+            specific revision of that object. Use this to append data to an
+            existing Appendable Object.
+
+            Setting to ``0`` makes the `writer.open()` succeed only if
+            there are no live versions of the object. (useful for not accidentally
+            overwriting existing objects).
+
+            Warning: If None, a new object is created. If an object with the
+            same name already exists, it will be overwritten.
 
         :type write_handle: bytes
         :param write_handle: (Optional) An existing handle for writing the object.
--- a/google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
+++ b/google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
@@ -346,6 +346,11 @@ class AsyncAppendableObjectWriter:
         self.offset = None
         return self.object_resource
 
+    @property
+    def is_stream_open(self) -> bool:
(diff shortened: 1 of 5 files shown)


google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
```diff
@@ -97,15 +97,29 @@ def __init__(
...
+        :param write_handle: (Optional) An handle for writing the object.
+            If provided, opening the bidi-gRPC connection will be faster.
+
+        :type writer_options: dict
+        :param writer_options: (Optional) A dictionary of writer options.
+            Supported options:
+            - "FLUSH_INTERVAL_BYTES": int
+                The number of bytes to append before "persisting" data in GCS
+                servers. Default is `_DEFAULT_FLUSH_INTERVAL_BYTES`.
+                Must be a multiple of `_MAX_CHUNK_SIZE_BYTES`.
```
ankitaluthra1, line 122: what will happen in case value is invalid like not multiple of max chunk size , would user get error at run time or just that value wont be used ? also lets add a test case if not already added

google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
```diff
@@ -97,15 +97,29 @@ def __init__(
...
-        :param generation: (Optional) If present, selects a specific revision of
-                            that object.
-                            If None, a new object is created.
-                            If None and Object already exists then it'll will be
-                            overwritten.
+        :param generation: (Optional) If present, creates writer for that
+            specific revision of that object. Use this to append data to an
+            existing Appendable Object.
+
+            Setting to ``0`` makes the `writer.open()` succeed only if
```
ankitaluthra1, line 104: with 0 value if the object with same name exist, user would get a run time exception ? should we also add a test case for this ?

google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
chandra-siri: tests already exits [here](https://github.com/googleapis/python-storage/blob/6c160794afded5e8f4179399f1fe5248e32bf707/tests/unit/asyncio/test_async_appendable_object_writer.py#L126-L150) , in this PR I'm just updating the doc string , because it was missed.

chandra-siri: > what will happen in case value is invalid like not multiple of max chunk size , would user get error at run time or just that value wont be used ? 

user will get runtime error, [code](https://github.com/googleapis/python-storage/blob/6c160794afded5e8f4179399f1fe5248e32bf707/google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py#L137-L144)

google/cloud/storage/_experimental/asyncio/async_appendable_object_writer.py
chandra-siri: yes, test case is already added in _test_zonal.py_ 

see -> `test_open_with_generation_zero`  & `test_open_existing_object_with_gen_None_overrides_existing`

ankitaluthra1 (approved):

PR #1714, opened 2026-01-21
refactor: move system tests to single event loop

chandra-siri:
refactor: move system tests to single event loop. 

* Without this more adding more tests were "shaky" , some of them crashed due to segmentation fault in underlying gRPC's c layer.


googlyrahman (approved):

jasha26 (commented):
May be not in this PR, but i guess we should refactor this file as there is a lot of duplication. I suggest following

1. Create a fixture that handles both the generation of a unique object_name and its subsequent cleanup. This removes the need to manually append to blobs_to_delete and call gc.collect() in every test. We do this almost in every test here
2. Instead of defining a nested async def _run() and calling run_until_complete in every function, create a module-level helper or a fixture that accepts a coroutine.
3. Can we implement context manager for reader and writer so we can simplify the usage like below:
```
async with AsyncAppendableObjectWriter(client, bucket, name) as writer:
await writer.append(data)
```
4. [Optional]: Since many tests switch between grpc_client and grpc_client_direct, the existing grpc_clients fixture is already a good step. To further reduce code, you can use a fixture that automatically selects the client based on a test parameter (e.g., attempt_direct_path).

PR #1757, opened 2026-02-20
feat: Triger bidi conformance tests

chandra-siri:
feat: Triger bidi conformance tests

Changes / fixes

- When (read/write) stream breaks, don't do `.close` on it.  (this leads to reading the same error message again and again until retry loop times out)
- Bidi Conf test were ran directly as a module , change that so that they get triggered in Kokoro (done that by adding `test_`  so that pytest pickup the tests)
- Provide a way to create anonymous connection in grpc client. 
- Unit tests and other minor changes.

~blocked because of b/489420625~

### DIFF
--- a/google/cloud/storage/asyncio/async_grpc_client.py
+++ b/google/cloud/storage/asyncio/async_grpc_client.py
@@ -19,6 +19,8 @@ from google.cloud._storage_v2.services.storage.transports.base import (
     DEFAULT_CLIENT_INFO,
 )
 from google.cloud.storage import __version__
+import grpc
+from google.auth import credentials as auth_credentials
 
 
 class AsyncGrpcClient:
@@ -51,7 +53,12 @@ class AsyncGrpcClient:
         client_options=None,
         *,
         attempt_direct_path=True,
+        create_insecure_channel=False,  # only for testing against testbench.
     ):
+        if create_insecure_channel:
+            self._grpc_client = self._create_insecure_grpc_client(client_options)
+            return
+
         if client_info is None:
             client_info = DEFAULT_CLIENT_INFO
         client_info.client_library_version = __version__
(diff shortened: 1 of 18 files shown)


tests/conformance/test_bidi_reads.py
```diff
@@ -121,12 +132,12 @@ async def main():
...
-            "method": "storage.objects.get",
-            "instruction": "return-broken-stream-after-2K",
-            "expected_error": None,
-        },
+        # {
+        #     "name": "Smarter Resumption: Retry 503 after partial data",
+        #     "method": "storage.objects.get",
+        #     "instruction": "return-broken-stream-after-2K",
+        #     "expected_error": None,
+        # },
```
Pulkit0110: Why smarter resumption scenarios are commented out?

chandra-siri:
closing for migration.

tests/conformance/test_bidi_reads.py
chandra-siri: fixed it!

tests/conformance/test_bidi_writes.py
```diff
@@ -1,22 +1,68 @@
...
+def testbench():
+    subprocess.run(_PULL_CMD)
+    proc = subprocess.Popen(_RUN_CMD)
+    time.sleep(10)
+    yield GRPC_ENDPOINT, HTTP_ENDPOINT
+    subprocess.run(_DOCKER_STOP_CMD)
+    proc.kill()
+
+
+CONTENT = b"A" * 1024 * 1024 * 10  # 10 KB
```
v-pratap: is it 10 MB?

google/cloud/storage/asyncio/async_grpc_client.py
```diff
@@ -68,6 +76,21 @@ def __init__(
             attempt_direct_path=attempt_direct_path,
         )
 
+    def _create_anonymous_client(self, client_options, credentials):
```
v-pratap, line 79: Just for my knowledge
is the what would happen if client_options is empty?
is that even a possible case?

v-pratap (approved):

tests/conformance/test_bidi_writes.py
chandra-siri: yes, thanks for pointing out.

google/cloud/storage/asyncio/async_grpc_client.py
chandra-siri: in that case, the code will break.  But IMO that's ok because it's private method (see leading _) and intented to be used only by us (for test-bench)

